PR #214, opened 2020-05-26
[test] | [eks] change: Sequence EKS training and inference tests

saimidu:
*Issue #, if available:*

## Checklist

*Description:*
- MXNet Single Node test jobs time out because of lack unavailability of pods.
- So use an EKS cluster unique to each CI test on a PR.
- Reduce number of nodes on PR EKS tests to 1 node per cluster. This seems to have no impact on the success of tests, and reduces cluster spin-up time.

*Tests run:*
- Executed test for MXNet specifically using local clone of repository.

*DLC image/dockerfile:*
- N/A

*Additional context:*


test/dlc_tests/eks/eks_manifest_templates/mxnet/inference/single_node_cpu_inference.yaml
```diff
@@ -45,7 +45,7 @@ spec:
         resources:
           limits:
             cpu: "1"
-            memory: 1Gi
+            memory: 2Gi
```
saimidu, line 48: This increase is for gluonnlp inference tests.

test/testrunner.py
```diff
@@ -130,11 +130,15 @@ def setup_eks_clusters(dlc_images):
...
     cluster_name = f"dlc-{short_name}-cluster-" \
                    f"{os.getenv('CODEBUILD_RESOLVED_SOURCE_VERSION')}-{random.randint(1, 10000)}"
-    eks_utils.create_eks_cluster(cluster_name, "gpu", num_nodes, "p3.16xlarge", "pytest.pem")
-    eks_utils.eks_setup(long_name, cluster_name)
+    try:
+        eks_utils.create_eks_cluster(cluster_name, "gpu", num_nodes, "p3.16xlarge", "pytest.pem")
+        eks_utils.eks_setup(long_name, cluster_name)
+    except Exception as e:
+        eks_utils.delete_eks_cluster(cluster_name)
+        raise RuntimeError(e)
```
saimidu, line 141: This is a modification that comes from a change request on an older PR - https://github.com/aws/deep-learning-containers/pull/218#discussion_r432798693

test/testrunner.py
```diff
@@ -160,14 +166,23 @@ def main():
...
+            # Execute dlc_tests pytest command
+            pytest_cmds = [["-s", "-rA", test_path, f"--junitxml={report}", "-n=auto"]]
         try:
-            sys.exit(pytest.main(pytest_cmd))
+            cmd_exit_statuses = [pytest.main(pytest_cmd) for pytest_cmd in pytest_cmds]
+            sys.exit(0) if all([status == 0 for status in cmd_exit_statuses]) else sys.exit(1)
         finally:
-            if specific_test_type == "eks":
-                eks_utils.delete_eks_cluster(new_eks_cluster_name)
+            if specific_test_type == "eks" and eks_cluster_name:
```
saimidu, line 184: This change is not really necessary, because eks_cluster_name is only assigned/used when `specific_test_type == "eks"`. This change is only to cover all bases, and make debugging easier for developers in the future.

test/testrunner.py
arjkesh: nit: just raise the existing exception instead of raising it as a runtime error

test/testrunner.py
```diff
@@ -160,14 +166,23 @@ def main():
         if specific_test_type == "sanity":
             pull_dlc_images(all_image_list)
         if specific_test_type == "eks":
-            new_eks_cluster_name = setup_eks_clusters(dlc_images)
-        # Execute dlc_tests pytest command
-        pytest_cmd = ["-s", "-rA", test_path, f"--junitxml={report}", "-n=auto"]
+            eks_cluster_name = setup_eks_clusters(dlc_images)
+            # We expect image URIs of only one framework for EKS tests.
+            framework = "mxnet" if "mxnet" in dlc_images else "pytorch" if "pytorch" in dlc_images else "tensorflow"
```
arjkesh, line 171: how about making setup_eks_clusters also return the framework name of the just-set-up eks cluster? that way you can avoid this repeated logic

test/testrunner.py
```diff
@@ -145,13 +149,15 @@ def main():
...
     standard_images_list = [image_uri for image_uri in all_image_list if "example" not in image_uri]
-    new_eks_cluster_name = None
+    eks_cluster_name = None
     benchmark_mode = "benchmark" in test_type
     specific_test_type = re.sub("benchmark-", "", test_type) if benchmark_mode else test_type
     test_path = os.path.join("benchmark", specific_test_type) if benchmark_mode else specific_test_type
 
     if specific_test_type in ("sanity", "ecs", "ec2", "eks"):
         report = os.path.join(os.getcwd(), "test", f"{test_type}.xml")
+        report_train = os.path.join(os.getcwd(), "test", f"{test_type}_train.xml")
```
arjkesh, line 159: I see that you're trying to make this extensible to other test types, but it could get confusing as to why this is here. Can you add a clarifying comment, and note that only eks tests leverage this currently

saimidu: Makes sense. I can add this.

test/testrunner.py
saimidu: I would rather not make the `setup_eks_clusters` function do anything apart from its intended purpose. It would make sense now. But at a later point in time, without this context, understanding its purpose would be harder.

On the other hand, `setup_eks_clusters` has no actual need for the entire list of images except to identify the framework for the cluster it will create. So we can just identify the framework here in `main()`, and set only that as the setup function's parameter.

test/testrunner.py
```diff
@@ -160,14 +162,29 @@ def main():
...
-        pytest_cmd = ["-s", "-rA", test_path, f"--junitxml={report}", "-n=auto"]
+            frameworks_in_images = [framework for framework in ("mxnet", "pytorch", "tensorflow")
+                                    if framework in dlc_images]
+            if len(frameworks_in_images) != 1:
+                raise ValueError(
+                    f"All images in dlc_images must be of a single framework for EKS tests.\n"
+                    f"Instead seeing {frameworks_in_images} frameworks."
+                )
+            framework = frameworks_in_images[0]
+            eks_cluster_name = setup_eks_clusters(dlc_images)
```
arjkesh, line 173: This will likely not work anymore now that you've changed setup_eks_clusters.

test/testrunner.py
```diff
@@ -120,21 +120,19 @@ def pull_dlc_images(images):
         run(f"docker pull {image}", hide="out")
 
 
-def setup_eks_clusters(dlc_images):
+def setup_eks_clusters(framework_name):
```
arjkesh, line 123: nit: please change to "setup_eks_cluster" as we are enforcing a singular cluster

test/testrunner.py
```diff
@@ -120,21 +120,19 @@ def pull_dlc_images(images):
...
     cluster_name = f"dlc-{short_name}-cluster-" \
                    f"{os.getenv('CODEBUILD_RESOLVED_SOURCE_VERSION')}-{random.randint(1, 10000)}"
-    eks_utils.create_eks_cluster(cluster_name, "gpu", num_nodes, "p3.16xlarge", "pytest.pem")
-    eks_utils.eks_setup(long_name, cluster_name)
+    try:
+        eks_utils.create_eks_cluster(cluster_name, "gpu", num_nodes, "p3.16xlarge", "pytest.pem")
+        eks_utils.eks_setup(long_name, cluster_name)
+    except Exception as e:
+        eks_utils.delete_eks_cluster(cluster_name)
+        raise Exception(e)
```
arjkesh, line 135: This will now raise exception type "Exception"

Consider the following two examples:

In the first, we do not overwrite the exception as "Exception" type and get an assertion error.
```
In [2]: try:
   ...:     assert 1==2, "not true"
   ...: except:
   ...:     print('test')
   ...:     raise
   ...:
   ...:
test
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-2-52478dbb11bf> in <module>
      1 try:
----> 2     assert 1==2, "not true"
      3 except:
      4     print('test')
      5     raise

AssertionError: not true
```

Here, we raise "Exception(e)" and ultimately surface a redundant "Exception" error
```
In [3]: try:
   ...:     assert 1==2, "not true"
   ...: except Exception as e:
   ...:     print('test')
   ...:     raise Exception(e)
   ...:
   ...:
   ...:
test
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-3-698ec5fdd6d8> in <module>
      1 try:
----> 2     assert 1==2, "not true"
      3 except Exception as e:

AssertionError: not true

During handling of the above exception, another exception occurred:

Exception                                 Traceback (most recent call last)
<ipython-input-3-698ec5fdd6d8> in <module>
      3 except Exception as e:
      4     print('test')
----> 5     raise Exception(e)
      6
      7

Exception: not true
```

test/testrunner.py
saimidu: Pushed the wrong commit. Correcting it now.

test/testrunner.py
saimidu: Will do.

test/testrunner.py
saimidu: Changing this. How about
```
try:
    assert False, "not true"
except Exception:
    print("Running cleanup tasks")
    raise
```
because it looks like an
```
except:
```
statement goes against PEP recommendations, but
```
except Exception:
```
doesn't, even though we do end up not specifically using the Exception information.

arjkesh: Sure that's fine. You can also do this if you want to be a little more clear.

```
In [2]: try:
   ...:     assert 1==2, "not true"
   ...: except Exception as e:
   ...:     print('test')
   ...:     raise(e)
   ...:
   ...:
test
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-2-d91d2ae79f44> in <module>
      3 except Exception as e:
      4     print('test')
----> 5     raise(e)
      6
      7

<ipython-input-2-d91d2ae79f44> in <module>
      1 try:
----> 2     assert 1==2, "not true"
      3 except Exception as e:
      4     print('test')
      5     raise(e)

AssertionError: not true
```

saimidu: I tried this on my system - It caused two instances of the same Exception, one from the line where it originally came, and the other where it was raised again.

That's why I went with the approach I used.

arjkesh:
One thing to take note is that this will affect logging -- we'll now see a summary at the end of the logs, but it will be for just inference. Will have to scroll up to see the training logs. XML reports looks like they are reported together appropriately, though

test/testrunner.py
```diff
@@ -121,21 +121,19 @@ def pull_dlc_images(images):
...
-        raise ValueError(
-            f"All images in dlc_images must be of a single framework for EKS tests.\n"
-            f"Instead seeing {frameworks_in_images} frameworks."
-        )
-    long_name = frameworks_in_images[0]
+    long_name = framework_name
     short_name = frameworks[long_name]
-    num_nodes = 2 if is_pr_context() else 3 if long_name != "pytorch" else 4
+    num_nodes = 1 if is_pr_context() else 3 if long_name != "pytorch" else 4
     cluster_name = f"dlc-{short_name}-cluster-" \
```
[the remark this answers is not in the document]
saimidu: Investigating. Thanks for calling this out.

test/testrunner.py
```diff
@@ -170,18 +172,32 @@ def main():
...
+            if len(frameworks_in_images) != 1:
+                raise ValueError(
+                    f"All images in dlc_images must be of a single framework for EKS tests.\n"
+                    f"Instead seeing {frameworks_in_images} frameworks."
+                )
+            framework = frameworks_in_images[0]
+            eks_cluster_name = setup_eks_cluster(framework)
+            # Split training and inference, and run one after the other, to prevent scheduling issues
+            pytest_cmds = [
+                ["-s", "-rA", os.path.join(test_path, framework, "training"), f"--junitxml={report_train}", "-n=4"],
```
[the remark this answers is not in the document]
saimidu: Addressed in comment.

TusharKanekiDey:
Need to have these packages I guess -
curl -o kubectl https://amazon-eks.s3.us-west-2.amazonaws.com/1.14.6/2019-08-22/bin/linux/amd64/kubectl
curl -o aws-iam-authenticator https://amazon-eks.s3-us-west-2.amazonaws.com/1.14.6/2019-08-22/bin/linux/amd64/aws-iam-authenticator

test/test_utils/eks.py
```diff
@@ -316,12 +316,13 @@ def eks_setup(framework, cluster_name=None):
 
     kubectl_download_command = (
         f"curl --silent --location https://amazon-eks.s3-us-west-2.amazonaws.com/"
-        f"{EKS_VERSION}/2019-08-14/bin/{platform.lower()}/amd64/kubectl -o /tmp/kubectl"
+        f"{EKS_VERSION}/2019-08-14/bin/{platform.lower()}/amd64/kubectl -o /usr/local/bin/kubectl"
```
TusharKanekiDey, line 319: Will it be 2019-08-22 or 2019-08-14 ?(https://docs.aws.amazon.com/eks/latest/userguide/install-kubectl.html)

saimidu: For some reason, I was able to curl the binary on my laptop, so I didn't double-check the date. Thanks for pointing this out!

test/dlc_tests/eks/mxnet/training/test_eks_mxnet_training.py
```diff
@@ -21,7 +22,7 @@ def test_eks_mxnet_single_node_training(mxnet_training):
     pod_name = f"mxnet-single-node-training-{rand_int}"
 
     args = (
-        "git clone https://github.com/apache/incubator-mxnet.git && python "
+        "git clone -b 1.6.0 https://github.com/apache/incubator-mxnet.git && python "
```
arjkesh, line 25: nit: is there a way we could get this value from the buildspec, to avoid having to manually change this when there are new versions?

test/test_utils/eks.py
```diff
@@ -279,19 +279,25 @@ def create_eks_cluster(eks_cluster_name, processor_type, num_nodes,
     eksctl_create_cluster_command += " --auto-kubeconfig "
     run(eksctl_create_cluster_command)
 
+    eks_write_kubeconfig(eks_cluster_name, "us-west-2")
+
+    run(
```
arjkesh, line 284: I know right now we're spinning up GPU clusters, but will this break with a CPU cluster?

test/dlc_tests/eks/mxnet/training/test_eks_mxnet_training.py
saimidu: Getting it from the buildspec might be hard, but it may be possible to recover this information from the DLC tag.

test/test_utils/eks.py
saimidu: I assume you're referring to the NVIDIA device plugin change.
We've been applying the device plugin for a while now in the code with all clusters that are created. So I don't think it should change anything significantly to do this here.

test/dlc_tests/eks/mxnet/training/test_eks_mxnet_training.py
```diff
@@ -18,12 +18,18 @@ def test_eks_mxnet_single_node_training(mxnet_training):
 
     rand_int = random.randint(4001, 6000)
 
+    framework_version_search = re.search(r"\d+(\.\d+){2}", mxnet_training)
```
arjkesh, line 21: you can actually use get_framework_and_version_from_tag from test_utils/__init__.py

also feel free to update the function with this regex logic

saimidu: Ah, yes, I forgot about this. Thanks for pointing it out. I'll see if this change can be made.

test/test_utils/eks.py
```diff
@@ -302,26 +308,20 @@ def eks_setup(framework, cluster_name=None):
...
     if eks_tools_installed:
-        eks_write_kubeconfig(eks_cluster_name, "us-west-2")
         return
 
     platform = run("uname -s").stdout.strip()
 
     kubectl_download_command = (
         f"curl --silent --location https://amazon-eks.s3-us-west-2.amazonaws.com/"
-        f"{EKS_VERSION}/2019-08-14/bin/{platform.lower()}/amd64/kubectl -o /tmp/kubectl"
+        f"{EKS_VERSION}/2019-08-22/bin/{platform.lower()}/amd64/kubectl -o /usr/local/bin/kubectl"
```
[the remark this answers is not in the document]
saimidu: The change in the date, or the change in the target path?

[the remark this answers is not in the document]
saimidu: The date was taken from https://docs.aws.amazon.com/eks/latest/userguide/install-kubectl.html, as pointed out by @TusharKanekiDey .

I changed the target path because it reduces the number of steps in the code.

PR #218, opened 2020-05-28
[test] | [eks] change: Create a new EKS cluster for each PR test set

saimidu:
*Issue #, if available:*

## Checklist

*Description:*
- PRs previously used one single central PR-cluster for EKS tests, which had a preconfigured number of nodes. The pods from all PRs were scheduled on this single cluster.
- This design leads to some pods failing to be scheduled when there are a large number of parallel EKS tests.
- This problem has been resolved by creating an EKS cluster for PRs with a smaller number of nodes, so that each PR will have its own set of nodes upon which pods can be scheduled.

*Tests run:*
Tested locally by executing EKS PyTorch tests for all PyTorch images at the same time, because PyTorch EKS tests are where this issue manifests.

*DLC image/dockerfile:*

*Additional context:*


test/testrunner.py
```diff
@@ -123,16 +123,21 @@ def pull_dlc_images(images):
...
     for long_name, short_name in frameworks.items():
         if long_name in dlc_images:
-            cluster_name = None
-            if not is_pr_context():
-                num_nodes = 3 if long_name != "pytorch" else 4
-                cluster_name = f"dlc-{short_name}-cluster-" \
-                               f"{os.getenv('CODEBUILD_RESOLVED_SOURCE_VERSION')}-{random.randint(1, 10000)}"
-                eks_utils.create_eks_cluster(cluster_name, "gpu", num_nodes, "p3.16xlarge", "pytest.pem")
-                terminable_clusters.append(cluster_name)
+            num_nodes = 1 if is_pr_context() else 3 if long_name != "pytorch" else 4
```
YYStreet, line 134: Will this change make an influence on TF and MXNet EKS tests? Should we also test them under this PR context?

test/testrunner.py
```diff
@@ -123,16 +123,21 @@ def pull_dlc_images(images):
...
+            cluster_name = f"dlc-{short_name}-cluster-" \
+                           f"{os.getenv('CODEBUILD_RESOLVED_SOURCE_VERSION')}-{random.randint(1, 10000)}"
+            eks_utils.create_eks_cluster(cluster_name, "gpu", num_nodes, "p3.16xlarge", "pytest.pem")
+            terminable_clusters.append(cluster_name)
             eks_utils.eks_setup(long_name, cluster_name)
+            break
```
arjkesh, line 140: you can instead just return terminable_clusters here -- but since we are only returning one cluster now, you can just return cluster name and get rid of terminable clusters altogether.

Some logic will need to be changed in main so that eks doesn't expect a list

test/testrunner.py
saimidu: Yes. It probably will. I'm trying to make it work with PyTorch first, and then resolve issues that may crop up with other frameworks in afterwards.

test/testrunner.py
saimidu: Sure. I will make that change. Thanks.

test/testrunner.py
```diff
@@ -121,19 +121,21 @@ def pull_dlc_images(images):
...
+        raise ValueError(
+            f"All images in dlc_images must be of a single framework for EKS tests.\n"
+            f"Instead seeing {frameworks_in_images} frameworks."
+        )
+    long_name = frameworks_in_images[0]
+    short_name = frameworks[long_name]
+    num_nodes = 2 if is_pr_context() else 3 if long_name != "pytorch" else 4
+    cluster_name = f"dlc-{short_name}-cluster-" \
+                   f"{os.getenv('CODEBUILD_RESOLVED_SOURCE_VERSION')}-{random.randint(1, 10000)}"
+    eks_utils.create_eks_cluster(cluster_name, "gpu", num_nodes, "p3.16xlarge", "pytest.pem")
```
TusharKanekiDey, line 136: Can we use p3.8xlarge for PyTorch ? Previously we used to have that instead.

test/testrunner.py
```diff
@@ -158,15 +160,14 @@ def main():
         if specific_test_type == "sanity":
             pull_dlc_images(all_image_list)
         if specific_test_type == "eks":
-            eks_terminable_clusters = setup_eks_clusters(dlc_images)
+            new_eks_cluster_name = setup_eks_clusters(dlc_images)
```
arjkesh, line 163: nit: let's name this something like "eks_test_cluster" or "eks_cluster_name" -- the "new" is not needed

test/testrunner.py
```diff
@@ -158,15 +160,14 @@ def main():
...
         # Execute dlc_tests pytest command
         pytest_cmd = ["-s", "-rA", test_path, f"--junitxml={report}", "-n=auto"]
         try:
             sys.exit(pytest.main(pytest_cmd))
         finally:
-            if specific_test_type == "eks" and eks_terminable_clusters:
-                for cluster in eks_terminable_clusters:
-                    eks_utils.delete_eks_cluster(cluster)
+            if specific_test_type == "eks":
```
arjkesh, line 169: Can you also add "new_eks_cluster_name" as a condition here? The reason being, why set it to None in the first place -- if it somehow sneaks through as None, we will try to call delete_eks_cluster on "None"

test/testrunner.py
```diff
@@ -121,19 +121,21 @@ def pull_dlc_images(images):
...
+            f"All images in dlc_images must be of a single framework for EKS tests.\n"
+            f"Instead seeing {frameworks_in_images} frameworks."
+        )
+    long_name = frameworks_in_images[0]
+    short_name = frameworks[long_name]
+    num_nodes = 2 if is_pr_context() else 3 if long_name != "pytorch" else 4
+    cluster_name = f"dlc-{short_name}-cluster-" \
+                   f"{os.getenv('CODEBUILD_RESOLVED_SOURCE_VERSION')}-{random.randint(1, 10000)}"
+    eks_utils.create_eks_cluster(cluster_name, "gpu", num_nodes, "p3.16xlarge", "pytest.pem")
+    eks_utils.eks_setup(long_name, cluster_name)
```
arjkesh, line 137: There is a condition i just realized where create could succeed but setup could fail, and we would not properly clean up the cluster. I think it might be wise to wrap the eks setup call in try/finally and always return cluster name after we've created the eks cluster, so that in this case we won't leave clusters spun up

PR #238, opened 2020-06-04
[test]|[ec2] Add instance type to ec2 key pair

arjkesh:
## Checklist

*Description:*
If running in mainline context, add the instance type to ensure we are using unique ec2 key pairs.

*Tests run:*
Testing in personal account

### DIFF
--- a/test/dlc_tests/conftest.py
+++ b/test/dlc_tests/conftest.py
@@ -207,14 +207,16 @@ def generate_unique_values_for_fixtures(metafunc_obj, images_to_parametrize, val
             if key in metafunc_obj.fixturenames:
                 fixtures_parametrized[new_fixture_name] = []
                 for index, image in enumerate(images_to_parametrize):
-                    if "gpu" in image:
-                        gpu_env = os.getenv("EC2_GPU_INSTANCE_TYPE")
-                        instance_tag = f"-{gpu_env.replace('.', '-')}" if gpu_env else ""
-                    elif "cpu" in image:
-                        cpu_env = os.getenv("EC2_CPU_INSTANCE_TYPE")
-                        instance_tag = f"-{cpu_env.replace('.', '-')}" if cpu_env else ""
-                    else:
-                        instance_tag = ""
+
+                    # Tag fixtures with EC2 instance types if env variable is present
+                    allowed_processors = ("gpu", "cpu", "eia")
+                    instance_tag = ""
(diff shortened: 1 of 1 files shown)


test/dlc_tests/conftest.py
```diff
@@ -207,12 +207,21 @@ def generate_unique_values_for_fixtures(metafunc_obj, images_to_parametrize, val
...
                 fixtures_parametrized[new_fixture_name] = []
                 for index, image in enumerate(images_to_parametrize):
+                    if "gpu" in image:
+                        gpu_env = os.getenv("EC2_GPU_INSTANCE_TYPE")
+                        instance_tag = f"-{gpu_env.replace('.', '-')}" if gpu_env else ""
+                    elif "cpu" in image:
+                        cpu_env = os.getenv("EC2_CPU_INSTANCE_TYPE")
+                        instance_tag = f"-{cpu_env.replace('.', '-')}" if cpu_env else ""
+                    else:
+                        instance_tag = ""
```
YYStreet, line 217: Is there a situation this line will be executed without exception?

test/dlc_tests/conftest.py
```diff
@@ -207,12 +207,23 @@ def generate_unique_values_for_fixtures(metafunc_obj, images_to_parametrize, val
...
+
+                    # Tag fixtures with EC2 instance types if env variable is present
+                    allowed_processors = ("gpu", "cpu", "eia")
+                    instance_tag = ""
+                    for processor in allowed_processors:
+                        if processor in image:
+                            instance_type = os.getenv(f"EC2_{processor.upper()}_INSTANCE_TYPE")
+                            if instance_type:
+                                instance_tag = f"-{instance_type.replace('.', '-')}"
+                                break
```
saimidu, line 219: It looks like this only inserts the instance type into the Name tag of the EC2 instance used for the test, and doesn't change the behavior of the actual test or the instance used for the test if there is no env variable of this name.

Is this understanding right?

arjkesh: yep, that is correct.

arjkesh: Oh sorry, misread your question. This should parametrize ec2_key_name with the instance type if the env variable is present.

saimidu: Yup. Thanks for stating it more precisely.

PR #308, opened 2020-06-25
[build]  Add support for specifying binary links in buildspec

nskool:
*Issue #, if available:*

## Checklist
- [] (If applicable) I've reviewed the licenses of updated and new binaries and their dependencies to make sure all licenses are on the Apache Software Foundation Third Party License Policy Category A or Category B license list.  See [https://www.apache.org/legal/resolved.html](https://www.apache.org/legal/resolved.html).
- [] (If applicable) I've scanned the updated and new binaries to make sure they do not have vulnerabilities associated with them.

*Description:*
1. Add functionality to specify a binary link in buildspec. A VAR_IN_DOCKERFILE ref can be specified. This allows the binary to be referred by a particular ARG in the dockerfile.
2. A test checks if the binary is public/private since the binary should be public before PR is merged.

*Tests run:*
Build command used to test:
```
python src/main.py --buildspec tensorflow/buildspec.yml --framework tensorflow --image_types training --device_types cpu --py_versions py3
```
*DLC image/dockerfile:*
Tensorflow 2.1.1 dockerfiles.

*Additional context:*


src/utils.py
```diff
@@ -59,6 +60,48 @@ def do_build_all_images():
...
+    Fetch remote files and save with provided local_path name
+    :param link_type: string
+    :param remote_url: string
+    :return: file_name: string
+    """
+    LOGGER.info(f"Downloading {remote_url}")
+
+    ctx = Context()
+    file_name = ctx.run(f"basename {remote_url}").stdout.strip()
+    LOGGER.info(f"basename: {file_name}")
```
saimidu, line 91: Why does file_name need to be identified through `ctx.run`? Why not use os.path.basename()?

src/utils.py
```diff
@@ -59,6 +60,48 @@ def do_build_all_images():
...
+
+    if link_type in ["s3"] and remote_url.startswith("s3://"):
+        match = re.match(r's3://(.+?)/(.+)', remote_url)
+        bucket_name = match.group(1)
```
saimidu, line 95: Add escape characters to "/"s in the regular expression.

saimidu: Also, how is `(.+?)` different from `(.+)`?

src/utils.py
nskool: Thanks, corrected to using os.path.basename.

src/utils.py
nskool: Forward slashes needn't be escaped. IDE also recommends it as redundant. However, for generic representation in line with other languages, I've added them back. 

(.+?) will match up to one entry between slashes, that'll form the 'bucket', the rest becomes the key, including slashes.

saimidu: Thanks. I saw recommendations to escape forward slashes, but if it is redundant, then that's fine.

Sure. `(.+?)` will match with a bucket name, but why do we need the `?`? Doesn't it signify that a particular character/group is optional? (it needs to occur between 0 and 1 times)

nskool: In this case - the bucket name is expected to exist and will match up to 1 occurrences of the bucket name.

PR #327, opened 2020-07-01
[eia] Build eia image

YYStreet:
*Issue #, if available:*

## Checklist

*Description:*

*Tests run:*

*DLC image/dockerfile:*

*Additional context:*


tensorflow/inference/docker/1.15.0/py2/Dockerfile.eia
```diff
@@ -47,23 +47,22 @@ RUN ${PIP} --no-cache-dir install --upgrade \
...
+    protobuf \
 # using --no-dependencies to avoid installing tensorflow binary
  && ${PIP} install --no-dependencies --no-cache-dir \
     tensorflow-serving-api==1.15.0
 
 WORKDIR /
 
 # Some TF tools expect a "python" binary
-RUN ln -s $(which ${PYTHON}) /usr/local/bin/python \
- && ln -s /usr/local/bin/pip /usr/bin/pip
```
TusharKanekiDey: Why did we remove this symlink ?

YYStreet: It fails the build. 
```
ln: failed to create symbolic link '/usr/bin/pip': File exists
```
We do not have the 2nd line of symlink for other py2 images

pytorch/buildspec-ei.yml
```diff
@@ -8,7 +8,7 @@
     inference_repository: &INFERENCE_REPOSITORY
         image_type: &INFERENCE_IMAGE_TYPE inference
         root: !join [ *FRAMEWORK, "/", *INFERENCE_IMAGE_TYPE ]
-        repository_name: &REPOSITORY_NAME !join [pr, "-", *FRAMEWORK, "-", *INFERENCE_IMAGE_TYPE, "-", eia]
```
TusharKanekiDey: After this change, Sagemaker test has always failed due to path issue.

YYStreet: I am using existing pr repos for eia. I am not sure about the repo name should be either "pr-<framework>-inference-eia" or "pr-framework-eia". 
Are we always testing normal SM tests on EI images?

TusharKanekiDey: With EI Images, we do not run the local test.
In Remote test, test_ei.py instead of test_tfs.py for TensorFlow inference and few similar change.
Few test like test_gluonnlp and test_gluoncv will be skipped.

TusharKanekiDey: My point is that it is looking for path like sagemaker_tests/mxnet/eia which does not exist in our repo.
We need to refer the sagemaker_tests/mxnet/integration instead for now or in the future, create a folder sagemaker_tests/mxnet/eia. I am in favor of the first one.

YYStreet: If they are sharing some of the same tests with normal inference tests, it is reasonable to manage the test scripts inside the same folder.

mxnet/buildspec-eia.yml
```diff
@@ -0,0 +1,48 @@
...
+        source: docker/artifacts/config.properties
+        target: config.properties
+      entrypoint:
+        source: docker/artifacts/mms-entrypoint.py
+        target:  mms-entrypoint.py
+
+  images:
+    BuildEIAMXNetInferencePy3DockerImage:
+      <<: *INFERENCE_REPOSITORY
+      build: &MXNET_CPU_INFERENCE_PY3 false
```
YYStreet, line 29: If a change this line from CPU to EIA, the build will not happen. Weird and have not find root cause

mxnet/buildspec-eia.yml
```diff
@@ -0,0 +1,48 @@
...
+      image_size_baseline: 5000
+      device_type: &DEVICE_TYPE eia
```
TusharKanekiDey, line 31: Please change the device type to cpu.

TusharKanekiDey:
Can u update TF-1 Dockerfile as a part of this PR to include the latest version of awscli or in separate PR ?

mxnet/buildspec-eia.yml
YYStreet: I will modify on it. Thanks
@saimidu Could you also confirm about it? This is a sample URI I am seeing in beta account:
```
841569659894.dkr.ecr.us-east-1.amazonaws.com/beta-eia-tensorflow-inference:2.0.0-py2-eia-build-2020-04-22-21-21-05
```

YYStreet:
> Can u update TF-1 Dockerfile as a part of this PR to include the latest version of awscli or in separate PR ?

Will have another PR to address it. Do you think it is a good practice to not pin the exact version, but leave it blank or take a restriction to fetch the latest version of awscli 1.x?

TusharKanekiDey:
> > Can u update TF-1 Dockerfile as a part of this PR to include the latest version of awscli or in separate PR ?
> 
> Will have another PR to address it. Do you think it is a good practice to not pin the exact version, but leave it blank or take a restriction to fetch the latest version of awscli 1.x?

With EIA, I think we can pin to the version that is required. I am fine with both the approach.

PR #336, opened 2020-07-06
Updating PT binaries for version 1.6.0

abhinavs95:
This is an automated PR by the PT deployment pipeline aws-pytorch-cicd-pipeline-inference


arjkesh:
Buildspec will also need to be updated if you want to test changes for 1.6.0

PR #351, opened 2020-07-13
[test] | [ec2] feature: Add p3dn tests to all frameworks in us-east-1

saimidu:
*Issue #, if available:*

## Checklist

*Description:*
p3dn.24xlarge instances can now be used by tests in the us-east-1 region, with the region parametrized alongside instance type for EC2 tests.

*Tests run:*
Locally ran EC2 tests on TensorFlow 1.15 training image

*DLC image/dockerfile:*
N/A

*Additional context:*
N/A


arjkesh:
Thoughts on a simpler implementation that does not require region and ec2_instance_type to be dependent parameters -- inside the ec2_instance fixture, you have access to the value of region and ec2_instance_type. Why not just set region to p3dn region if ec2_instance_type is p3dn?

test/test_utils/sagemaker.py
```diff
@@ -200,7 +200,7 @@ def execute_local_tests(image, ec2_client):
     try:
         key_file = generate_ssh_keypair(ec2_client, ec2_key_name)
         print(f"Launching new Instance for image: {image}")
-        instance_id, ip_address = launch_sagemaker_local_ec2_instance(image, UBUNTU_16_BASE_DLAMI, ec2_key_name, region)
+        instance_id, ip_address = launch_sagemaker_local_ec2_instance(image, UBUNTU_16_BASE_DLAMI_US_WEST_2, ec2_key_name, region)
```
arjkesh, line 203: can we conditionally set this AMI? I see we are getting region from AWS_REGION in line 197, we can use that to set the correct AMI

test/test_utils/__init__.py
```diff
@@ -306,7 +310,7 @@ def generate_ssh_keypair(ec2_client, key_name):
             if os.path.exists(key_filename):
                 run(f"chmod 400 {key_filename}")
                 return key_filename
-        raise ClientError(e)
+        raise e
```
arjkesh, line 313: good catch

test/test_utils/ec2.py
```diff
@@ -38,7 +38,7 @@ def get_ec2_instance_type(default, processor, enable_p3dn=False):
         )
     instance_type = os.getenv(f"EC2_{processor.upper()}_INSTANCE_TYPE", default)
     if instance_type == p3dn and not enable_p3dn:
-        return [default]
+        instance_type = default
```
arjkesh, line 41: is part of the scope of this PR to enable all the p3dn tests? if so, how about we make enable_p3dn=True by default in the function, since it seems the new default behavior should have p3dns enabled?

test/test_utils/sagemaker.py
saimidu: We could. I'll make the change.

test/test_utils/ec2.py
saimidu: I think this is better because we want to affirmatively enable p3dn instance usage through a PR rather than incidentally enabling p3dn instance usage through a random change.

Our focus seems to be more on preventing p3dn instances from being used unnecessarily than allowing the use of p3dn instances on all tests.

arjkesh: the initial reason behind having enable_p3dn as an option was because p3dn testing was disabled by default due to resource issues. If this fixes the resource issues, I think we should no longer make p3dn an opt-in, to avoid mistakenly not enabling

arjkesh: Given that we are now empowered to use p3dns, and your concerns about unnecessary p3dn usage, I think an implementation like the following might make sense:

1) Change the semantics to "disable_p3dn" -- this way we are communicating that p3dn testing is enabled by default
2) never allow p3dn to be the "default" instance type -- i.e., raise error if someone uses p3dn as the default instance type -- this will help avoid unnecessary p3dn usage in PR context.

saimidu: 1 make sense.
2 seems unnecessary. We might have some tests where a p3dn instance is the best to use.

PR #353, opened 2020-07-13
Set awscli version up-to-date

YYStreet:
*Issue #, if available:*

## Checklist
- [x] I've prepended PR tag with frameworks/job this applies to : [mxnet, tensorflow, pytorch] | [build] | [test] | [build, test]

*Description:*
Set awscli version up-to-date
*Tests run:*

*DLC image/dockerfile:*

*Additional context:*

### DIFF
--- a/pytorch/inference/docker/1.3.1/py3/Dockerfile.eia
+++ b/pytorch/inference/docker/1.3.1/py3/Dockerfile.eia
@@ -95,7 +95,7 @@ RUN curl https://aws-dlc-licenses.s3.amazonaws.com/pytorch/license.txt -o /licen
 RUN conda install -y -c conda-forge pyyaml==5.3.1 \
  && pip install -U \
     sagemaker-containers==2.8.6 \
-    awscli
+    "awscli<2"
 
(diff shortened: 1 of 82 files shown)


arjkesh:
Just for clarification, why do we need to ensure awscli is less than 2 for all images? Once awscli starts releasing awscli2 on pypi, won't we want to consume from that, at least from py3 images?

YYStreet:
> Just for clarification, why do we need to ensure awscli is less than 2 for all images? Once awscli starts releasing awscli2 on pypi, won't we want to consume from that, at least from py3 images?

This is just out of my personal decision. Awscli2 is existing and just not available on pypi. I am not sure if awscli2 is  completely backwards compatible with awscli1 or whether there could be issues emerging, and add this restriction in case. 
However I also agree with you that we might be able to consume awscli2 from a certain day. In addition, we should believe that awscli team would guarantee the compatibility and any other stuff before they release it on pypi.
To leave it with no restriction is also acceptable to me.
Which one are you preferring? I can make the change

PR #357, opened 2020-07-14
[tensorflow][ec2][test]Use native tf.keras to run the test

Elizaaaaa:
*Description:*
Current keras test is: clone the keras-team/keras.git, replace the import keras with import tensorflow.keras as keras and run the example/mnist_cnn.py. Since keras has become a part of the tensorflow api, directly running test with native tf.keras will be a better approach. 

Adding the tensorboard profiler when training. Will combine the keras and tensorboard tests.

*Tests run:*

*DLC image/dockerfile:*

*Additional context:*


test/dlc_tests/ec2/tensorflow/training/test_tensorflow_training.py
```diff
@@ -36,13 +36,17 @@ def test_tensorflow_standalone_cpu(tensorflow_training, ec2_connection, cpu_only
 @pytest.mark.model("mnist")
 @pytest.mark.parametrize("ec2_instance_type", TF_EC2_GPU_INSTANCE_TYPE, indirect=True)
 def test_tensorflow_train_mnist_gpu(tensorflow_training, ec2_connection, gpu_only):
+    if is_tf1(tensorflow_training):
```
YYStreet, line 39: It should be better to remain the original approach for tf1 and use the new approach for tf2, as 2 different tests.
We can merge in this one and open another one for recover original tf1 test

YYStreet:
Please take a look of TF1 failure. The logic of tf1 might not be fully reverted back.

Elizaaaaa:
> Please take a look of TF1 failure. The logic of tf1 might not be fully reverted back.

fixed it. currently the tf1-ec2 failed with TF1HVD, tf2-ec2 failed with telemetry test.

PR #361, opened 2020-07-16
[test]|[sagemaker] Add test report generator logic to SageMaker pytest frameworks

arjkesh:
*Issue #, if available:*

## Checklist

*Description:*
- Refactor test generator logic to be reusable
- Add test generator logic to each individual SageMaker pytest framework

*Tests run:*
- Ran integration test
- Tested manually generating the report with the following command
```
export DLC_IMAGES='test' && pytest -s pytest -s sanity/test_generate_coverage_doc.py
```

*Additional context:*
- Needed to install all the requirements for each sagemaker repo in order to use the framework
- We still make boto3 calls to determine whether an instance is single gpu or not
- Not sure how long the test takes if run for the first time, but with requirements installed, it takes roughly 30 seconds to generate the report (downside of making boto3 calls and installing requirements each time)


arjkesh:
note: I did not register the model/integration/multinode markers -- will follow up with this in the next revision

test/test_utils/test_reporting.py
```diff
@@ -0,0 +1,270 @@
...
+    markers = [mark for mark in item_obj.iter_markers(name=marker_name)]
+    if not markers:
+        if not default:
+            raise RequiredMarkerNotFound(f"PyTest Marker {marker_name} is required on function {item_obj.name}")
+        return default
+    else:
+        return markers[0].args[0]
+
+
+def generate_sagemaker_reports():
```
arjkesh, line 159: this function is new

test/test_utils/test_reporting.py
```diff
@@ -0,0 +1,270 @@
...
+    # Handle local tests
+    with ctx.cd(tf_inf_path):
+        ctx.run("pytest -s --collect-only --generate-coverage-doc --framework-version 2 local/")
+
+    # Handle remote integration tests
+    with ctx.cd(tf_inf_path):
+        ctx.run("pytest -s --collect-only --generate-coverage-doc sagemaker/")
+
+
+def generate_coverage_doc(items, sagemaker=False, framework=None, job_type=None):
```
arjkesh, line 190: this is a generalization for what used to be in pytest_collection_modifyitems

test/sagemaker_tests/tensorflow/inference/test/integration/local/conftest.py
```diff
@@ -13,7 +13,8 @@
 
 import pytest
 
-FRAMEWORK_LATEST_VERSION = '1.13'
+
+FRAMEWORK_LATEST_VERSION = '2.2'
```
nskool, line 17: If this is being updated in this PR, did it impact previous inference tests?

arjkesh: No, it shouldn't have. We are explicitly calling the "--framework_version" command. I can probably remove this and make framework_version a required argument to avoid confusion here

test/test_utils/test_reporting.py
```diff
@@ -0,0 +1,290 @@
...
+class TestReportGenerator:
+    """
+    Class to generate test report files
+    """
+
+    ALLOWED_SINGLE_GPU_TESTS = ("telemetry", "test_framework_version_gpu")
+    SM_REPOS = (
+        "pytorch-training",
+        "pytorch-inference",
+        "tensorflow-tensorflow1_training",
```
nskool, line 51: Why are these tensorflow-tensorflow1_training?

test/test_utils/test_reporting.py
```diff
@@ -0,0 +1,290 @@
...
+    else:
+        return markers[0].args[0]
+
+
+class TestReportGenerator:
+    """
+    Class to generate test report files
+    """
+
+    ALLOWED_SINGLE_GPU_TESTS = ("telemetry", "test_framework_version_gpu")
```
nskool, line 47: Curious: How does it help to define these as class variables? Is it a convention to capitalize? Was it done in response to an earlier comment to make these tuples more visible?

test/test_utils/test_reporting.py
```diff
@@ -0,0 +1,290 @@
...
+    SM_REPOS = (
+        "pytorch-training",
+        "pytorch-inference",
+        "tensorflow-tensorflow1_training",
+        "tensorflow-tensorflow2_training",
+        "mxnet-training",
+        "mxnet-inference",
+    )
+    COVERAGE_DOC_EXECUTABLE = "pytest -s --collect-only --generate-coverage-doc"
```
nskool, line 56: Nit: change `executable` to `command`

test/test_utils/test_reporting.py
```diff
@@ -0,0 +1,290 @@
...
+            for test in allowed_single_gpu:
+                if test in function_key:
+                    whitelist_single_gpu = True
+                    break
+            if num_gpus == 1:
+                processor = "single_gpu"
+                if not whitelist_single_gpu:
+                    single_gpu_failure_message = (
+                        f"Function uses single-gpu instance type {instance_type}. "
+                        f"Please use multi-gpu instance type."
```
nskool, line 104: It'll also help to add a sentence here that informs that the test can be added to the 'allowed list'.

test/test_utils/test_reporting.py
```diff
@@ -0,0 +1,290 @@
...
+    If the default does not exist and the option does not exist, raise an error.
+    :param item_obj: pytest item object
+    :param marker_name: name of the pytest marker
+    :param default: default return value -- if None, assume this is a required marker
+    :return: First arg value for the marker or the default value
+    """
+    markers = [mark for mark in item_obj.iter_markers(name=marker_name)]
+    if not markers:
+        if not default:
+            raise RequiredMarkerNotFound(f"PyTest Marker {marker_name} is required on function {item_obj.name}")
```
nskool, line 36: Nice.

test/test_utils/test_reporting.py
```diff
@@ -0,0 +1,290 @@
...
+        with open(self.test_coverage_file, file_open_type) as tc_file:
+            writer = csv.DictWriter(tc_file, delimiter=",", fieldnames=field_names)
+            if file_open_type == "w+":
+                writer.writeheader()
+
+            for _func_key, info in test_coverage_info.items():
+                writer.writerow(info)
+
+    @staticmethod
+    def _infer_field_value(default, options, *comparison_str):
```
nskool, line 157: Why is this a staticmethod?

test/test_utils/test_reporting.py
arjkesh: These are all constants that are are associated with this class -- AFAIK there is no specific naming convention for class variables, but I put them as all caps because they are essentially constants associated with the class.

I can move this outside of the class if you think that's useful

test/test_utils/test_reporting.py
arjkesh: Oops, looks like my editor changed this on me because I didn't need to use any instance attributes/methods in this method. It's not exactly a static method, but also not exactly a general purpose function. I can move this to be just a private function outside of the class

test/test_utils/test_reporting.py
arjkesh: Good catch

test/test_utils/test_reporting.py
arjkesh: We have separate tf1 and tf2 training repos for SageMaker. I think this might be more clear if I make these paths vs dash separated strings

test/test_utils/test_reporting.py
nskool: No preference, I'm not aware of any convention here. So it should be alright.

PR #375, opened 2020-07-18
cleanup oidc provider created by kubeflow

tejaschumbalkar:
*Issue #, if available:*

## Checklist
- [ ] I've prepended PR tag with frameworks/job this applies to : [mxnet, tensorflow, pytorch] | [build] | [test] | [build, test] | [ec2, ecs, eks, sagemaker]

*Description:*

Delete the  oidc provider created by kubeflow after testing is done.

*Tests run:*

Tested locally

*DLC image/dockerfile:*

*Additional context:*


test/test_utils/eks.py
```diff
@@ -215,10 +228,29 @@ def delete_cfn_stack_and_wait(stack_name):
         LOGGER.error(f"Error: Cannot delete stack: {stack_name}. Full Exception:\n{e}")
         describe_cfn_stack_events(stack_name)
 
+def delete_oidc_provider(eks_cluster_name):
+    """Function to delete the oidc provider created by kubeflow
+    Args:
+        eks_cluster_name: str
+    """
+    eks_client = init_eks_client()
+    iam_client = init_iam_client()
```
saimidu, line 237: nit: Why have separate functions for these? It seems kind of redundant, because it's still a one-line function call.

not nit: Please create the EKS client with a region (afaik, eks clients are regional, right?), with the region set to test_utils.REGION by default, and any other region if an override is required. We want to make the code region-agnostic as far as possible in case we'd like to run tests in multiple regions in the future.

test/test_utils/eks.py
```diff
@@ -215,10 +228,29 @@ def delete_cfn_stack_and_wait(stack_name):
...
+    account_id = os.getenv("ACCOUNT_ID")
```
saimidu, line 238: Could you please use boto3 APIs to find the account ID instead?

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sts.html#STS.Client.get_caller_identity

We can never be 100% sure that an environment variable will always be populated unless it is a "necessary" value for a previous section in the same workflow. It's always best to have a default value for all `getenv()` calls. In this case, we don't actually need environment variables as the account ID can be identified through boto3 APIs.

test/test_utils/eks.py
```diff
@@ -215,10 +228,29 @@ def delete_cfn_stack_and_wait(stack_name):
...
+    """Function to delete the oidc provider created by kubeflow
+    Args:
+        eks_cluster_name: str
+    """
+    eks_client = init_eks_client()
+    iam_client = init_iam_client()
+    account_id = os.getenv("ACCOUNT_ID")    
+
+    try:
+        response = eks_client.describe_cluster(name=eks_cluster_name)
```
saimidu, line 241: Wouldn't you have already deleted the cluster at this point in the `delete_eks_cluster` function?

test/test_utils/eks.py
```diff
@@ -215,10 +228,29 @@ def delete_cfn_stack_and_wait(stack_name):
...
+        oidc_issuer = response['cluster']['identity']['oidc']['issuer']
+        oidc_url = oidc_issuer.rsplit('//', 1)[-1]
+        oidc_provider_arn = f"arn:aws:iam::{account_id}:oidc-provider/{oidc_url}"
+
+        LOGGER.info(f"Deleting oidc provider: {oidc_provider_arn}")
+        iam_client.delete_open_id_connect_provider(OpenIDConnectProviderArn=oidc_provider_arn)
+    except ClientError as e:
+        LOGGER.error(f"Error: Cannot delete the EKS cluster: {eks_cluster_name}. Full Exception:\n{e}")
```
saimidu, line 249: Could you make this LOGGER.error message more specific, and say that the OIDC provider could not be deleted, followed by the arn of the provider that couldn't be deleted?

test/test_utils/eks.py
TusharKanekiDey: +1

tejaschumbalkar: Modified accordingly

test/test_utils/eks.py
tejaschumbalkar: uses boto3 sts client

test/test_utils/eks.py
tejaschumbalkar: True, I was waiting for CDK changes to catch this. Made the changes

test/test_utils/eks.py
tejaschumbalkar: Delete oidc provider call is idempotent; it does not fail or return an error if you call the operation for a provider that does not exist.

PR #403, opened 2020-07-24
[test] [sanity] Sanity check for dlc_major_version label

saimidu:
*Issue #, if available:*

## Checklist

*Description:*
Inserted a test that asserts the presence of a LABEL on every image that has been built, named "dlc_major_version", without which the test fails.

*Tests run:*
Tested locally on laptop with 1 failing image and 1 previously released successful image. Both work as expected.

*DLC image/dockerfile:*
N/A

*Additional context:*
N/A


test/dlc_tests/sanity/test_dlc_labels.py
```diff
@@ -0,0 +1,33 @@
...
+    ecr_client = boto3.client("ecr", region_name=region)
+
+    image_repository, image_tag = get_repository_and_tag_from_image_uri(image)
+    response = ecr_client.batch_get_image(
+        repositoryName=image_repository,
+        imageIds=[{"imageTag": image_tag}],
+        acceptedMediaTypes=["application/vnd.docker.distribution.manifest.v1+json"],
+    )
+    if not response.get("images"):
+        raise ValueError(
```
arjkesh, line 19: why is one of these a value error and the other a key error?

test/dlc_tests/sanity/test_dlc_labels.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def test_dlc_major_version_label(image, region):
+    ecr_client = boto3.client("ecr", region_name=region)
+
+    image_repository, image_tag = get_repository_and_tag_from_image_uri(image)
+    response = ecr_client.batch_get_image(
+        repositoryName=image_repository,
+        imageIds=[{"imageTag": image_tag}],
+        acceptedMediaTypes=["application/vnd.docker.distribution.manifest.v1+json"],
```
arjkesh, line 16: can you provide more info on why this is needed?

test/dlc_tests/sanity/test_dlc_labels.py
```diff
@@ -0,0 +1,33 @@
+import json
+
+import boto3
+import pytest
+
+from test.test_utils import get_repository_and_tag_from_image_uri, LOGGER
+
+
+def test_dlc_major_version_label(image, region):
```
arjkesh, line 9: nit: add docstring

also you can add a marker for the integration `@pytest.mark.integration('dlc_major_version_label')` or something

saimidu: We didn't have markers on other sanity tests, so didn't add one here. Can add.

test/dlc_tests/sanity/test_dlc_labels.py
saimidu: The first is an error to be raised when the "images" were not obtained through the ecr client request. I thought that semantically, it would be a missing value, and hence, ValueError.

The other is a key error because values were obtained through the ecr client request, but the specific attribute we need -- imageManifest -- is missing, hence the KeyError to denote a missing dictionary key.

Please recommend any changes that would better represent the errors.

test/dlc_tests/sanity/test_dlc_labels.py
saimidu: Adding comment.

Basically this is a format in which the response needs to be sent, so that the image LABELS can be found on the manifest.

test/dlc_tests/sanity/test_dlc_labels.py
arjkesh: I think they should both be key errors -- think of it this way, if you didn't use .get, you'd get key errors in both scenarios

arjkesh:
can you update the tests run section?

saimidu:
Done. Thanks @arjkesh

PR #410, opened 2020-07-24
[Pytorch][Build]Addition of curand.h in pytorch

TusharKanekiDey:
*Issue #, if available:*

## Checklist

*Description:*
Addition of curand package for Pytorch
*Tests run:*

*DLC image/dockerfile:*

*Additional context:*


arjkesh:
Do we have a test for these changes? Also w.r.t the diff, I don't see any addition of the .h file.

TusharKanekiDey:
> Do we have a test for these changes? Also w.r.t the diff, I don't see any addition of the .h file.

Sure, I can add a test for that. 
I have manually test the image and check the required package is available.
We do not need to add any file, we just need to make sure that cuda-curand-dev-10-1 is not removed.

arjkesh:
> > Do we have a test for these changes? Also w.r.t the diff, I don't see any addition of the .h file.
> 
> Sure, I can add a test for that.
> I have manually test the image and check the required package is available.
> We do not need to add any file, we just need to make sure that cuda-curand-dev-10-1 is not removed.

From this PR though, it looks like cuda-curand-dev-10-1 is being removed. Is this expected?

TusharKanekiDey:
> > > Do we have a test for these changes? Also w.r.t the diff, I don't see any addition of the .h file.
> > 
> > 
> > Sure, I can add a test for that.
> > I have manually test the image and check the required package is available.
> > We do not need to add any file, we just need to make sure that cuda-curand-dev-10-1 is not removed.
> 
> From this PR though, it looks like cuda-curand-dev-10-1 is being removed. Is this expected?

To reduce the size, cuda-curand-dev-10-1 and other similar packages have been removed,

test/dlc_tests/container_tests/bin/pytorch_tests/testCurand
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+if ls /usr/local/cuda/include | grep curand; then
```
arjkesh, line 3: is it a file called curand that we're looking for, or curand.h? are there multiple files and is there a way we can be deliberate about looking for them?

test/dlc_tests/ec2/pytorch/training/test_pytorch_training.py
```diff
@@ -96,3 +97,8 @@ def test_pytorch_telemetry_gpu(pytorch_training, ec2_connection, gpu_only):
 @pytest.mark.parametrize("ec2_instance_type", ["c5.4xlarge"], indirect=True)
 def test_pytorch_telemetry_cpu(pytorch_training, ec2_connection, cpu_only):
     execute_ec2_training_test(ec2_connection, pytorch_training, PT_TELEMETRY_CMD)
+
+
+@pytest.mark.parametrize("ec2_instance_type", ["p2.xlarge"], indirect=True)
```
arjkesh, line 102: nice, thanks for adding. Out of curiosity, is this same change needed in other containers, or is it already present?

reason for asking - if already present, we could make this test framework agnostic.

pytorch/training/docker/1.4.0/py2/Dockerfile.gpu
```diff
@@ -51,7 +51,6 @@ RUN apt-get update \
     wget \
     zlib1g-dev \
  && apt-get remove -y cuda-cufft-dev-10-1 \
-    cuda-curand-dev-10-1 \
```
arjkesh, line 54: just want to double check -- is this change not necessary in inference containers?

test/dlc_tests/ec2/pytorch/training/test_pytorch_training.py
TusharKanekiDey: It is present in TF2.2 only till now.

test/dlc_tests/container_tests/bin/pytorch_tests/testCurand
TusharKanekiDey: Yes, It is looking for a particular package known as curand.h. I will grep curand.h instead of curand in the next iteration.

pytorch/training/docker/1.4.0/py2/Dockerfile.gpu
TusharKanekiDey: We need it for training images only.

test/dlc_tests/ec2/pytorch/training/test_pytorch_training.py
arjkesh: Perhaps then we can add the test inside `test/dlc_tests/ec2/`, kind of like smdebug? Do you anticipate this test should be required to pass for every GPU training container?

arjkesh: if it currently fails for mxnet and tf1, for instance, we can skip for those images

arjkesh: note: you now need to add model as a required marker.

You can decorate this function as follows:

```
@pytest.mark.model("N/A")
```

arjkesh:
The single gpu test still needs to be whitelisted, else tests will continue to fail

TusharKanekiDey:
> The single gpu test still needs to be whitelisted, else tests will continue to fail

Oh, I though this line (`ALLOWED_SINGLE_GPU_TESTS = ("telemetry", "test_framework_version_gpu", "test_curand_in_gpu")` ) would whitelist it. Will make the required change.

test/test_utils/test_reporting.py
```diff
@@ -43,7 +43,7 @@ class TestReportGenerator:
     Class to generate test report files
     """
 
-    ALLOWED_SINGLE_GPU_TESTS = ("telemetry", "test_framework_version_gpu")
+    ALLOWED_SINGLE_GPU_TESTS = ("telemetry", "test_framework_version_gpu", "test_curand_in_gpu")
```
arjkesh, line 46: this needs to be the name of the function, which is "test_curand_gpu"

test/dlc_tests/ec2/test_curand.py
```diff
@@ -0,0 +1,17 @@
...
+
+import pytest
+
+from test.test_utils import CONTAINER_TESTS_PREFIX, is_tf1
+from test.test_utils.ec2 import execute_ec2_training_test
+
+CURAND_CMD = os.path.join(CONTAINER_TESTS_PREFIX, "testCurand")
+
+
+@pytest.mark.integration("test_curand_in_gpu")
```
arjkesh, line 11: you can just leave integration as curand

test/dlc_tests/ec2/test_curand.py
```diff
@@ -0,0 +1,17 @@
...
+from test.test_utils import CONTAINER_TESTS_PREFIX, is_tf1
+from test.test_utils.ec2 import execute_ec2_training_test
+
+CURAND_CMD = os.path.join(CONTAINER_TESTS_PREFIX, "testCurand")
+
+
+@pytest.mark.integration("curand")
+@pytest.mark.model("N/A")
+@pytest.mark.parametrize("ec2_instance_type", ["p2.xlarge"], indirect=True)
+def test_curand_gpu(training, ec2_connection, gpu_only):
```
arjkesh, line 14: nit: you could add more description here, but this should suffice

PR #418, opened 2020-07-27
[build]|[test] Ensure cuda version is in the Dockerfile path, add sanity tests for cuda versions

arjkesh:
*Issue #, if available:*

## Checklist

*Description:*
- Update filepaths for gpu Dockerfiles to include cuda version, also add tests

*Tests run:*
- Added new sanity integration tests to:
a) test that the cuda version in the tag is also the same as the cuda version in the container
b) check that gpu image has an associated dockerfile in a path with the cu version
c) ensure the tagged cuda version is also the same as the one in the buildspec

*DLC image/dockerfile:*
All gpu dockerfiles


test/dlc_tests/sanity/test_pre_release.py
```diff
@@ -197,6 +208,53 @@ def test_emacs(image):
...
+            break
+    assert framework, f"Cannot find any frameworks {frameworks} in image uri {image}"
+
+    cuda_version_match = re.search(r'-cu\d+-', image)
+    cuda_version = cuda_version_match.group().strip('-')
+
+    framework_version_match = re.search(r':\d+.\d+.\d+', image)
+    framework_version = framework_version_match.group().strip(':')
+
+    python_version = "py3" if "py3" in image else "py2"
```
YYStreet, line 232: python_version should also take "py36" into consideration

arjkesh: good point, will update

test/dlc_tests/sanity/test_pre_release.py
```diff
@@ -197,6 +208,53 @@ def test_emacs(image):
     _run_cmd_on_container(container_name, ctx, "emacs -version")
 
 
+@pytest.mark.model("N/A")
+def test_cuda_paths(gpu):
```
arjkesh, line 212: adding a docstring in the next revision

test/dlc_tests/sanity/test_pre_release.py
```diff
@@ -197,6 +208,71 @@ def test_emacs(image):
...
+    """
+    Test to ensure directory structure for GPU files has cuda version in it
+
+    :param gpu: gpu image uris
+    """
```
saimidu, line 217: Could you change the function description in the docstring to
"Test to ensure directory structure for GPU **docker**files has cuda version in it"

test/dlc_tests/sanity/test_pre_release.py
```diff
@@ -197,6 +208,71 @@ def test_emacs(image):
...
+    image = gpu
+    if "example" in image:
+        pytest.skip("Skipping example containers which are not explicitly tied to a cuda version")
```
saimidu, line 220: Please change to "Skipping example **dockerfiles**~~containers~~ which are not explicitly tied to a cuda version"

test/dlc_tests/sanity/test_pre_release.py
```diff
@@ -197,6 +208,71 @@ def test_emacs(image):
...
+
+    dlc_path = os.getcwd().split('/test/')[0]
+    job_type = "training" if "training" in image else "inference"
```
saimidu, line 223: Please run `black` or something to standardize the entire file to PEP8 standards.

This particular comment is about the use of `'<str>'` vs `"<str>"`.

test/dlc_tests/sanity/test_pre_release.py
```diff
@@ -197,6 +208,71 @@ def test_emacs(image):
...
+        if fw in image:
+            framework = fw
+            break
+    assert framework, f"Cannot find any frameworks {frameworks} in image uri {image}"
+
+    # Get cuda and framework version through regex
+    cuda_version_match = re.search(r'-cu\d+-', image)
+    cuda_version = cuda_version_match.group().strip('-')
+
+    framework_version_match = re.search(r':\d+.\d+.\d+', image)
```
saimidu, line 238: nit: Can use regular expression `r":(\d+(.\d+){2})"`, and get the value of `fw_version_match.group(1)`

test/dlc_tests/sanity/test_pre_release.py
```diff
@@ -197,6 +208,71 @@ def test_emacs(image):
...
+    framework = ''
+    for fw in frameworks:
+        if fw in image:
+            framework = fw
+            break
+    assert framework, f"Cannot find any frameworks {frameworks} in image uri {image}"
+
+    # Get cuda and framework version through regex
+    cuda_version_match = re.search(r'-cu\d+-', image)
+    cuda_version = cuda_version_match.group().strip('-')
```
saimidu, line 236: Generally, you don't have to write two lines of code to do a regex query on a string, especially if all you want is the value you are looking for (as opposed to all the other information on the `re.Match` object).

You can do:
```
cuda_version = re.search(...).group().strip("-")
```

Only downside to writing this in a single line is when your search/match fails, and you get a `NoneType` output instead of `re.Match` from the `re.search` function call. In that case, you would want to throw an Exception anyway.

Also, you can use groups, so that you wouldn't need to use the `.strip("-")` here.

test/dlc_tests/sanity/test_pre_release.py
```diff
@@ -197,6 +208,71 @@ def test_emacs(image):
...
+                cuda_in_buildspec = True
+                break
+
+    try:
+        assert cuda_in_buildspec, f"Can't find {cuda_in_buildspec_ref} in {buildspec_path}"
+    except AssertionError as e:
+        if not is_dlc_cicd_context():
+            LOGGER.warn(f"{e} - not failing, as this is a(n) {os.getenv('BUILD_CONTEXT', 'empty')} build context.")
+        else:
+            raise
```
saimidu, line 268: Why would we not want this Error to be raised outside of cicd_context?

It seems like the sort of test that you would always want, and can easily skip if you're an "advanced" enough user that you're building your own images and running DLC tests.

test/dlc_tests/sanity/test_pre_release.py
```diff
@@ -197,6 +208,71 @@ def test_emacs(image):
...
+
+    # Check that a Dockerfile exists in the right directory
+    dockerfile_path = os.path.join(framework_version_path, python_version, cuda_version, 'Dockerfile.gpu')
+
+    assert os.path.exists(dockerfile_path), f"Cannot find dockerfile for image {image} in {dockerfile_path}"
```
saimidu, line 273: What is the rationale behind not throwing the previous AssertionError but allowing it to happen here?

Is it that we want both tests to be run by a user running it outside of the CI/CD?

In that case, doesn't it make sense that both of these features are tested on the CI/CD as well, rather than having to fail the test once, fix one issue, try the test again, fail again, then fix the second issue?

test/dlc_tests/sanity/test_pre_release.py
arjkesh: I want to avoid doing too much string manipulation in one line for readability.

test/dlc_tests/sanity/test_pre_release.py
arjkesh: If you are testing older images locally, this will fail. Wanted to avoid causing unnecessary failures and warn instead

test/dlc_tests/sanity/test_pre_release.py
arjkesh: No, this assertion is checking that the dockerfile itself exists. The latest dockerfile is not a prereq, hence it should always fail.

test/dlc_tests/sanity/test_pre_release.py
arjkesh: sure, will do

test/dlc_tests/sanity/test_pre_release.py
```diff
@@ -157,6 +158,16 @@ def test_framework_version_gpu(gpu, ec2_connection):
     else:
         assert tag_framework_version == output.stdout.strip()
 
+    # Check cuda version
+    cuda_version_match = re.search(r'-cu\d+-', image)
+    cuda_version = cuda_version_match.group().split('cu')[-1].strip('-')
+    full_cuda_version = f"{cuda_version[:2]}.{cuda_version[2:]}"
```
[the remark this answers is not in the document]
arjkesh: sure, will make the change

test/dlc_tests/sanity/test_pre_release.py
saimidu: Sure. Makes sense to do that for readability.

Please do consider using regex match-groups, though.

test/dlc_tests/sanity/test_pre_release.py
saimidu: Smart.

test/dlc_tests/sanity/test_pre_release.py
saimidu: Makes sense. Thanks.

tensorflow/buildspec.yml
```diff
@@ -81,7 +81,7 @@
       python_version: &DOCKER_PYTHON_VERSION py3
       tag_python_version: &TAG_PYTHON_VERSION py37
       tag: !join [ *VERSION, "-", *DEVICE_TYPE, "-", *TAG_PYTHON_VERSION, "-", *CUDA_VERSION, "-", *OS_VERSION ]
-      docker_file: !join [ docker/, *VERSION, /, *DOCKER_PYTHON_VERSION, /Dockerfile., *DEVICE_TYPE ]
+      docker_file: !join [ docker/, *VERSION, /, *DOCKER_PYTHON_VERSION, /, *CUDA_VERSION, /Dockerfile., *DEVICE_TYPE ]
```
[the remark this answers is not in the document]
arjkesh: I'm not sure, but I kept as * because I figure it could be a reasonable possibility if we need to build cu101/cu102 images simulaneously

test/dlc_tests/sanity/test_pre_release.py
```diff
@@ -157,6 +158,16 @@ def test_framework_version_gpu(gpu, ec2_connection):
     else:
         assert tag_framework_version == output.stdout.strip()
 
+    # Check cuda version
+    cuda_version_match = re.search(r'-cu\d+-', image)
+    cuda_version = cuda_version_match.group().split('cu')[-1].strip('-')
```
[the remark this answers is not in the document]
arjkesh: made change

test/dlc_tests/sanity/test_pre_release.py
arjkesh: updated

test/dlc_tests/sanity/test_pre_release.py
```diff
@@ -211,49 +219,45 @@ def test_emacs(image):
...
-    # Get cuda and framework version through regex
-    cuda_version_match = re.search(r'-cu\d+-', image)
-    cuda_version = cuda_version_match.group().strip('-')
-
-    framework_version_match = re.search(r':\d+.\d+.\d+', image)
-    framework_version = framework_version_match.group().strip(':')
+    # Get cuda, framework version, python version through regex
+    cuda_version = re.search(r"-(cu\d+)-", image).group(1)
+    framework_version = re.search(r":(\d+(.\d+){2})", image).group(1)
+    python_version = re.search(r"(py\d+)", image).group(1)
```
saimidu, line 245: nit: For python_version, you don't need to use groups, though there is no harm in using it.

test/dlc_tests/sanity/test_pre_release.py
```diff
@@ -150,26 +150,31 @@ def test_framework_and_cuda_version_gpu(gpu, ec2_connection):
...
+        # Module name is "torch"
+        if tested_framework == "pytorch":
+            tested_framework = "torch"
+        cmd = f"import {tested_framework}; print({tested_framework}.__version__)"
+        output = ec2.execute_ec2_training_test(ec2_connection, image, cmd, executable="python")
+
+        if is_canary_context():
+            assert tag_framework_version in output.stdout.strip()
+        else:
+            assert tag_framework_version == output.stdout.strip()
```
saimidu, line 167: Question: Why is there a difference in the output of canary vs others?

test/dlc_tests/sanity/test_pre_release.py
```diff
@@ -132,30 +141,43 @@ def test_framework_version_cpu(cpu):
...
+
+    # CUDA Version Check #
+    cuda_version = re.search(r"-cu(\d+)-", image).group(1)
+
+    # MXNet inference containers do not currently have nvcc in $PATH, so check cuda directories
+    if "mxnet-inference" in image:
+        cuda_cmd = "bash -c 'ls /usr/local/cuda*'"
```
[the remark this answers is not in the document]
arjkesh: I purposefully left it this way because this would require interpolating the period inside the cuda version, which previously you were not comfortable with. Do you want me to revert the behavior?

test/dlc_tests/sanity/test_pre_release.py
arjkesh: Canary images generally take the major.minor version instead of major.minor.patch due to the fact that they use SM tags. This was not part of this revision, i just moved all of the framework version test part into an if condition so it looks like i added this

test/dlc_tests/sanity/test_pre_release.py
```diff
@@ -150,26 +150,31 @@ def test_framework_and_cuda_version_gpu(gpu, ec2_connection):
...
+        else:
+            assert tag_framework_version == output.stdout.strip()
 
-    # Check cuda version
+    # CUDA Version Check #
     cuda_version = re.search(r"-cu(\d+)-", image).group(1)
-    cuda_cmd = "nvcc --version"
-    cuda_output = ec2.execute_ec2_training_test(ec2_connection, image, cuda_cmd, container_name="cuda_container")
+
+    # MXNet inference containers do not currently have nvcc in $PATH, so check cuda directories
```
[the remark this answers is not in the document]
arjkesh: it's not in /usr/local/cuda/bin so i believe it is not available. I updated the message

test/dlc_tests/sanity/test_pre_release.py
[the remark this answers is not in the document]
arjkesh: Changed behavior in new revision to use readlink

PR #438, opened 2020-07-31
[tensoflow][benchmark][test] Upgrading tensorflow serving api to 2.2.0

sboshin:
Removing unnecessary binary argument, It does nothing, Removing need to download unncessary binary from S3, fixing deprecated API calls in script. Upgraded serving_api to 2.2

*Issue #, if available:*

## Checklist
- [ ] I've prepended PR tag with frameworks/job this applies to : [mxnet, tensorflow, pytorch] | [build] | [test] | [build, test] | [ec2, ecs, eks, sagemaker]

#### EIA Checklist
* When creating a PR:
- [ ] I've modified ```src/config/build_config.py``` in my PR branch by setting ```ENABLE_EI_MODE = True```

*Description:*

*Tests run:*

*DLC image/dockerfile:*

*Additional context:*


sboshin:
Updated tf1 serving script to behave like tf2, So we don't need to pass in a binary argument.

YYStreet:
> Updated tf1 serving script to behave like tf2, So we don't need to pass in a binary argument.

It seems currently tf1 script is identical with tf2 script. Should this script also be working for TF2.0 and TF2.1 inference?

YYStreet:
> Updated tf1 serving script to behave like tf2, So we don't need to pass in a binary argument.

Here is a script provided by Akihil for TF2.0, with very minor difference https://github.com/aws/deep-learning-containers/blob/bbe004e8b98a1b3ab1b657674ce1a0b492b7acd6/test/dlc_tests/container_tests/bin/benchmark/tf20_serving_perf.py

sboshin:
I think its okay to leave it separated. I don't remember the initial reason why we had it split.
Keeping it split will allow tf2 version to keep growing without impacting tf1 in the future

YYStreet:
> I think its okay to leave it separated. I don't remember the initial reason why we had it split.
> Keeping it split will allow tf2 version to keep growing without impacting tf1 in the future.
> 
> I assume the minor difference came from using tf.contrib vs tf.compat. Its the same implementation just different API calls for compatibility

Could you give us an advice that if we should maintain different tf2 benchmark script for different tf2.x versions? If we still need to maintain the original test script as a copy for tf2.1, I can finish this part in another PR

sboshin:
As the DLC testing is only testing the latest TF which will be 2.x, I don't see why we need a 2.0, 2.1, 2.2 script. And the current script should work for all versions of TF 2.x.
So the only split should be tf1 and tf2.

YYStreet:
> As the DLC testing is only testing the latest TF which will be 2.x, I don't see why we need a 2.0, 2.1, 2.2 script. And the current script should work for all versions of TF 2.x.
> So the only split should be tf1 and tf2.

Thanks. Tests are all passing. I will revert the config change back and approve it.

PR #453, opened 2020-08-04
[MXNet,Pytorch][EIA]Skip the p3dn test in EIA

TusharKanekiDey:
*Issue #, if available:*

## Checklist

#### EIA Checklist
* When creating a PR:
- [x] I've modified ```src/config/build_config.py``` in my PR branch by setting ```ENABLE_EI_MODE = True```
- [x] I've reverted the code change on the config file mentioned above

*Description:*
Trying to skip the p3dn test for PT and MXNet EIA


arjkesh:
@TusharKanekiDey can you add back the EI related information to the checklist, since this is an EI related PR?

test/test_utils/ec2.py
```diff
@@ -37,8 +37,7 @@ def get_ec2_instance_type(default, processor, disable_p3dn=False):
             "Default instance type should never be p3dn.24xlarge"
         )
     instance_type = os.getenv(f"EC2_{processor.upper()}_INSTANCE_TYPE", default)
-    #Not running eia test on p3dn instances
-    if instance_type == p3dn and (disable_p3dn or processor == "eia"):
+    if instance_type == p3dn and disable_p3dn:
```
arjkesh, line 40: is this function used for eia at all? it probably can be instead of get_ec2_accelerator type

TusharKanekiDey: No, This function is used for CPU and GPU processor type.
To Decouple them, we kept get_ec2_accelerator type to fetch the EI accelerator type.
I can combine both of them if needed.

TusharKanekiDey:
> @TusharKanekiDey can you add back the EI related information to the checklist, since this is an EI related PR?

Done, I was planning to add them back when they are ready for review.
Thanks for pointing that out.

test/test_utils/ec2.py
arjkesh: no need to combine in this PR, but probably we should remove "eia" from allowed_processors in line 28 if we are specifically decoupling eia from this function

TusharKanekiDey: Sure, I will remove eia from allowed_processors in  line 28.

PR #466, opened 2020-08-06
[test][ec2] Fail with error messages on SMDebug during timeouts

arjkesh:
*Issue #, if available:*

## Checklist

*Description:*
SMDebug tests often timeout with p2.8xlarge instance types. We do not get logs, however, when this happens. Making updates to enable us to get logs.

Move the timeouts to DLC side, remove extra layer from smdebug tests

*Tests run:*
- PR tests


src/config/build_config.py
```diff
@@ -4,7 +4,7 @@
 # Frameworks for which you want to disable both builds and tests
 DISABLE_FRAMEWORK_TESTS = []
```
TusharKanekiDey, line 5: You can include MXNet and Pytorch in the list to disable their build and test.

YYStreet: Should be Okay to test them all together

test/dlc_tests/ec2/test_smdebug.py
```diff
@@ -16,54 +16,58 @@
...
+    try:
+        test_output = ec2_connection.run(
+            f"{docker_executable} exec --user root {container_name} "
+            f"/bin/bash -c '{test_script} {framework}' | tee {logfile}",
+            hide=True,
+            warn=True,
+            timeout=3000,
+        )
+    except Exception:
+        debug_output = ec2_connection.run(f"cat {logfile}")
```
tejaschumbalkar, line 68: isn't the logfile inside the docker container?

arjkesh: it shouldn't be in this case, basically i'm running the docker exec command and piping it to a file on the host since docker runs on the host

PR #479, opened 2020-08-07
[build, test][tensorflow] Add sanity test to validate versioning scheme

arjkesh:
*Issue #, if available:*

## Checklist

*Description:*
Update the TF2.3 gpu major version, add a sanity test to validate major version scheme

*Tests run:*
Ran newly created sanity test

*DLC image/dockerfile:*
N/A

*Additional context:*


test/dlc_tests/sanity/test_dlc_labels.py
```diff
@@ -42,4 +44,71 @@ def test_dlc_major_version_label(image, region):
...
+        "pytorch": "1.5.0"
+    }
+    if test_utils.is_tf1(image):
+        reference_fw = "tensorflow1"
+    elif test_utils.is_tf2(image):
+        reference_fw = "tensorflow2"
+    else:
+        reference_fw = framework
+    if processor != "eia" and fw_version < references[reference_fw]:
+        pytest.skip("Not enforcing new versioning scheme on old images")
```
arjkesh, line 79: nit: perhaps worth logging the image URI in the skip notification

test/dlc_tests/sanity/test_dlc_labels.py
```diff
@@ -42,4 +44,71 @@ def test_dlc_major_version_label(image, region):
...
+            if version >= 2:
+                version += 1
+            expected_versions.append(version)
+    else:
+        expected_versions = actual_versions
+
+    for _, version in versions.items():
+        actual_versions.append(version)
+
+    assert sorted(actual_versions) == expected_versions, f"A major version is missing: {versions}"
```
arjkesh, line 114: nit: Change to 
```
f"Found DLC major versions {actual_versions} but expected {expected_versions} for {framework} {job_type} {processor}. Full version info: {versions}"
```

test/dlc_tests/sanity/test_dlc_labels.py
saimidu: We have semantic versioning on all the older images in this repository as well.

test/dlc_tests/sanity/test_dlc_labels.py
```diff
@@ -42,4 +44,71 @@ def test_dlc_major_version_label(image, region):
...
+        reference_fw = "tensorflow1"
+    elif test_utils.is_tf2(image):
+        reference_fw = "tensorflow2"
+    else:
+        reference_fw = framework
+    if processor != "eia" and fw_version < references[reference_fw]:
+        pytest.skip(
+            f"Not enforcing new versioning scheme on old image {image}. "
+            f"Started enforcing version scheme on the following: {references}"
+        )
```
saimidu, line 77: Use [packaging.package.Version](https://packaging.pypa.io/en/latest/version/#packaging.version.Version) to compare version numbers using `<`, `>`, `=`, and `!=`.

This code is not future proof, and can fail when you compare `1.10` to `1.9`, for example.

For example:

https://github.com/aws/deep-learning-containers/blob/master/test/dlc_tests/sanity/test_safety_check.py#L118-L124

test/dlc_tests/sanity/test_dlc_labels.py
```diff
@@ -42,4 +44,71 @@ def test_dlc_major_version_label(image, region):
...
+
+    dockerfiles = []
+    fw_version_major_minor = re.match(r"(\d+.\d+)", fw_version).group(1)
+    for root, dirnames, filenames in os.walk(root_dir):
+        for filename in filenames:
+            if filename == f"Dockerfile.{processor}":
+                dockerfile_path = os.path.join(root_dir, root, filename)
+                if "example" not in dockerfile_path and f"{os.sep}{fw_version_major_minor}" in dockerfile_path:
+                    dockerfiles.append(dockerfile_path)
```
saimidu, line 86: Could you add a short comment before this nested-loop to explain what the purpose of this loop is? It's not hard to understand. It would just take less time to understand if the purpose were to be explained beforehand.

Something like
`# Get paths to dockerfiles for all patch versions matching this framework's major.minor version`

test/dlc_tests/sanity/test_dlc_labels.py
```diff
@@ -42,4 +44,71 @@ def test_dlc_major_version_label(image, region):
...
+
+    versions = {}
+    dlc_label_regex = re.compile(r'LABEL dlc_major_version="(\d+)"')
+    for dockerfile in dockerfiles:
+        with open(dockerfile, "r") as df:
+            for line in df:
+                major_version_match = dlc_label_regex.match(line)
+                if major_version_match:
+                    versions[dockerfile] = int(major_version_match.group(1))
+
```
saimidu, line 96: Can you add an Exception here that will be thrown if `major_version_match` is not found after the loop is complete?

test/dlc_tests/sanity/test_dlc_labels.py
```diff
@@ -42,4 +44,71 @@ def test_dlc_major_version_label(image, region):
...
+    expected_versions = list(range(1, len(dockerfiles) + 1))
+    actual_versions = []
+
+    for _, version in versions.items():
+        actual_versions.append(version)
```
saimidu, line 101: nit:
```
actual_versions = versions.values()
```
is more efficient.

test/dlc_tests/sanity/test_dlc_labels.py
saimidu, line 86: This loop doesn't deal with multiple python versions.

If I were to create a Python 3.6 version in parallel with a Python 3.7 version of a framework version's CPU and GPU dockerfiles, this loop would catch both versions.

It looks like the logic further below will fail, in that case.

test/test_utils/__init__.py
```diff
@@ -554,3 +554,20 @@ def get_repository_and_tag_from_image_uri(image_uri):
...
+
+    Assumes image uri includes -<processor>- in it's tag, where <processor> is one of cpu, gpu or eia.
+
+    :param image_uri: ECR image URI
+    :return: cpu, gpu, or eia
+    """
+    allowed_processors = ("cpu", "gpu", "eia")
+
+    for processor in allowed_processors:
+        match = re.search(rf'-({processor})-', image_uri)
```
saimidu, line 571: Please remove the second `-` in this line, and make it `rf"-({processor})"`, so that this function can be used on TensorFlow Inference image tags such as `tensorflow-inference:2.2.0-cpu`

test/test_utils/__init__.py
```diff
@@ -554,3 +554,20 @@ def get_repository_and_tag_from_image_uri(image_uri):
...
+        if match:
+            return match.group(1)
```
saimidu, line 573: nit: You don't actually need the value of `match`, here. All you need is to check whether `match` is `None` or an `re.Match` object, and you can `return processor`.

test/dlc_tests/sanity/test_dlc_labels.py
```diff
@@ -42,4 +44,71 @@ def test_dlc_major_version_label(image, region):
 
     assert major_version, f"{image} has no LABEL named 'dlc_major_version'. Please insert label."
 
-    LOGGER.info(f"{image} has 'dlc_major_version' = {major_version}")
+    test_utils.LOGGER.info(f"{image} has 'dlc_major_version' = {major_version}")
+
+
+@pytest.mark.integration("dlc_major_version_label")
+@pytest.mark.model("N/A")
+def test_dlc_version_dockerfiles(image):
```
YYStreet, line 52: Could it be better to change the name to "test_dlc_major_version_dockerfiles"?

test/dlc_tests/sanity/test_dlc_labels.py
arjkesh: It should catch both versions, right? Would our version scheme not expect two different DLC major versions for different python versions?

test/dlc_tests/sanity/test_dlc_labels.py
arjkesh: Sure good point, didn't think to do it because of the test above

test/dlc_tests/sanity/test_dlc_labels.py
arjkesh: Thanks for the suggestion, will check this out

test/dlc_tests/sanity/test_dlc_labels.py
arjkesh: will do

test/test_utils/__init__.py
arjkesh: Many of our other tests are not explicitly used for SM tags, but will add this for now.

test/dlc_tests/sanity/test_dlc_labels.py
```diff
@@ -42,4 +44,71 @@ def test_dlc_major_version_label(image, region):
...
+
+    # Test case explicitly for TF2.3 gpu, since v1.0 is banned
+    if (framework, fw_version_major_minor, processor) == ("tensorflow", "2.3", "gpu"):
+        expected_versions = [v + 1 for v in expected_versions]
+        assert 1 not in actual_versions, (
+            f"DLC v1.0 is deprecated in TF2.3 gpu containers, but found major version 1 "
+            f"in one of the Dockerfiles. Please inspect {versions}"
+        )
+
+    assert sorted(actual_versions) == expected_versions, (
```
YYStreet, line 111: From the logic of this test, we are expecting the dlc major versions of the Dockerfiles of 
1. a certain framework(e.g. tf) 
2. with a certain version x.y(e.g. 2.2) 
3. for a certain processor (e.g. gpu)

in a pattern of a consecutive set beginning with 1(e.g. {1, 2, 3}).

For example, there are 2 images for tf2.2 gpu. One is cu101 with major version 1, the other is cu102 with major version 2.

We are assuming every time we are going to increase the major version (for a certain framework version x.y and a certain processor, e.g. gpu), it introduces a new Dockerfile, and in the mean time, we are still retaining all the previous dockerfiles for it. However, there might be some other cases that break assumption.

According to our dlc version design, increment of major versions could also depend on “backward-compatibility”.
For example, we have cuda101 TF2.2 with major version 1 and cuda102 TF2.2 with major version 2 at the beginning. However, there is a breaking/backward-incompatible change of a dependency and we decide to upgrade the image for cuda102 TF2.2, then it should update it to major version 3 without introducing new Dockerfile. Now they would be in versions {1, 3} and breaks the assumed pattern.

test/test_utils/__init__.py
arjkesh: I think it makes it more readable to explicitly set match to a variable. I will raise an exception if no match is returned, rather than sending "None" as I am explicitly expecting to return one of the processors

test/dlc_tests/sanity/test_dlc_labels.py
arjkesh: If we come across this scenario, though, shouldn't we still be displaying all of the major versions in Dockerfiles, so a user can build with a version that may have a breaking change?

test/dlc_tests/sanity/test_dlc_labels.py
arjkesh: sure, can update the name

test/dlc_tests/sanity/test_dlc_labels.py
YYStreet: That is a good idea to display all existing versions information. However, the assertion would fail in this case but we still want to proceed with the given versions.
We can keep this test like this and allow the exception for it once there is such a scenario. A comment or an error prompt to remind release lead could be helpful.

test/test_utils/__init__.py
saimidu: Thanks. This function is usable across a wide range of tests, and having it be generic to test on all our DLC image tags will be better than locking our code to be specifically usable only for CI builds.

test/test_utils/__init__.py
saimidu: Alright, makes sense to me.

test/dlc_tests/sanity/test_dlc_labels.py
```diff
@@ -85,19 +91,35 @@ def test_dlc_version_dockerfiles(image):
                 if "example" not in dockerfile_path and f"{os.sep}{fw_version_major_minor}" in dockerfile_path:
                     dockerfiles.append(dockerfile_path)
 
+    # For the collected dockerfiles above, note the DLC major versions in each Dockerfile if python version matches
+    # the current image under test
     versions = {}
     dlc_label_regex = re.compile(r'LABEL dlc_major_version="(\d+)"')
+    python_version_regex = re.compile(r'ARG PYTHON_VERSION=(\d+.\d+)')
```
saimidu, line 98: nit: It doesn't matter in this particular case, because our definition of python version always has the format `X.Y.Z`, but when you have a regular expression, it is best to use `r"\d+\.\d+"` rather than `r"\d+.\d+"` to specify that the `.` is meant to be matched literally, and not as "any-character".

test/dlc_tests/sanity/test_dlc_labels.py
```diff
@@ -85,19 +91,35 @@ def test_dlc_version_dockerfiles(image):
...
+                raise DLCPythonVersionNotFound(f"Cannot find PYTHON_VERSION arg in {dockerfile}")
+            if python_version == python_major_minor_version:
+                versions[dockerfile] = dlc_version
 
     expected_versions = list(range(1, len(dockerfiles) + 1))
     actual_versions = []
 
-    for _, version in versions.items():
+    for version in versions.values():
         actual_versions.append(version)
```
saimidu: I still don't see the purpose of having this loop, rather than `actual_versions = versions.values()`.

saimidu: Not sure why it says "Outdated", but I can still see this in the code.

test/dlc_tests/sanity/test_dlc_labels.py
```diff
@@ -42,4 +45,100 @@ def test_dlc_major_version_label(image, region):
...
+                    dlc_version = int(major_version_match.group(1))
+                elif python_version_match:
+                    python_version = python_version_match.group(1).replace('.', '')
+
+            # Raise errors if dlc major version label and python version arg are not found in Dockerfile
+            if not dlc_version:
+                raise DLCMajorVersionLabelNotFound(f"Cannot find dlc_major_version label in {dockerfile}")
+            if not python_version:
+                raise DLCPythonVersionNotFound(f"Cannot find PYTHON_VERSION arg in {dockerfile}")
+            if python_version == python_major_minor_version:
```
YYStreet, line 116: What if we add the support for a certain python version after several release? e.g. we add support for py38 for images, and py38 image DLC versions is not beginning with version 1.0. Do we also need to address this case especially once it emerges?

PR #491, opened 2020-08-12
[build] Add yaml linter step for all top level yml files

arjkesh:
*Issue #, if available:*

## Checklist

*Description:*
Currently, we have no yaml linting process. Since changes to some top level yaml files do not result in builds/tests, added a pre-build step to run yaml linting.

Note: I did not add this for $FRAMEWORK/*.yml files because a) they require significant changes and would be better scoped to a different PR and b) they are exercised in the build process

*Tests run:*
- Ran tool locally
- See builds on "add a failure condition for testing purposes" for example failure.

### DIFF
--- a/buildspec.yml
+++ b/buildspec.yml
@@ -12,6 +12,7 @@ phases:
       - $(aws ecr get-login --no-include-email --region $AWS_DEFAULT_REGION)
       - pip install -r src/requirements.txt
       - bash src/setup.sh $FRAMEWORK
+      - yamllint -d .yamllint *.yml
   build:
     commands:
       - echo Build started on `date`
--- a/.yamllint
+++ b/.yamllint
@@ -1,7 +1,6 @@
 extends: default
 
 rules:
-  # 80 chars should be enough, but don't fail if a line is longer
   line-length:
     max: 120
     level: warning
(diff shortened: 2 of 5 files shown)


buildspec.yml
```diff
@@ -12,6 +12,7 @@ phases:
       - $(aws ecr get-login --no-include-email --region $AWS_DEFAULT_REGION)
       - pip install -r src/requirements.txt
       - bash src/setup.sh $FRAMEWORK
+      - yamllint -d .yamllint *.yml
```
YYStreet, line 15: Is the testing content/result the same across building workers? It might be not easy to check the content within the one with building task. We might utilize the eia builder to check the linter results.

arjkesh: yeah so the top level content should be the same. this can easily be extended to also test $FRAMEWORK/*.yml but wanted to do that in a different PR(s) since that will potentially have significant ramifications on the build process

YYStreet: If we want to have a better experience to retrieve the result of linter, it might be better to put it into post_build stage because there are too many building information during build stage. For PR context, we can check the eia builder for help. However for mainline/nightly context might be not easy to check.

PR #495, opened 2020-08-13
[test] remove aws account id

tejaschumbalkar:
## Checklist

*Description:*

- Remove AWS account ID


TusharKanekiDey (dismissed):

test/test_utils/ecs.py
```diff
@@ -25,7 +25,7 @@
     {"containerPort": 8080, "hostPort": 80, "protocol": "tcp"},
 ]
 
-ECS_INSTANCE_ROLE_ARN = "arn:aws:iam::669063966089:instance-profile/ecsInstanceRole"
+ECS_INSTANCE_ROLE_ARN = "arn:aws:iam::{}:instance-profile/ecsInstanceRole"
```
arjkesh, line 28: nit: is this only used in one place? we probably don't need to store it as a constant at all, and can instead construct it in the only function that uses it via the following:
```
f"arn:aws:iam::{account_id}:instance-profile/{ecs_role_name}"
```
Do you see any value to it being a constant?

tejaschumbalkar: agreed. will make the changes.

PR #506, opened 2020-08-17
Updated logic for image size check

SergTogul:
*Issue #, if available: n/a*

## Checklist
- [x] I've prepended PR tag with frameworks/job this applies to : [build]
- [n/a] (If applicable) I've documented below the DLC image/dockerfile this relates to
- [n/a] (If applicable) I've documented below the tests I've run on the DLC image
- [n/a] (If applicable) I've reviewed the licenses of updated and new binaries and their dependencies to make sure all licenses are on the Apache Software Foundation Third Party License Policy Category A or Category B license list.  See [https://www.apache.org/legal/resolved.html](https://www.apache.org/legal/resolved.html).
- [n/a] (If applicable) I've scanned the updated and new binaries to make sure they do not have vulnerabilities associated with them.

#### EIA Checklist
* When creating a PR:
- [ ] I've modified ```src/config/build_config.py``` in my PR branch by setting ```ENABLE_EI_MODE = True```

*Description:*
If an image size is more then 120% of the 'image_size_baseline' in buildspec.yml then the build is failed with respective message but the image is stored in the repository. Modules and packages installed in the image are listed in the log.

*Tests run:*
Tested locally with different frameworks

*DLC image/dockerfile:*

*Additional context:*
When built image is too big and its size is breaching the specified image limit, it's hard to investigate the cause since the new image isn't saved in the repository.


src/image.py
```diff
@@ -115,11 +129,10 @@ def build(self):
             if self.summary["image_size"] > self.info["image_size_baseline"] * 1.20:
                 response.append("Image size baseline exceeded")
                 response.append(f"{self.summary['image_size']} > 1.2 * {self.info['image_size_baseline']}")
-                self.log = response
-                self.build_status = constants.FAIL
-                self.summary["status"] = constants.STATUS_MESSAGE[self.build_status]
-                self.summary["end_time"] = datetime.now()
-                return self.build_status
+                response += self.collect_installed_packages_information()
```
YYStreet, line 132: Where can we have a check of these installed_packages_information?

PR #550, opened 2020-08-28
[test] Pass canary tests when no release tags exist

saimidu:
*Issue #, if available:*

## Checklist

## Reviewer Checklist

*Description:*
The canary test assumes that at least one release tag exists for each framework. The test should be able to run when there are no tags, as well.

*Tests run:*
Tested locally.

*DLC image/dockerfile:*
N/A

*Additional context:*
N/A

### DIFF
--- a/src/send_status.py
+++ b/src/send_status.py
@@ -67,7 +67,11 @@ def post_status(state):
     context = f"{trigger_job}_{project_name}"
     description = set_build_description(state, project_name, trigger_job)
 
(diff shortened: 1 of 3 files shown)


arjkesh:
While I like this change, by making it, we aren't testing the same experience used in get_dlc_images and thus the way we are consuming images into our pytest framework. If you make the change this way, you could update the function to call get_dlc_images and set canary context to True, or do the ".join" in the parse canary images function.

PR #595, opened 2020-09-14
[test] Test coverage reports should be put into a specific bucket

SergTogul:
*Description:*
Test coverage report files should be put into a bucket specific for execution account

*Tests run:*

*DLC image/dockerfile:*

*Additional context:*


test/dlc_tests/sanity/test_generate_coverage_doc.py
```diff
@@ -9,8 +9,8 @@
 from test.test_utils import LOGGER, is_mainline_context
 from test.test_utils.test_reporting import get_test_coverage_file_path
 
-
-TEST_COVERAGE_REPORT_BUCKET = f"dlc-test-coverage-reports"
+account_id = boto3.client('sts').get_caller_identity().get('Account')
```
arjkesh, line 12: Why not use $ACCOUNT_ID which we have included in our documentation instead of the sts call?

saimidu: Why use $ACCOUNT_ID when we know that the sts call will always work, and will always be reliable?

saimidu, line 12: Please leave two spaces between the import statements and the account_id assignment.

saimidu (approved):

test/dlc_tests/sanity/test_generate_coverage_doc.py
arjkesh: Main reason is just grabbing from an env variable is faster, and we can always default to sts

saimidu: I don't think it would make much of a difference, but sure, I wouldn't mind seeing it implemented as something like
```
account_id = os.getenv("ACCOUNT_ID", boto3.client("sts").get_caller_identity().get("Account"))
```

saimidu: Also nit: Please use double-quotes to comply with formatting rules we normally enforce using `black`.

Specifically, `black -l 120`.

saimidu (approved):

arjkesh (approved):

PR #602, opened 2020-09-15
[test][canary] Update canary logic to include py2 deprecation version

arjkesh:
*Issue #, if available:*

## Checklist
- [x] I've prepended PR tag with frameworks/job this applies to : [mxnet, tensorflow, pytorch] | [ei/neuron] | [build] | [test] | 

*Description:*
MXNet 1.7.0 is py3 only, which caused canary tests to fail. We need a way of automatically determining when py2 is no longer supported for a DLC. Adding this logic to the function, as well as another function that will be used to handle python minor versions.

*Tests run:*
Canary pull sanity test


test/test_utils/__init__.py
```diff
@@ -416,6 +419,22 @@ def get_dlc_images():
     raise RuntimeError(f"Cannot find any images for in {test_images}")
 
 
+def get_canary_default_tag_py3_version(framework, version):
```
arjkesh, line 422: non-formatting changes are from here to line 546

tejaschumbalkar (approved):
LGTM

PR #616, opened 2020-09-21
 [mxnet, tensorflow, pytorch] [test] Check Exit Code For Each SMDebug Framework Test

NihalHarish:
*Issue #, if available:*

## PR Checklist

## Pytest Marker Checklist

## Reviewer Checklist

*Description:*
- The PR collects the exit code for each pytest call
- if the exit code is 0, it prints out the name of the test that passed successfully
- exit code 5; is a special pytest case when it does not collect any "_tests_" in the test file. 
This happens when we run pytest for non test files like utils
- any exit code other than 0 or 5 is deemed a failure and the test exits.
- See: https://github.com/pytest-dev/pytest/blob/050c2df737f7c6c3b1de7a6f375de107edcb6873/src/_pytest/config/__init__.py#L98 for details about pytest exit_code=5

*Tests run:*
- manually tested both failure scenarios.

*DLC image/dockerfile:*
- n/a

*Additional context:*

### DIFF
--- a/test/dlc_tests/container_tests/bin/testSmdebug
+++ b/test/dlc_tests/container_tests/bin/testSmdebug
@@ -60,6 +60,13 @@ else
             echo "Ignoring $test_file"
         else
             pytest $test_file --ignore=mock/tests/mxnet/test_training_end.py
+            if [ "$?" -eq "0" ]
+            then
+                echo "SMDebug test succeeded: $test_file"
+            else
+                echo "SMDebug test failed: $test_file"
+                exit 1
+            fi
         fi
     done
 
--- a/test/dlc_tests/container_tests/bin/testSmdebug
+++ b/test/dlc_tests/container_tests/bin/testSmdebug
@@ -66,6 +66,8 @@ else
                 echo "SMDebug test succeeded: $test_file"
             elif [ "$exit_code" -eq "5" ]
             then
+                # pytest enum ExitCode.NO_TESTS_COLLECTED = 5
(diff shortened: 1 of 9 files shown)


test/dlc_tests/container_tests/bin/testSmdebug
```diff
@@ -60,6 +60,17 @@ else
             echo "Ignoring $test_file"
         else
             pytest $test_file --ignore=mock/tests/mxnet/test_training_end.py
+            exit_code=$?
+            if [ "$exit_code" -eq "0" ]
+            then
+                echo "SMDebug test succeeded: $test_file"
+            elif [ "$exit_code" -eq "5" ]
+            then
+                echo "SMDebug no tests collected: $test_file"
```
saimidu, line 69: When does this happen, and why?

Also, could you add a comment with an explanation in case you have special cases like this?

Finally, in line 79, the script checks `"$?"`, but the exit code variable applies to only the last-executed bash command. This needs to be more concrete.

NihalHarish: The check below is redundant since we check the result of every test.
I have removed them

saimidu: Could you add a check for the results of the test in line 53?

saimidu (approved):

test/dlc_tests/container_tests/bin/testSmdebug
NihalHarish: Done

saimidu (approved):

PR #621, opened 2020-09-22
[test] [sagemaker] Stop AMI updates before sm local tests

saimidu:
*Issue #, if available:*

## PR Checklist

## Reviewer Checklist

*Description:*
Stop all automated AMI updates after the EC2 instance for SageMaker Local tests is spun up, before the test begins, so that the following error doesn't occur:
```
E: Could not get lock /var/lib/dpkg/lock-frontend - open (11: Resource temporarily unavailable)
E: Unable to acquire the dpkg frontend lock (/var/lib/dpkg/lock-frontend), is another process using it?
```

*Tests run:*
Tested this locally

*DLC image/dockerfile:*
N/A

*Additional context:*
N/A

### DIFF
--- a/test/test_utils/sagemaker.py
+++ b/test/test_utils/sagemaker.py
@@ -184,7 +184,7 @@ def install_sm_local_dependencies(framework, job_type, image, ec2_conn):
     is_py3 = " python3 -m"
     # To avoid the dpkg lock with apt-daily service if exists
     sleep(300)
-    # using virtualenv to avoid package conflicts with the current packages            
+    # using virtualenv to avoid package conflicts with the current packages
     ec2_conn.run(f"sudo apt-get install virtualenv -y ")
     if framework == "tensorflow" and job_type == "inference":
         # TF inference test fail if run as soon as instance boots, even after health check pass. rootcause:
@@ -227,7 +227,22 @@ def execute_local_tests(image, ec2_client):
         ec2_conn.run(f"$(aws ecr get-login --no-include-email --region {region})")
         ec2_conn.run(f"docker pull {image}")
         ec2_conn.run(f"tar -xzf {sm_tests_tar_name}")
-        ec2_conn.run("sleep 300")  # Sleep for 5 minutes while EC2 instance file-system changes are completed
+
+        stop_updates_command = """
+                    APT_DAILYS="apt-daily.service apt-daily-upgrade.service unattended-upgrades.service";
+                    (sudo systemctl stop ${APT_DAILYS} &&
+                        sudo systemctl kill --kill-who=all ${APT_DAILYS} &&
+                        (while ! (systemctl list-units --all ${APT_DAILYS} | egrep -q '(dead|failed)'); do
+                            echo 'stopping unattended upgrades';
+                            sleep 1;
+                        done);
+                    ) || echo "skip stopping of unattended upgrades";
+                    sudo rm -rf /var/lib/dpkg/lock*;
+                    sudo dpkg --configure -a;
+                    sudo apt-get update
+                    """
+        ec2_conn.run(stop_updates_command)
+
         with ec2_conn.cd(path):
             install_sm_local_dependencies(framework, job_type, image, ec2_conn)
             # Workaround for mxnet cpu training images as test distributed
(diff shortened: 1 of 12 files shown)


test/test_utils/sagemaker.py
```diff
@@ -227,6 +227,22 @@ def execute_local_tests(image, ec2_client):
         ec2_conn.run(f"$(aws ecr get-login --no-include-email --region {region})")
         ec2_conn.run(f"docker pull {image}")
         ec2_conn.run(f"tar -xzf {sm_tests_tar_name}")
+
+        stop_updates_command = """
```
arjkesh, line 231: can you provide more context to what this is in a comment? is there a simpler way to accomplish this instead of running a long shell command?

test/test_utils/sagemaker.py
```diff
@@ -227,6 +227,22 @@ def execute_local_tests(image, ec2_client):
...
+                    APT_DAILYS="apt-daily.service apt-daily-upgrade.service unattended-upgrades.service";
```
arjkesh, line 232: why set bash env variables instead of string interpolation??

test/test_utils/sagemaker.py
saimidu: Sure. Will add a comment in the code.

The idea here is that there are daily apt updates run in the background when an instance is spun-up. This background process conflicts with the foreground `virtualenv` process, because of which we experience the `dpkg frontend lock` conflicts.

This shell-command/mini-script kills the background process, and runs the apt-get update in the foreground, making the update process synchronous, and preventing the test from moving forward until the host instance is "ready".

test/test_utils/sagemaker.py
saimidu: Do you mean `APT_DAILYS`? The string `"apt-daily.service ..."` is repeated thrice in the command, so it makes sense to modularize it.

test/test_utils/sagemaker.py
```diff
@@ -183,8 +183,8 @@ def install_sm_local_dependencies(framework, job_type, image, ec2_conn):
     # Install custom packages which need to be latest version"
     is_py3 = " python3 -m"
     # To avoid the dpkg lock with apt-daily service if exists
-    sleep(300)
-    # using virtualenv to avoid package conflicts with the current packages            
+    ec2_conn.run(f"sleep 300")
```
arjkesh, line 186: nit: why is this an f string?

saimidu: I wanted to make the sleep time a variable controlled from the python script, but wanted to try experimenting with 300 seconds hardcoded first. Will change it as soon as this attempt succeeds (which it looks like it will).

test/test_utils/sagemaker.py
```diff
@@ -198,6 +194,39 @@ def install_sm_local_dependencies(framework, job_type, image, ec2_conn):
...
+
+    :param ec2_conn: Fabric SSH connection
+    :return:
+    """
+    apt_daily_services_list = ["apt-daily.service", "apt-daily-upgrade.service", "unattended-upgrades.service"]
+    apt_daily_services = " ".join(apt_daily_services_list)
+    ec2_conn.run(f"sudo systemctl stop {apt_daily_services}")
+    ec2_conn.run(f"sudo systemctl kill --kill-who=all {apt_daily_services}")
+    num_stopped_services = 0
+    for _ in range(60):
```
arjkesh, line 211: can you explain how you came up with the 60x execution of this loop?

saimidu: It is expected to normally take about 1-2 seconds. The 60 second loop exists to force the execution to wait for a longer amount of time than it would normally take to kill the services.

Also, the loop is meant to be run for 60 seconds, with a sleep of 1 second on each iteration. I missed this while adding the changes. Will fix this.

test/test_utils/sagemaker.py
```diff
@@ -3,7 +3,6 @@
 import subprocess
 import random
 import re
-
```
arjkesh, line 6: nit: i believe this newline should not be removed

saimidu: I specifically removed it here to make it follow the import convention on Python.

Reference:
https://www.python.org/dev/peps/pep-0008/
https://stackoverflow.com/questions/20762662/whats-the-correct-way-to-sort-python-import-x-and-from-x-import-y-statement

arjkesh (approved):

PR #623, opened 2020-09-23
[test] Update smdebug test failure logic and keypair naming logic

arjkesh:
*Issue #, if available:*

## PR Checklist

*Description:*
- While timeouts were being properly addressed and logs, other errors may go undetected due to tee vs stdout redirect behavior and fabric exit codes. This only affects smdebug tests, updating logic.

- Fabric sometimes stalls or times out, even if the actual docker command finishes successfully. Redid the logic here to check docker logs to see if the script has finished to completion.

- Also update ec2 keypair logic in conftest - bug was causing framework to not get added, which results in conflicts btwn PT 1.6 and MX 1.6, for instance

*Tests run:*
- Tested on m4.16xlarge and p2.8xlarge
- Tested on small instance types (t2.2x and g3.4x)
- MXNet and TF1 are currently failing due to actual SMDebug issues. This PR should be merged in to properly expose those issues.


test/dlc_tests/ec2/test_smdebug.py
```diff
@@ -61,9 +61,8 @@ def run_smdebug_test(
     try:
         test_output = ec2_connection.run(
             f"{docker_executable} exec --user root {container_name} "
-            f"/bin/bash -c '{test_script} {framework}' | tee {logfile}",
+            f"/bin/bash -c '{test_script} {framework}' >> {logfile}",
             hide=True,
-            warn=True,
```
NihalHarish, line 66: Why do we remove this flag?

test/dlc_tests/conftest.py
```diff
@@ -335,14 +332,18 @@ def generate_unique_values_for_fixtures(metafunc_obj, images_to_parametrize, val
 
                     image_tag = image.split(":")[-1].replace(".", "-")
 
-                    framework = image.split(":")[0].split("/")[1].split("-")[0]
```
arjkesh, line 338: note: this was not actually grabbing framework and job_type - simplified implementation so we won't run into the mxnet1.6/pytorch1.6 bugs

test/dlc_tests/ec2/test_smdebug.py
arjkesh: This is set warn=True so that we can raise the actual fabric error along with additional context

```
        raise SMDebugTestFailure(
            f"SMDebug test failed on {image_uri} on {ec2_instance_type}. Full output:\n{debug_stdout}"
        ) from e
```

Now we will raise more than just assertion error if the test fails - we will have the fabric error referenced in the "**from e**" and the full debug information from docker logs in debug_stdout

test/dlc_tests/ec2/test_smdebug.py
```diff
@@ -10,73 +10,70 @@
 
 
 SMDEBUG_EC2_GPU_INSTANCE_TYPE = get_ec2_instance_type(default="p3.8xlarge", processor="gpu")
-SMDEBUG_EC2_CPU_INSTANCE_TYPE = get_ec2_instance_type(default="c5.9xlarge", processor="cpu")
+SMDEBUG_EC2_CPU_INSTANCE_TYPE = get_ec2_instance_type(default="c4.8xlarge", processor="cpu")
```
arjkesh, line 13: Changed instance type to reflect a commonly tested cpu instance type in PR context. Left p3.8x as is so that this is exercised on p3 instance types in PR context. Can change back if folks have strong opinions about this.

test/dlc_tests/ec2/test_smdebug.py
```diff
@@ -10,73 +10,70 @@
...
-        LOGGER.error(f"Caught exception while trying to run test via fabric. Output: {debug_output.stdout}")
-        raise
-
-    # LOGGER.info(test_output.stdout)  # Uncomment this line for a complete log dump
-
-    assert test_output.ok, f"SMDebug tests failed. Output:\n{test_output.stdout}"
+    except Exception as e:
+        debug_output = ec2_connection.run(f"docker logs {container_name}")
+        debug_stdout = debug_output.stdout
+        if "All SMDebug tests succeeded!" in debug_stdout:
```
arjkesh, line 69: In some instance types, fabric connection times out even though the command continues to run on the container. We can check the docker logs on the container itself to see if the test completed successfully - in this case, we will warn, but not fail the test, since SMDebug test passed.

A potential future solution to this is to break SMDebug tests into several smaller tests, and add more regular logging to the tests. In cases of timeout, fabric times out waiting for the next "." to be populated from pytest. However, since we do not control the tests in the sagemaker-debugger repo, this workaround allows us to account for false negatives.

I will follow up in a future PR to test breaking SMDebug tests into smaller parts, but that is out of scope for this fix.

test/dlc_tests/ec2/test_smdebug.py
```diff
@@ -10,73 +10,70 @@
...
-            f"{docker_executable} exec --user root {container_name} "
-            f"/bin/bash -c '{test_script} {framework}' | tee {logfile}",
+        ec2_connection.run(
+            f"{docker_executable} run --name {container_name} -v "
+            f"{container_test_local_dir}:{os.path.join(os.sep, 'test')}{shm_setting}{image_uri} "
+            f"./{test_script} {framework}",
             hide=True,
-            warn=True,
-            timeout=3000,
+            timeout=2400,
```
arjkesh, line 64: 30 mins seems like a more reasonable timeout - tested on small instance types (g3.4x and t2.2x) - tests still pass in this time frame.

test/dlc_tests/conftest.py
```diff
@@ -335,14 +332,18 @@ def generate_unique_values_for_fixtures(metafunc_obj, images_to_parametrize, val
...
-                    job_type = image.split(":")[0].split("/")[1].split("-")[1]
+                    framework_tag = (
+                        "tf" if "tensorflow" in image else "pt" if "pytorch" in image else "mx"
+                        if "mxnet" in image else ""
+                    )
```
saimidu, line 338: Please use the `get_framework_from_image_uri()` function to get the image framework.

test/dlc_tests/ec2/test_smdebug.py
```diff
@@ -10,73 +10,70 @@
...
-        hide=True,
-    )
-
     try:
-        test_output = ec2_connection.run(
-            f"{docker_executable} exec --user root {container_name} "
-            f"/bin/bash -c '{test_script} {framework}' | tee {logfile}",
+        ec2_connection.run(
+            f"{docker_executable} run --name {container_name} -v "
+            f"{container_test_local_dir}:{os.path.join(os.sep, 'test')}{shm_setting}{image_uri} "
```
saimidu, line 61: nit: Why put the spaces in the `shm_setting` string, rather than placing the spaces in this main string here?

I think it is more readable if the arguments are properly spaced in this string.

test/dlc_tests/ec2/test_smdebug.py
saimidu: What does
> fabric times out waiting for the next "." to be populated from pytest

mean?

Since we don't run pytest commands inside test scripts on any other tests apart from this, is this a unique error that occurs only in smdebug tests?

saimidu (approved):

test/dlc_tests/ec2/test_smdebug.py
arjkesh: These are the scenarios I see:

scenario 1:
```
shm_setting = "--shm-size=1g" if ec2_instance_type in large_shm_instance_types else ""
...
f"{container_test_local_dir}:{os.path.join(os.sep, 'test')} {shm_setting} {image_uri} "
```
In this scenario, you have 2 empty spaces between two arguments - this may be harmless, but better to avoid

scenario 2: 
```
shm_setting = "--shm-size=1g " if ec2_instance_type in large_shm_instance_types else ""
...
f"{container_test_local_dir}:{os.path.join(os.sep, 'test')} {shm_setting}{image_uri} "
```
This could also go the other way (i.e. space can be before). This seems like the least readable option to me.

scenario 3: 
```
shm_setting = " --shm-size=1g " if ec2_instance_type in large_shm_instance_types else " "
...
f"{container_test_local_dir}:{os.path.join(os.sep, 'test')}{shm_setting}{image_uri} "
```
How it is currently implemented - avoids scenario adding 2 spaces, more consistent than scenario 2

test/dlc_tests/ec2/test_smdebug.py
arjkesh: When pytest runs, it will populate a "." to the terminal when a test is completed. We could provide more logging here and see if that helps, but that will **intensely** flood logs when there is a failure, making it especially difficult to identify failures.

Yes, this is unique to smdebug tests because we are not only running pytest on an ec2 instance via a shell script thru fabric, but for longer running tests (like the keras tests), there is a long wait time between shell updates.

test/dlc_tests/ec2/test_smdebug.py
saimidu: I think scenario 1 is acceptable. But I can live with scenario 3.

arjkesh: Scenario 1 can introduce functional problems. Let's go with scenario 3.

saimidu (approved):

PR #792, opened 2021-01-15
add an env var to collect the number of cores

WillianPaiva:
related to the Case ID 7825958121

as discussed the PyTorch images was using only one worker, and on instances like a **g4dn-xlarge**, for example, the GPU was heavily underused.
 the workaround to fix this was adding an env var to the **entrypoint** file.

to avoid having to build one docker for each instance here I check first the number of cores and add the variable **SAGEMAKER_MODEL_SERVER_WORKERS** to the number of cores available on the machine.

### DIFF
--- a/pytorch/inference/docker/build_artifacts/mms-entrypoint.py
+++ b/pytorch/inference/docker/build_artifacts/mms-entrypoint.py
@@ -15,14 +15,17 @@ from __future__ import absolute_import
 import shlex
 import subprocess
 import sys
+import os
+import multiprocessing
 
+os.environ["SAGEMAKER_MODEL_SERVER_WORKERS"] = str(multiprocessing.cpu_count())
 
-if sys.argv[1] == 'serve':
+if sys.argv[1] == "serve":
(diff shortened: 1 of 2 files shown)


juckuc (approved):

pytorch/inference/docker/build_artifacts/mms-entrypoint.py
```diff
@@ -15,14 +15,17 @@
 import shlex
 import subprocess
 import sys
+import os
+import multiprocessing
 
+os.environ["SAGEMAKER_MODEL_SERVER_WORKERS"] = str(multiprocessing.cpu_count())
```
gradientsky, line 21: This should happen only if `SAGEMAKER_MODEL_SERVER_WORKERS` is not set; otherwise there is a risk of loading multiple copies of the same model and going beyond RAM/GPU RAM limits. Also probably a good idea to have upper cap (i.e. if there is a large instance with lots of cores, it's not necessarily optimal to have the same number of workers).

PR #973, opened 2021-03-18
[build] Update deep_learning_containers.py

saimidu:
*Issue #, if available:*

## PR Checklist

## Pytest Marker Checklist
- [x] (If applicable) I have added the marker `@pytest.mark.model("<model-type>")` to the new tests which I have added, to specify the Deep Learning model that is used in the test (use `"N/A"` if the test doesn't use a model)
- [x] (If applicable) I have added the marker `@pytest.mark.integration("<feature-being-tested>")` to the new tests which I have added, to specify the feature that will be tested
- [x] (If applicable) I have added the marker `@pytest.mark.processor(<"cpu"/"gpu"/"eia"/"neuron">)` to the new tests which I have added, if a test is specifically applicable to only one processor type

#### EIA/NEURON Checklist
* When creating a PR:

## Reviewer Checklist

*Description:*
Update the deep_learning_containers.py script to improve functionality and add new tests

*Tests run:*

*DLC image/dockerfile:*

*Additional context:*


src/deep_learning_container.py
```diff
@@ -135,14 +190,39 @@ def query_bucket():
...
+    tag = f"{framework}_{container_type}_{framework_version}_python{py_version}_{device}{cuda_version}_{os_version}"
+    tag_struct = {"Key": "aws-dlc-autogenerated-tag-do-not-delete", "Value": tag}
+
+    request_status = None
+    if instance_id and region:
+        try:
+            session = botocore.session.get_session()
+            ec2_client = session.create_client("ec2", region_name=region)
+            response = ec2_client.create_tags(Resources=[instance_id], Tags=[tag_struct])
+            request_status = response.get("ResponseMetadata").get("HTTPStatusCode")
```
junpuf, line 215: why not use `response.raise_for_status()`?

src/deep_learning_container.py
```diff
@@ -83,27 +89,76 @@ def _retrieve_instance_region():
...
+
+def _retrieve_os():
+    version = ""
+    name = ""
+    with open("/etc/os-release", "r") as f:
+        for line in f.readlines():
+            if re.match(r"^ID=\w+$", line):
+                name = re.search(r"^ID=(\w+)$", line).group(1)
+            if re.match(r'^VERSION_ID="\d+\.\d+"$', line):
+                version = re.search(r'^VERSION_ID="(\d+\.\d+)"$', line).group(1)
```
SergTogul, line 123: What about Amazon Linux? Shall we handle that case as well?

tensorflow/training/docker/2.4/py3/cu110/Dockerfile.gpu
```diff
@@ -262,7 +262,7 @@ RUN SMDATAPARALLEL_TF=1 pip install --no-cache-dir ${SMDATAPARALLEL_BINARY}
 
 ENV LD_LIBRARY_PATH="/usr/local/lib/python3.7/site-packages/smdistributed/dataparallel/lib:$LD_LIBRARY_PATH"
 
-ADD https://raw.githubusercontent.com/aws/deep-learning-containers/master/src/deep_learning_container.py /usr/local/bin/deep_learning_container.py
+COPY deep_learning_container.py /usr/local/bin/deep_learning_container.py
 
 RUN chmod +x /usr/local/bin/deep_learning_container.py
```
junpuf, line 267: when is `/usr/local/bin/deep_learning_container.py` executed?

SergTogul (approved):
Approved with a comment

test/dlc_tests/sanity/test_telemetry.py
```diff
@@ -0,0 +1,139 @@
...
+
+    container_name = f"{repo_name}-telemetry_tag_instance_success-ec2"
+
+    docker_cmd = "nvidia-docker" if processor == "gpu" else "docker"
+
+    test_utils.login_to_ecr_registry(ec2_connection, account_id, image_region)
+    ec2_connection.run(f"{docker_cmd} pull -q {image_uri}")
+
+    preexisting_ec2_instance_tags = ec2_utils.get_ec2_instance_tags(ec2_instance_id, ec2_client=ec2_client)
+    if expected_tag_key in preexisting_ec2_instance_tags:
```
junpuf, line 122: why explicitly delete existing tag?

tensorflow/training/docker/2.4/py3/cu110/Dockerfile.gpu
saimidu: It is executed by `import <framework>` Python statements within a container.

junpuf: ok

src/deep_learning_container.py
saimidu: In this case, `response` is a dict as per the convention in https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.Client.create_tags.

I return the request HTTP status code just to make the function signature equivalent for `tag_instance` and `query_bucket`.

src/deep_learning_container.py
saimidu: Let's cross that bridge when we come to it.

Right now, none of the rest of our code is capable of handling Amazon Linux, either.

src/deep_learning_container.py
junpuf: ahh I mistake this is wrriten in `requests`.  Would raising Exception cause anything happen?

test/dlc_tests/sanity/test_telemetry.py
saimidu: The validation for the test is that a tag has been applied correctly. Therefore, by applying a tag to the instance (unintentionally or not), it would be possible to make the test succeed even when the `deep_learning_container.py` script has actually failed to apply a tag on the instance.

junpuf: ok. I guess example could be like:

1. Someone manually created instance tag.
2. Some framework being import has been done before this?

junpuf (dismissed):

saimidu:
We cannot get the HuggingFace DLCs to pass the newly added tests until we rebuild those images on top of DLCs which consume the latest dockerfiles with the changes to `deep_learning_containers.py`.

src/deep_learning_container.py
saimidu: It would still be invisible to a user that runs this script indirectly through `import <framework>`, but not to those who directly execute this script, because the `import <framework>` statements have been written to suppress the output from the process that executes this script.

However, I'd like to force this script to run as quietly as possible regardless, so that it fails silently even if the parent process doesn't suppress its output.

test/dlc_tests/sanity/test_telemetry.py
saimidu: Exactly.

Also, we create separate instances for each test, right now. But it is possible, in theory, that we would refactor our tests to reduce this parallelism (maybe to cut down on the number of instances created?). This bit of code will ensure that a previously-run test on the same instance doesn't corrupt the preconditions for this test.

junpuf: cool

SergTogul (approved):

saimidu (commented):
Treating Approval before the config-revert commit as approval for this final revision.

PR #1018, opened 2021-04-08
[test][sagemaker] Execute SM remote tests in parallel

tejaschumbalkar:
*Issue #, if available:*

## PR Checklist
- [ ] I've prepended PR tag with frameworks/job this applies to : [mxnet, tensorflow, pytorch] | [ei/neuron] | [build] | [test] | [benchmark] | [ec2, ecs, eks, sagemaker]

## Pytest Marker Checklist

#### EIA/NEURON Checklist
* When creating a PR:
#### Benchmark Checklist
* When creating a PR:
- [ ] I've modified `src/config/test_config.py` in my PR branch by setting `ENABLE_BENCHMARK_DEV_MODE = True`

## Reviewer Checklist

*Description:*

- Add pytest `rerun` config for SM remote tests. This will make sure to rerun the failed test. `rerun` option will rerun the failed test that does not have `pytest.mark.flaky` fixture. The failed test with `pytest.mark.flaky` fixture will respect the configuration on the `flaky` fixture and override the `rerun` option for that test.
- Configuring `-n=auto` in pytest for SM remote test to enable running the test in parallel. This will significantly drop the test time from 2-3 hr to ~40-50 min.
- Occasionally, the test codebuild job timesout and it is difficult to find out the test that led to the timeout. Adding a pytest feature that will log the name of the test currently being executed by pytest.


test/sagemaker_tests/tensorflow/tensorflow2_training/integration/conftest.py
```diff
@@ -146,3 +146,16 @@ def skip_py2_containers(request, tag):
     if request.node.get_closest_marker('skip_py2_containers'):
         if 'py2' in tag:
             pytest.skip('Skipping python2 container with tag {}'.format(tag))
+
+@pytest.fixture(autouse=True)
+def log_current_test():
+    """
+    Log the name of the test currently being executed by pytest
+    """
```
saimidu, line 154: Nice.

test/requirements.txt
```diff
@@ -2,10 +2,10 @@ docker==4.2.0
 fabric==2.5.0
 junit-xml==1.9
 packaging==20.1
-pytest==5.3.5
-pytest-rerunfailures==9.0
+pytest==6.2.3
+pytest-rerunfailures==9.1.1
 pytest-timeout
-pytest-xdist==1.31.0
+pytest-xdist==2.2.1
```
saimidu, line 8: Do we need to upgrade these versions to make use of some functionality that isn't available right now, or is it more of a general currency update?

(Either way, thanks.)

test/test_utils/sagemaker.py
```diff
@@ -88,7 +88,8 @@ def generate_sagemaker_pytest_cmd(image, sagemaker_test_type):
     :param sagemaker_test_type: local or remote test type
     :return: <tuple> pytest command to be run, path where it should be executed, image tag
     """
-    reruns = 4
+    reruns = 1
```
saimidu, line 91: Does this override the number of reruns we set for individual SM tests? If so, is `reruns = 1` sufficient?

src/config/build_config.py
```diff
@@ -7,7 +7,7 @@
 # Frameworks for which you want to disable both builds and tests
 DISABLE_FRAMEWORK_TESTS = []
 # Disable new builds or build without datetime tag
-DISABLE_DATETIME_TAG = False
+DISABLE_DATETIME_TAG = True
```
tejaschumbalkar, line 10: To be reverted once PR is ready for merge

src/config/test_config.py
```diff
@@ -2,9 +2,9 @@
 # Do remember to revert it back to False before merging any PR (including Benchmark dedicated PR)
 ENABLE_BENCHMARK_DEV_MODE = False
 # Disable the test codebuild jobs to be run
-DISABLE_SANITY_TESTS = False
+DISABLE_SANITY_TESTS = True
```
tejaschumbalkar, line 5: To be reverted once PR is ready for merge

test/requirements.txt
tejaschumbalkar: There is an inconsistency in functionality of pytest when `rerun` and `-n=auto` options are used in a single command for older version of `pytest` and related plugins. 
Upgrading the same to the latest version resolves the issue.

test/test_utils/sagemaker.py
tejaschumbalkar: updated the description on this.

tejaschumbalkar:
Current test fails due to limits on SM jobs. Working on increasing the same.

saimidu:
Looks like the `dlc-pr-tensorflow-2_dlc-pr-sagemaker-test` job hit this error:
```
botocore.exceptions.ClientError: An error occurred (ThrottlingException) when calling the CreateTrainingJob operation (reached max retries: 4): Rate exceeded
```

test/dlc_tests/conftest.py
```diff
@@ -350,6 +350,18 @@ def pt14_and_above_only():
...
 
 
+@pytest.fixture(autouse=True)
+def log_current_test():
+    """
+    Log the name of the test currently being executed by pytest
+    """
+    try:
+        test_name = f"{os.environ.get('PYTEST_CURRENT_TEST').split('::')[-1].split(' ')[0]}"
+        LOGGER.info(f"============================= Executing test :: {test_name} :: =============================")
```
saimidu, line 360: Does each test set the common environment variable when it is begun even if the test is being executed in parallel with other tests in other threads?

Multiple processes would have separate env variables, but would the multiple parallel pytest threads be able to accurately print out their test name from a shared environment variable?

saimidu: It looks like each test gets its own process, and not its own thread -- https://docs.pytest.org/en/stable/example/simple.html#dynamically-adding-command-line-options

So I guess you can ignore this comment.

tejaschumbalkar:
> Looks like the `dlc-pr-tensorflow-2_dlc-pr-sagemaker-test` job hit this error:
> 
> ```
> botocore.exceptions.ClientError: An error occurred (ThrottlingException) when calling the CreateTrainingJob operation (reached max retries: 4): Rate exceeded
> ```

I have reached out to the SM python SDK team on possible solution for retry mechanism

tejaschumbalkar:
> > Looks like the `dlc-pr-tensorflow-2_dlc-pr-sagemaker-test` job hit this error:
> > ```
> > botocore.exceptions.ClientError: An error occurred (ThrottlingException) when calling the CreateTrainingJob operation (reached max retries: 4): Rate exceeded
> > ```
> 
> Th is due to throttling of `CreateTrainingJob` API on SM side. I have reached out to the SM python SDK team on possible solution for retry mechanism on the `sagemaker` library.

BTW, it looks like retry would be the only option as per https://github.com/aws/sagemaker-python-sdk/issues/645. This is already done as a part of `rerun` Do you think we should increase the `rerun` to 2 or 3 in that case?

tejaschumbalkar:
> Current test fails due to limits on SM jobs. Working on increasing the same.

SM limits are in place. The current test failures are due to `CapaticyErrors`.

saimidu (commented):
I'm okay with this implementation, but it seems to result in CapacityErrors. I don't think it would be wise to merge this as-is if parallelism introduces CapacityErrors.

How about we set `-n=TUNABLE_PARAMETER` instead of `-n=auto`?

We can configure `TUNABLE_PARAMETER` with some heuristics and guesswork (how about `-n=4`)?

tejaschumbalkar:
> I'm okay with this implementation, but it seems to result in CapacityErrors. I don't think it would be wise to merge this as-is if parallelism introduces CapacityErrors.
> 
> How about we set `-n=TUNABLE_PARAMETER` instead of `-n=auto`?
> 
> We can configure `TUNABLE_PARAMETER` with some heuristics and guesswork (how about `-n=4`)?

The `CapaticyErrors` are not due to parallel execution in this case as we were already seeing the issue in the current PR's. 
However, I agree that it will lead to `CapaticyErrors` couple of times.

tejaschumbalkar:
Addressed `CapacityErrors` in #1054

PR #1127, opened 2021-05-28
Clean up setup instructions

juliensimon:
- Add step to create ECR repository
- Use env variables more consistently


juliensimon:
Pretty disappointing that this simple PR couldn't be merged after almost a year. Closing it.

PR #1160, opened 2021-06-17
[test] Use Allowlists with ECR Scan Test

saimidu:
*Issue #, if available:*

## PR Checklist

## Reviewer Checklist

*Description:*
Include the capability to use Allowlists with the ECR Scan test results on DLCs

*Tests run:*

*DLC image/dockerfile:*

*Additional context:*


test/test_utils/security.py
```diff
@@ -0,0 +1,80 @@
...
+            and device_type in self.exceptions.get("device_type", [])
+        ):
+            return True
+
+
+class ScanAllowList:
+    def __init__(self, file_path):
+        with open(file_path, "r") as f:
+            allowlist_json = json.load(f)
+        self.vulnerability_dict = dict()
```
arjkesh, line 63: nit: can we name `self.vulnerability_dict` something like `self.allowed_vulnerabilities` and `vulnerability_list` something like `allowlist_vulnerabilities` so it's clear that we are not blocking on these vulnerabilities?

test/evaluate_ecr_scan.py
```diff
@@ -0,0 +1,51 @@
...
+            break
+        sleep(1)
+    if scan_status != "COMPLETE":
+        raise TimeoutError(f"ECR Scan is still in {scan_status} state. Exiting.")
+    severity_counts = ecr_utils.get_ecr_image_scan_severity_count(ecr_client, image)
+    scan_results = ecr_utils.get_ecr_image_scan_results(ecr_client, image, minimum_vulnerability=minimum_sev_threshold)
+    LOGGER.info(json.dumps(scan_results, indent=4))
+
+    # for vulnerability in scan_results:
+
```
arjkesh, line 46: nit: I believe two newlines suffice here

test/evaluate_ecr_scan.py
```diff
@@ -0,0 +1,51 @@
...
+        if scan_status == "COMPLETE":
+            break
+        sleep(1)
+    if scan_status != "COMPLETE":
+        raise TimeoutError(f"ECR Scan is still in {scan_status} state. Exiting.")
+    severity_counts = ecr_utils.get_ecr_image_scan_severity_count(ecr_client, image)
+    scan_results = ecr_utils.get_ecr_image_scan_results(ecr_client, image, minimum_vulnerability=minimum_sev_threshold)
+    LOGGER.info(json.dumps(scan_results, indent=4))
+
+    # for vulnerability in scan_results:
```
arjkesh, line 45: nit: is this leftover code?

test/evaluate_ecr_scan.py
```diff
@@ -0,0 +1,51 @@
+import argparse
```
arjkesh, line 1: Is there a reason to write this as a standalone script as opposed to a pytest function? Does it run in parallel across images? If it needs to be a standalone script, let's call out a justification for the decision in the docstring

test/evaluate_ecr_scan.py
```diff
@@ -0,0 +1,51 @@
...
+
+from datetime import datetime
+from time import sleep, time
+
+import boto3
+import pytest
+
+from packaging.version import Version
+
+from invoke import run
```
arjkesh, line 12: nit: these two imports can be merged

test/evaluate_ecr_scan.py
```diff
@@ -0,0 +1,51 @@
+import argparse
+import json
+
+from datetime import datetime
+from time import sleep, time
+
+import boto3
+import pytest
```
arjkesh, line 8: nit: do we need pytest imported here?

test/evaluate_ecr_scan.py
saimidu: This was experimental code that I uploaded to give more context, during intermediate commits.

test/test_utils/security.py
```diff
@@ -0,0 +1,80 @@
...
+            framework in self.exceptions.get("framework", [])
+            and framework_version in self.exceptions.get("framework_version", [])
+            and image_type in self.exceptions.get("image_type", [])
+            and python_version in self.exceptions.get("python_version", [])
+            and device_type in self.exceptions.get("device_type", [])
+        ):
+            return True
+
+
+class ScanAllowList:
```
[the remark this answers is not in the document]
saimidu: Added

src/config/build_config.py
```diff
@@ -7,7 +7,7 @@
 # Frameworks for which you want to disable both builds and tests
 DISABLE_FRAMEWORK_TESTS = []
 # Disable new builds or build without datetime tag
-DISABLE_DATETIME_TAG = False
+DISABLE_DATETIME_TAG = True
```
saimidu, line 10: To be reverted.

src/config/test_config.py
```diff
@@ -5,11 +5,11 @@
 # Disable the test codebuild jobs to be run
 
 # It is recommended to set DISABLE_EFA_TESTS to True to disable EFA tests if there is no change to EFA installer version or Frameworks.
-DISABLE_EFA_TESTS = False
+DISABLE_EFA_TESTS = True
```
saimidu, line 8: To be reverted.

tensorflow/buildspec.yml
```diff
@@ -1,8 +1,8 @@
 account_id: &ACCOUNT_ID <set-$ACCOUNT_ID-in-environment>
 region: &REGION <set-$REGION-in-environment>
 framework: &FRAMEWORK tensorflow
-version: &VERSION 2.5.1
-short_version: &SHORT_VERSION 2.5
+version: &VERSION 2.4.1
+short_version: &SHORT_VERSION 2.4
```
saimidu, line 5: To be reverted.

tensorflow/training/docker/2.4/py3/Dockerfile.cpu.os_scan_allowlist.json
```diff
@@ -0,0 +1,328 @@
+{
```
saimidu, line 1: To be reverted.

tensorflow/training/docker/2.4/py3/Dockerfile.cpu.os_scan_allowlist.json
```diff
@@ -0,0 +1,328 @@
...
+    "avahi": [
+        {
+            "name": "CVE-2021-3468",
+            "description": "A flaw was found in avahi in versions 0.6 up to 0.8. The event used to signal the termination of the client connection on the avahi Unix socket is not correctly handled in the client_work function, allowing a local attacker to trigger an infinite loop. The highest threat from this vulnerability is to the availability of the avahi service, which becomes unresponsive after this flaw is triggered.",
+            "uri": "http://people.ubuntu.com/~ubuntu-security/cve/CVE-2021-3468",
+            "severity": "MEDIUM",
+            "attributes": [
```
jeet4320, line 8: It can format like below which is more readable

```
"attributes": {
         package_version: "0.7-3.1ubuntu1.2",
         package_name: avahi
}
```

tensorflow/training/docker/2.4/py3/Dockerfile.cpu.os_scan_allowlist.json
```diff
@@ -0,0 +1,328 @@
...
+            "description": "Arm Armv8-A core implementations utilizing speculative execution past unconditional changes in control flow may allow unauthorized disclosure of information to an attacker with local user access via a side-channel analysis, aka \"straight-line speculation.\"",
+            "uri": "http://people.ubuntu.com/~ubuntu-security/cve/CVE-2020-13844",
+            "severity": "MEDIUM",
+            "attributes": [
+                {
+                    "key": "package_version",
+                    "value": "7.5.0-3ubuntu1~18.04"
+                },
+                {
+                    "key": "package_name",
```
jeet4320, line 40: package_name is already defined on line 28

tensorflow/training/docker/2.4/py3/Dockerfile.cpu.os_scan_allowlist.json
```diff
@@ -0,0 +1,328 @@
...
+                    "value": "AV:A/AC:M/Au:N/C:N/I:N/A:P"
+                },
+                {
+                    "key": "CVSS2_SCORE",
+                    "value": "2.9"
+                }
+            ]
+        }
+    ],
+    "wget": [
```
jeet4320, line 302: we should add reason field on why we allowed it

tensorflow/training/docker/2.4/py3/Dockerfile.cpu.os_scan_allowlist.json
jeet4320: Isn't this lot of info in code, could the below format be okay?
```
[
                {
                    "package": "gcc-7",
                    "versions": ["7.5.0-3ubuntu1~18.04"],
                    "CVEs": ["CVE-2020-13844"],
                    "reason": "New version not yet available.",
                },
                {
                    "package": "tensorflow",
                    "versions": ["2.3.2"],
                    "CVEs": ["CVE-2020-13844"],
                    "reason": "New version not yet available.",
                }
]
```

test/dlc_tests/sanity/test_security_check.py
```diff
@@ -77,11 +79,15 @@ def test_ecr_scan(image, ecr_client, sts_client, region):
...
+    image_scan_allowlist_path = get_dockerfile_path_for_image(image) + ".os_scan_allowlist.json"
+    image_scan_allowlist = ScanAllowList(image_scan_allowlist_path)
+    reduced_vulnerabilities_list = []
+    for vulnerability in scan_results:
+        if vulnerability not in image_scan_allowlist:
+            reduced_vulnerabilities_list.append(vulnerability)
+
+    LOGGER.info(json.dumps(reduced_vulnerabilities_list, indent=4))
+
+    assert not reduced_vulnerabilities_list, json.dumps(reduced_vulnerabilities_list, indent=4)
```
jeet4320, line 93: assert same var reduced_vulnerabilities_list again actual and expected?

tensorflow/training/docker/2.4/py3/Dockerfile.cpu.os_scan_allowlist.json
saimidu: Unfortunately, cleaning up the format there is going to take manual effort for whoever updates this file manually, because it is a deviation from the format of attributes provided by the ECR Scan Tool.

tensorflow/training/docker/2.4/py3/Dockerfile.cpu.os_scan_allowlist.json
saimidu: I want to make the task of whoever updates this file as simple as Copying and Pasting a Scan Result for an image. By changing the format to be more concise, we improve the formatting, but at the cost of more time taken to update these vulnerabilities, as well as losing the ability to directly (through code or even visually) compare the `allowlist` with the `scan_results` from the scanner.

tensorflow/training/docker/2.4/py3/Dockerfile.cpu.os_scan_allowlist.json
saimidu: Sure.

test/dlc_tests/sanity/test_security_check.py
saimidu: The expected list should be 0.

tensorflow/training/docker/2.4/py3/Dockerfile.cpu.os_scan_allowlist.json
jeet4320: Ohok, I thought its format defined by you. Lets keep as it is then.

test/evaluate_ecr_scan.py
saimidu: This was experimental code that I uploaded to give more context, during intermediate commits. I have deleted it now.

test/dlc_tests/sanity/test_security_check.py
```diff
@@ -10,9 +11,14 @@
...
-from test.test_utils import get_account_id_from_image_uri, get_region_from_image_uri, login_to_ecr_registry
+from test.test_utils import (
+    LOGGER, get_account_id_from_image_uri, get_dockerfile_path_for_image
+)
 from test.test_utils import ecr as ecr_utils
-from test.test_utils.ecr import CVESeverity
+from test.test_utils.security import CVESeverity, ScanVulnerabilityList
+
+
+MINIMUM_SEV_THRESHOLD = "MEDIUM"
```
saimidu, line 21: This will be set back to "HIGH" until we actually have allowlists for dockerfiles.

tensorflow/training/docker/2.4/py3/Dockerfile.cpu.os_scan_allowlist.json
```diff
@@ -0,0 +1,341 @@
...
+                {
+                    "key": "CVSS2_VECTOR",
+                    "value": "AV:L/AC:L/Au:N/C:N/I:N/A:P"
+                },
+                {
+                    "key": "CVSS2_SCORE",
+                    "value": "2.1"
+                }
+            ],
+            "reason": "Fix is Deferred"
```
shantanutrip, line 26: In the actual allow lists it would be a dictionary of the CVE status. Just to list it down so that we do not forget it.

test/dlc_tests/sanity/test_security_check.py
```diff
@@ -35,6 +41,21 @@ def test_security(image):
...
 
 
+def run_scan(ecr_client, image):
+    scan_status = None
+    start_time = time()
+    ecr_utils.start_ecr_image_scan(ecr_client, image)
+    while (time() - start_time) <= 600:
+        scan_status, scan_status_description = ecr_utils.get_ecr_image_scan_status(ecr_client, image)
+        if scan_status == "FAILED" or scan_status not in [None, "IN_PROGRESS", "COMPLETE"]:
+            raise RuntimeError(scan_status_description)
```
arjkesh, line 51: nit: can we provide more descriptive error

tensorflow/training/docker/2.4/py3/Dockerfile.cpu.os_scan_allowlist.json
saimidu: "reason" is completely unused in the code. So it could be any format.

test/test_utils/__init__.py
```diff
@@ -938,6 +995,20 @@ def get_processor_from_image_uri(image_uri):
     raise RuntimeError("Cannot find processor")
 
 
+def get_python_version_from_image_uri(image_uri):
```
arjkesh, line 998: is this used anywhere in your code? seems like other functions like this follow a pattern of raising exception if no version is found

test/test_utils/__init__.py
```diff
@@ -975,3 +1046,16 @@ def run_cmd_on_container(container_name, context, cmd, executable="bash", warn=F
...
+def ordered(obj):
+    if isinstance(obj, dict):
+        return sorted((k, ordered(v)) for k, v in obj.items())
+    if isinstance(obj, list):
+        return sorted(ordered(x) for x in obj)
+    else:
+        return obj
+
+
+def are_json_objects_equivalent(obj1, obj2):
```
arjkesh, line 1060: is this being used anywhere?

test/test_utils/security.py
```diff
@@ -0,0 +1,172 @@
...
+class CVESeverity(IntEnum):
+    UNDEFINED = 0
+    INFORMATIONAL = 1
+    LOW = 2
+    MEDIUM = 3
+    HIGH = 4
+    CRITICAL = 5
+
+
+class ScanVulnerabilityList:
```
arjkesh, line 18: is there a place to use your equivalence function from test_utils here?

test/dlc_tests/sanity/test_security_check.py
```diff
@@ -64,24 +85,61 @@ def test_ecr_scan(image, ecr_client, sts_client, region):
...
+        f"The following vulnerabilities have not been fixed on {image}:\n"
+        f"{json.dumps(remaining_vulnerabilities.vulnerability_list, indent=4)}"
+    )
+
+
+@pytest.mark.model("N/A")
+@pytest.mark.canary("Check for packages that can be upgraded on production images.")
+@pytest.mark.integration("check OS dependencies")
+def test_is_ecr_scan_allowlist_outdated(image, ecr_client, sts_client, region):
+    """
```
arjkesh, line 112: great test to have, thanks for adding

test/test_utils/__init__.py
saimidu: Good call. It is better to return an Exception.

test/test_utils/__init__.py
saimidu: It's not being used anywhere right now. I added it as a helper function for the `ScanVulnerabilityList` class, but I ended up realizing that I can't do an exact match of the JSON vulnerability lists.

I forgot that I added this, and can remove it now. But this seems like a pretty useful utility function to have.

test/test_utils/security.py
saimidu: I initially added it because I was planning on comparing the Allowlist directly with the ECR Scan Vulnerabilities List.

But the subtraction function is helpful in cases where there are vulnerabilities that are now fixed, and the allowlist is just old. Using `__sub__` instead of `__cmp__` helps us differentiate between when there are unknown vulnerabilities vs when there are outdated vulnerabilities on allowlist.

But the `__cmp__` is useful for further development that @shantanutrip might need, which is why I've left it in.

tensorflow/training/docker/2.4/py3/Dockerfile.cpu.os_scan_allowlist.json
shantanutrip: True, it is not being used in the code right now. Also, it can be any format for the time being.

However, I was just thinking from the perspective of the cron job. In future, when the cron job will extract the allow list x + 1, it will compare it with the current allow list x. When allow lists x and x + 1 will be compared for the difference, the "reason" key for them will turn out to be different. In version x + 1, reason would be {"upstream_status": "deferred", "ubuntu_status": "deferred"} and in version x it will be "deferred". But yes, that would not be that big of an issue.

saimidu: As long as the package version on the allowlist here is not either missing from, or less than the version in, the newly generated ECR Scan result on a CRON job, the reason doesn't really matter.

Ideally, if there is a new version available that is also vulnerable, we still want to upgrade to that new version, and should therefore treat it as a necessary upgrade.

test/test_utils/__init__.py
arjkesh: IMO this doesn't have a clear use case at the moment. Let's remove for now so as not to add test_utils that don't have clear usage examples int he codebase

arjkesh:
Looks like some test failures on a few diff types of images

test/dlc_tests/sanity/test_security_check.py
```diff
@@ -35,6 +37,21 @@ def test_security(image):
         run(f"docker rm -f {container_name}", hide=True)
 
 
+def run_scan(ecr_client, image):
+    scan_status = None
+    start_time = time()
+    ecr_utils.start_ecr_image_scan(ecr_client, image)
+    while (time() - start_time) <= 600:
```
jeet4320, line 44: is 600 timeout enough?

test/dlc_tests/sanity/test_security_check.py
```diff
@@ -35,6 +37,21 @@ def test_security(image):
...
+        scan_status, scan_status_description = ecr_utils.get_ecr_image_scan_status(ecr_client, image)
+        if scan_status == "FAILED" or scan_status not in [None, "IN_PROGRESS", "COMPLETE"]:
+            raise ECRScanFailureException(f"ECR Scan failed for {image} with description: {scan_status_description}")
+        if scan_status == "COMPLETE":
+            break
+        sleep(1)
```
jeet4320, line 50: it would be lot of API calls if we wait for 1 sec. Before making first call to check get_ecr_image_scan_status, we can wait for 3-5 mins based on how long ECR scan takes on average, then reduce wait time to check for every 1 min.

Or we can check for every 1 min always.

test/dlc_tests/sanity/test_security_check.py
```diff
@@ -64,24 +81,61 @@ def test_ecr_scan(image, ecr_client, sts_client, region):
...
+    if os.path.exists(image_scan_allowlist_path):
+        image_scan_allowlist.construct_allowlist_from_file(image_scan_allowlist_path)
+
+    ecr_image_vulnerability_list = ScanVulnerabilityList(minimum_severity=CVESeverity[MINIMUM_SEV_THRESHOLD])
+    ecr_image_vulnerability_list.construct_allowlist_from_ecr_scan_result(scan_results)
+
+    remaining_vulnerabilities = ecr_image_vulnerability_list - image_scan_allowlist
+
+    assert not remaining_vulnerabilities, (
+        f"The following vulnerabilities have not been fixed on {image}:\n"
```
jeet4320, line 99: can we change it to `The following vulnerabilities have been found on ..`

test/test_utils/security.py
```diff
@@ -0,0 +1,189 @@
...
+
+        :param vulnerability: dict JSON object consisting of information about the vulnerability in the format
+                              presented by the ECR Scan Tool
+        :return: bool True if the vulnerability is allowed on the allow-list.
+        """
+        package_name = get_ecr_vulnerability_package_name(vulnerability)
+        if package_name not in self.vulnerability_list:
+            return False
+        for allowed_vulnerability in self.vulnerability_list[package_name]:
+            if are_vulnerabilities_equivalent(vulnerability, allowed_vulnerability):
```
jeet4320, line 79: is this is checking if vulnerability is allowed and can be skipped? If yes are_vulnerabilities_equivalent seems misleading, can be more like are_vulnerabilities_allowed?

test/dlc_tests/sanity/test_security_check.py
```diff
@@ -64,24 +81,61 @@ def test_ecr_scan(image, ecr_client, sts_client, region):
...
+    image_scan_allowlist_path = get_ecr_scan_allowlist_path(image)
+    if os.path.exists(image_scan_allowlist_path):
+        image_scan_allowlist.construct_allowlist_from_file(image_scan_allowlist_path)
+
+    ecr_image_vulnerability_list = ScanVulnerabilityList(minimum_severity=CVESeverity[MINIMUM_SEV_THRESHOLD])
+    ecr_image_vulnerability_list.construct_allowlist_from_ecr_scan_result(scan_results)
+
+    remaining_vulnerabilities = ecr_image_vulnerability_list - image_scan_allowlist
+
+    assert not remaining_vulnerabilities, (
```
shantanutrip, line 98: Is the motive of the ECR Scan tests to just check if new vulnerabilities have been found on a given image? Or, do the tests also check if security patches for the existing vulnerabilities have been released?

test/dlc_tests/sanity/test_security_check.py
```diff
@@ -64,24 +81,61 @@ def test_ecr_scan(image, ecr_client, sts_client, region):
...
+    )
+
+
+@pytest.mark.model("N/A")
+@pytest.mark.integration("check OS dependencies")
+@pytest.mark.skipif(is_dlc_cicd_context(), reason="Temporarily allow slack in allowlist w.r.t. actual vulnerabilities")
+def test_is_ecr_scan_allowlist_outdated(image, ecr_client, sts_client, region):
+    """
+    Run ECR Scan Tool on an image being tested, and test if the vulnerabilities in the allowlist for the image
+    are still valid, or if any vulnerabilities must be removed from the list.
```
shantanutrip, line 110: Let us say we run ecr_scan on an image `i` and we get vulnerabilities  [a,b,c]. Now, let us assume that in a week's time, security patch was released for a . Now, if we run the ecr_scan again on the image, will it show [a,b,c] as vulnerabilities or just [b,c]?

I am asking this because the image `i` is still using the old version of packages and has not consumed any patches yet. Hence vulnerability `a` would still be present in `i'` and when we run the `ecr_scan` on `i`, it will consider the older version of `a` as vulnerable and show `[a,b,c]` as vulnerable packages.

test/dlc_tests/sanity/test_security_check.py
saimidu: It is more than enough, yes. The ECR Scan only takes about 5 minutes, and though it is mostly reliable, I have placed a timeout to prevent test/job timeouts as a general principle.

test/dlc_tests/sanity/test_security_check.py
saimidu: I haven't seen any throttling on ECR API calls caused by these requests sent every second, which is why I've left it as is.

This implementation is sub-optimal, and can use Boto3 Waiters, but that is something we had planned to implement at a later date. I would like to limit the changes on this PR, for now, to not blow its scope up.

test/dlc_tests/sanity/test_security_check.py
saimidu: Changed it to `The following vulnerabilities need to be fixed on`

test/test_utils/security.py
saimidu: This just invokes a helper function that checks for "equivalence" between the vulnerability on the allow list and a vulnerability that is given as input for comparison.

The `are_vulnerabilities_equivalent` function only checks a subset of keys within the two dict objects being compared, which is why the function's name is `are_vulnerabilities_equivalent`.

`are_vulnerabilities_allowed` is a more abstract representation of what this class does, and is something that the pytest test function would check by running `if vulnerability in ScanVulnerabilityList_object:`

test/dlc_tests/sanity/test_security_check.py
saimidu: > Now, if we run the ecr_scan again on the image, will it show [a,b,c] as vulnerabilities or just [b,c]?

The new image `i'` should show `[b, c]` if the newly built image consumed that security patch after its release, yes.

Image `i`, which is the older version, will still have vulnerability `a`. It will therefore be in sync with the allowlist, and will pass the `is_ecr_scan_allowlist_outdated` test. However, `i'`, which is the newer version, would not have vulnerability `a`, and would fail the `is_ecr_scan_allowlist_outdated` test.

Both `i` and `i'` would pass the `test_ecr_scan` test.

jeet4320 (dismissed):

test/dlc_tests/sanity/test_security_check.py
shantanutrip: Nice. Thanks 👍

shantanutrip: If i passes the is_ecr_scan_allowlist_outdated test even when the allow_list is outdated, is the test strong enough?

shantanutrip: We want the `is_ecr_scan_allowlist_outdated` test to fail to figure out if the list is outdated, correct?

saimidu: Correct. Yes, `is_ecr_scan_allowlist_outdated` is supposed to show the difference between `i'` and the vulnerability list.

shantanutrip: Okay, so just to confirm, to run `is_ecr_scan_allowlist_outdated` test do we create i' and run the test on that?

jeet4320 (approved):

PR #1186, opened 2021-06-25
[build] Add safety report to docker image

tejaschumbalkar:
*Issue #, if available:*

## PR Checklist

## Pytest Marker Checklist

#### EIA/NEURON Checklist
* When creating a PR:
#### Benchmark Checklist
* When creating a PR:
- [ ] I've modified `src/config/test_config.py` in my PR branch by setting `ENABLE_BENCHMARK_DEV_MODE = True`

## Reviewer Checklist

*Description:*
- This functionality will add the safety report as a part of the docker image. 
- Safety check will run on the image during build time and the safety report will be embedded into the image itself.
- For any kind of image, it first builds the original image, then builds its common stage image and then pushes the common stage image to the repo. For eg., it first builds Standard images, then common stage images for Standard images and then pushes these common stage images to the repo. Thereafter it start processing the Example images and does the same for those as well.
- During the sanity test, the tests will check if the safety report has been embedded or not and check if all the packages have passed the safety check.
- In case any of the packages have failed the safety check, the sanity tests would fail.


src/config/build_config.py
```diff
@@ -5,9 +5,9 @@
 # Do remember to revert it back to False before merging any PR (including NEURON dedicated PR)
 ENABLE_NEURON_MODE = False
 # Frameworks for which you want to disable both builds and tests
-DISABLE_FRAMEWORK_TESTS = []
+DISABLE_FRAMEWORK_TESTS = ["mxnet","pytorch"]
```
tejaschumbalkar, line 8: Temp change: To be reverted before PR merge

src/config/test_config.py
```diff
@@ -5,11 +5,11 @@
 # Disable the test codebuild jobs to be run
 
 # It is recommended to set DISABLE_EFA_TESTS to True to disable EFA tests if there is no change to EFA installer version or Frameworks.
-DISABLE_EFA_TESTS = False
+DISABLE_EFA_TESTS = True
```
tejaschumbalkar, line 8: Temp change: To be reverted before PR merge

src/safety/safety_check_v2.py
```diff
@@ -0,0 +1,110 @@
...
+        vulns_list.append(v)
+
+    print(json.dumps(vulns_list))
+
+
+if __name__ == "__main__":
+    try:
+        sys.exit(main())
+    except KeyboardInterrupt:
+        pass
```
tejaschumbalkar, line 110: Will add an empty next line

test/dlc_tests/eks/mxnet/inference/test_eks_mxnet_inference.py
```diff
@@ -46,7 +46,8 @@ def test_eks_mxnet_neuron_inference(mxnet_inference, neuron_only):
         if eks_utils.is_service_running(selector_name):
             eks_utils.eks_forward_port_between_host_and_container(selector_name, port_to_forward, "8080")
 
-        assert test_utils.request_mxnet_inference(port=port_to_forward, model="mxnet-resnet50")
+
```
tejaschumbalkar, line 49: Additional new line will be reverted

test/dlc_tests/sanity/test_safety_check.py
```diff
@@ -148,17 +148,11 @@ def _get_latest_package_version(package):
 @pytest.mark.model("N/A")
 @pytest.mark.canary("Run safety tests regularly on production images")
 @pytest.mark.skipif(not is_dlc_cicd_context(), reason="Skipping test because it is not running in dlc cicd infra")
```
tejaschumbalkar, line 150: Temp change: To be reverted before PR merge

src/safety/safety_check_v2.py
```diff
@@ -0,0 +1,110 @@
...
+                "package": pkg.key,
+                "affected": f"No known vulnerability found, PASSED_SAFETY_CHECK on {timestamp}.",
+                "installed": pkg.version,
+                "vulnerabilities": [{"vid": "N/A", "advisory": "N/A"}],
+            }
+        
+    for (k, v) in vulns_dict.items():
+        vulns_list.append(v)
+
+    print(json.dumps(vulns_list))
```
arjkesh, line 103: nit: let's use python logging library to print the json obj

src/safety/test_and_generate_python_safety_report.py
```diff
@@ -0,0 +1,98 @@
...
+failed_count = 0
+for result in scan_results:
+    for report_item in result.report:
+        if "PASSED_SAFETY_CHECK" not in report_item.affected:
+            if (report_item.package not in ignored_packages) or (
+                report_item.package in ignored_packages
+                and report_item.installed != ignored_packages[report_item.package]
+            ):
+                failed_count += 1
+                print(
```
arjkesh, line 81: nit: let's use python logging library here

src/safety/test_and_generate_python_safety_report.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+@dataclass
+class SafetyPythonEnvironmentVulnerabilityReport:
+    report: List[SafetyPackageVulnerabilityReport]
+
+    def __post_init__(self):
+        self.report = [SafetyPackageVulnerabilityReport(**i) for i in self.report]
+
+parser = argparse.ArgumentParser()
```
arjkesh, line 49: for maintainability and future use, can we organize this file into functions?

test/dlc_tests/sanity/test_safety_check.py
```diff
@@ -178,22 +172,27 @@ def test_safety(image):
...
+                    # Version(x) gives an object that can be easily compared with another version, or with a SpecifierSet.
+                    # Comparing two versions as a string has some edge cases which require us to write more code.
+                    # SpecifierSet(x) takes a version constraint, such as "<=4.5.6", ">1.2.3", or ">=1.2,<3.4.5", and
+                    # gives an object that can be easily compared against a Version object.
+                    # https://packaging.pypa.io/en/latest/specifiers/
+                    ignore_str += f" -i {vulnerability_id}"
+            assert (safety_check.run_safety_check_with_ignore_list(docker_exec_cmd, ignore_str) == 0), \
+                f"Safety test failed for {image}"
+        else:
+            LOGGER.info(f"Safety check is complete as a part of docker build and report exist at {SAFETY_FILE}")
```
arjkesh, line 196: nit: typo - exists

test/dlc_tests/sanity/test_safety_check.py
```diff
@@ -178,22 +172,27 @@ def test_safety(image):
...
-                # Version(x) gives an object that can be easily compared with another version, or with a SpecifierSet.
-                # Comparing two versions as a string has some edge cases which require us to write more code.
-                # SpecifierSet(x) takes a version constraint, such as "<=4.5.6", ">1.2.3", or ">=1.2,<3.4.5", and
-                # gives an object that can be easily compared against a Version object.
-                # https://packaging.pypa.io/en/latest/specifiers/
-                ignore_str += f" -i {vulnerability_id}"
-        assert (safety_check.run_safety_check_with_ignore_list(docker_exec_cmd, ignore_str) == 0), \
-            f"Safety test failed for {image}"
+        SAFETY_FILE = "/opt/aws/dlc/info/safety_test_report.json"
+        safety_file_check = run(f"{docker_exec_cmd} test -e {SAFETY_FILE}", warn=True, hide=True)
```
arjkesh, line 176: What happens when we want to run these on a canary basis, etc? How can we be sure we're checking for new vulnerabilities if this file exists?

buildspec.yml
```diff
@@ -1,4 +1,7 @@
 version: 0.2
+env:
+  secrets-manager: 
+    SAFETY_KEY: "/codebuild/safety/key"
```
arjkesh, line 4: can we avoid setting env variables in buildspec and instead do directly on CB job?

src/safety/safety_check_v2.py
junpuf: as long as it output the json text without anything else, then it works.

src/safety/test_and_generate_python_safety_report.py
junpuf: we don't want to see any character other than what's in the print statement. I don't think logging module add any value.

tensorflow/training/docker/2.5/py3/Dockerfile.cpu
```diff
@@ -102,6 +103,7 @@ RUN apt-get update \
     ffmpeg \
     libsm6 \
     libxext6 \
+    jq \
```
jeet4320, line 106: is this needed

tensorflow/training/docker/2.5/py3/cu112/Dockerfile.gpu
```diff
@@ -301,6 +302,18 @@ RUN SMDATAPARALLEL_TF=1 ${PIP} install --no-cache-dir ${SMDATAPARALLEL_BINARY}
 
 ENV LD_LIBRARY_PATH="/usr/local/lib/python3.7/site-packages/smdistributed/dataparallel/lib:$LD_LIBRARY_PATH"
 
+COPY safety /tmp/safety 
+
+RUN mkdir -p /opt/aws/dlc/info 
+
+RUN ${PIP} install --no-cache-dir -U safety \
```
jeet4320, line 309: would this fail build if safety reports CVEs

buildspec.yml
tejaschumbalkar: using multi part docker build instead

src/image_builder.py
```diff
@@ -154,109 +155,177 @@ def image_builder(buildspec):
...
-        # Standard images must be built before example images
-        # Example images will use standard images as base
-        standard_images = [image for image in IMAGES if "example" not in image.name.lower()]
-        example_images = [image for image in IMAGES if "example" in image.name.lower()]
+        FORMATTER.separator()
 
-        for image in standard_images:
-            THREADS[image.name] = executor.submit(image.build)
+        FIRST_STAGES_IMAGES.append(first_stage_image_object)
+        SECOND_STAGES_IMAGES.append(second_stage_image_object)
```
shantanutrip, line 185: This line would fail if the above `if` statement is not true. Would say something like `local variable 'second_stage_image_object' referenced before assignment`

src/image.py
```diff
@@ -84,19 +81,42 @@ def build(self):
...
+        if self.context:
+            with open(self.context.context_path, "rb") as context_file:
+                print("within context")
+                self.docker_build(fileobj=context_file, custom_context=True)
+                self.context.remove()  
+        else:
+            print("out of context")
+            self.docker_build()
+        #check the size after image is built.
+        self.image_size_check()
```
shantanutrip, line 112: We need to return `self.build_status`. If we do not return that, the OutputFormatter fails as `futures[image].result()` remains None

src/image_builder.py
```diff
@@ -154,109 +155,177 @@ def image_builder(buildspec):
...
+    build_images(example_images)
+    push_images(example_images)
+
+    #After the build, display logs/sumary for all the images.
+
+    show_build_logs(ALL_IMAGES)
+    show_build_summary(ALL_IMAGES)
+    is_any_build_failed, is_any_build_failed_size_limit = show_build_errors(ALL_IMAGES)
+
+    #change logic here. upload metrics only for the second stage image
```
shantanutrip, line 227: If I look at `push_image_metrocs` function in ~/deep-learning-containers/src/metrics.py  I observe that build time is generally `image.summary["end_time"] - image.summary["start_time"]`. If we only push the metrics for the second stage image, we will not include the amount of time it takes to build the entire first stage image. We should account for that somehow. 

Also, I feel that we should not build second stage images in case the first stage images fail. We should add that logic handling in the code.

mxnet/buildspec.yml
```diff
@@ -49,56 +49,4 @@ images:
...
-    python_version: &DOCKER_PYTHON_VERSION py3
-    tag_python_version: &TAG_PYTHON_VERSION py37
-    cuda_version: &CUDA_VERSION cu110
-    os_version: &OS_VERSION ubuntu16.04
-    tag: !join [ *VERSION, "-", *DEVICE_TYPE, "-", *TAG_PYTHON_VERSION, "-", *CUDA_VERSION, "-", *OS_VERSION,
-                 "-example" ]
-    docker_file: !join [ docker/, *SHORT_VERSION, /, *DOCKER_PYTHON_VERSION, /example, /Dockerfile., *DEVICE_TYPE ]
-    context:
-      <<: *TRAINING_CONTEXT
+
```
shantanutrip, line 52: Will revert this deletions. Just made this change to run the builds fast enough locally and see if the code is working or not.

mxnet/training/docker/1.8/py3/Dockerfile.cpu
```diff
@@ -3,186 +3,186 @@ FROM ubuntu:16.04
...
- && cp ${HOME_DIR}/oss_compliance/test/testOSSCompliance /usr/local/bin/testOSSCompliance \
- && chmod +x /usr/local/bin/testOSSCompliance \
- && chmod +x ${HOME_DIR}/oss_compliance/generate_oss_compliance.sh \
- && ${HOME_DIR}/oss_compliance/generate_oss_compliance.sh ${HOME_DIR} ${PYTHON} \
- && rm -rf ${HOME_DIR}/oss_compliance*
-
-RUN curl -o /license.txt https://aws-dlc-licenses.s3.amazonaws.com/aws-mx-1.8.0/license.txt
-
-CMD ["/bin/bash"]
+# ARG PYTHON=python3
```
shantanutrip, line 6: Will revert this deletions. Just made this change to run the builds fast enough locally and see if the code is working or not.

src/image_builder.py
```diff
@@ -157,6 +157,7 @@ def image_builder(buildspec):
         }
         
         #Create first stage docker object
+        print(f"[SHAN_TRIP] Image Config Dockerfile {image_config['docker_file']}")
```
shantanutrip, line 160: Need to remove this

src/image_builder.py
```diff
@@ -168,10 +169,13 @@ def image_builder(buildspec):
         )
 
         #Create second stage docker object
-        if "example" not in image_name.lower() and build_context == "MAINLINE":
+        second_stage_image_object = None
+        # if "example" not in image_name.lower() and build_context == "MAINLINE":
+        ###### UNDO THIS CHANGE ########
```
shantanutrip, line 174: Need to revert the if condition.

src/test_build.py
```diff
@@ -0,0 +1,12 @@
+from docker import APIClient
```
shantanutrip, line 1: This will be removed. 
This is an extra file that I have added to work with the Docker APIS.

mxnet/training/docker/1.8/py3/Dockerfile.cpu
```diff
@@ -106,15 +106,15 @@ WORKDIR /
 
 RUN ${PIP} install --no-cache --upgrade \
```
shantanutrip, line 107: This file has no changes. It is like the current master. The change has been made to avoid conflict.

test/dlc_tests/sanity/test_safety_check_v2.py
```diff
@@ -0,0 +1,97 @@
...
+class SafetyPythonEnvironmentVulnerabilityReport:
+    report: List[SafetyPackageVulnerabilityReport]
+
+    def __post_init__(self):
+        self.report = [SafetyPackageVulnerabilityReport(**i) for i in self.report]
+
+
+
+@pytest.mark.model("N/A")
+@pytest.mark.canary("Run safety tests regularly on production images")
```
junpuf, line 53: Will the canary safety test override `/var/safety_report.json` before running this function?

shantanutrip: Thanks for pointing this out. 

The `test_safety_check_v2.py` will run only on PR and Mainline Context. On the Canaries, we will still run the old script that existed. This is done because the old images do not have a baked in safety report in them. Additionally, the purpose of the canaries is to run the safety scans on the images and generate fresh safety reports every day. 

I just have to make a small change to prevent `test_safety_check_v2.py` from running on canary and prevent the old script from running on PR and Mainline Context.

src/safety_report_generator.py
```diff
@@ -0,0 +1,139 @@
...
+        return self.vulns_list
+
+
+
+
+
+    
+
+
+
```
saimidu, line 139: Can this extra space at the end be deleted?

src/safety_report_generator.py
```diff
@@ -0,0 +1,139 @@
...
+        self.container_id = container_id
+        self.vulns_dict = {}
+        self.vulns_list = []
+        self.ignore_dict = ignore_dict
+        self.ignored_vulnerability_count = {}
+        self.ctx = Context()
+        self.docker_exec_cmd = f"docker exec -i {container_id}"
+        self.safety_check_output = None
+    
+    def insert_vulnerabilites_into_report(self, vulns):
```
saimidu, line 43: Please use `vulnerabilities` or whole words instead of shortened words for variable names for clarity.

Also, please add docstrings to all functions in this class.

src/safety_report_generator.py
```diff
@@ -0,0 +1,139 @@
...
+            ]
+            "date":
+        }
+        ...
+    ]
+    """
+
+    def __init__(self, container_id, ignore_dict = {}):
+        self.container_id = container_id
+        self.vulns_dict = {}
```
saimidu, line 35: Please use whole words in variable names for clarity.

src/safety_report_generator.py
```diff
@@ -0,0 +1,139 @@
...
+        self.docker_exec_cmd = f"docker exec -i {container_id}"
+        self.safety_check_output = None
+    
+    def insert_vulnerabilites_into_report(self, vulns):
+        for v in vulns:
+            package = v[0]
+            spec = v[1]
+            installed = v[2]
+            advisory = v[3]
+            vid = v[4]
```
saimidu, line 49: This can be written as `package, spec, installed, advisory, vid = v`.

Also, please use non-single-letter variable names outside of loop indices.

src/image_builder.py
```diff
@@ -154,109 +156,198 @@ def image_builder(buildspec):
...
+
+#### TODO: Remove this entire method when https://github.com/boto/boto3/issues/1592 is resolved ####
+def get_dummy_boto_client():
+    # In absence of this method, the behaviour documented in https://github.com/boto/boto3/issues/1592 is observed.
+    # If this function is not added, boto3 fails because boto3 sessions are not thread safe.
+    # However, once a dummy client is created, it is ensured that the calls are thread safe.
+    import boto3
+    return boto3.client("secretsmanager", region_name=os.getenv('REGION'))
+
+def push_images(images):
```
saimidu, line 342: Please add a docstring to this function.

src/image_builder.py
```diff
@@ -154,109 +156,198 @@ def image_builder(buildspec):
...
+        if make_dummy_boto_client: 
+            get_dummy_boto_client()
+        for image in images:
+            FORMATTER.print(f"image_object.context {image.context}")
+            THREADS[image.name] = executor.submit(image.build)
+    # the FORMATTER.progress(THREADS) function call also waits until all threads have completed
+    FORMATTER.progress(THREADS)
+
+#### TODO: Remove this entire method when https://github.com/boto/boto3/issues/1592 is resolved ####
+def get_dummy_boto_client():
```
saimidu, line 335: Please add a docstring to this function, and place the explanation below into that docstring.

src/image_builder.py
```diff
@@ -154,109 +156,198 @@ def image_builder(buildspec):
...
-                    raise Exception(f"Build passed. {e}")
+    if is_any_build_failed_size_limit:
+        raise Exception("Build failed because of file limit")
 
-        if is_any_build_failed_size_limit:
-            raise Exception("Build failed because of file limit")
+    FORMATTER.separator()
 
-        FORMATTER.separator()
+def build_images(images, make_dummy_boto_client=False):
```
saimidu, line 320: Please add a docstring to this function.

src/image_builder.py
```diff
@@ -154,109 +156,198 @@ def image_builder(buildspec):
...
+    else:
+        if is_any_build_failed_size_limit:
+            FORMATTER.print("Build failed. Image size limit breached.")
         else:
-            if is_any_build_failed_size_limit:
-                FORMATTER.print("Build failed. Image size limit breached.")
+            FORMATTER.print("No errors")
+    return is_any_build_failed, is_any_build_failed_size_limit
+
+def upload_metrics(images, BUILDSPEC, is_any_build_failed, is_any_build_failed_size_limit):
```
saimidu, line 296: Please add a docstring to this function.

src/image_builder.py
```diff
@@ -154,109 +156,198 @@ def image_builder(buildspec):
...
-            FORMATTER.print_lines(image.log)
-            with open(f"logs/{image.name}", "w") as fp:
-                fp.write("/n".join(image.log))
-                image.summary["log"] = f"logs/{image.name}"
+    for image in images:
+        FORMATTER.title(image.name)
+        FORMATTER.table(image.summary.items())
 
-        FORMATTER.title("Summary")
+def show_build_errors(images):
```
saimidu, line 274: Please add a docstring to this function.

src/image_builder.py
```diff
@@ -154,109 +156,198 @@ def image_builder(buildspec):
...
+        FORMATTER.title(image_description)
+        FORMATTER.table(image.info.items())
+        FORMATTER.separator()
+        FORMATTER.print_lines(image.log)
+        with open(f"logs/{image_description}", "w") as fp:
+            fp.write("/n".join(image.log))
+            image.summary["log"] = f"logs/{image_description}"
 
-        FORMATTER.title("Build Logs")
+def show_build_summary(images):
```
saimidu, line 266: Please add a docstring to this function.

src/image_builder.py
```diff
@@ -154,109 +156,198 @@ def image_builder(buildspec):
...
+        dockerfile=os.path.join(os.sep, os.getenv("ROOT_FOLDER_PATH"), "src", "Dockerfile.multipart"),
+        repository=initial_stage_image_object.repository,
+        tag=initial_stage_image_object.tag,
+        to_build=initial_stage_image_object.to_build,
+        stage=constants.CONCLUSION_STAGE,
+    )
+    initial_stage_image_object.to_push = False
+    return conclusion_stage_image_object
+
+def show_build_logs(images):
```
saimidu, line 249: Please add a docstring to this function.

src/image_builder.py
```diff
@@ -154,109 +156,198 @@ def image_builder(buildspec):
...
+    # Set environment variables to be consumed by test jobs
+    test_trigger_job = utils.get_codebuild_project_name()
+    # Tests should only run on images that were pushed to the repository
+    utils.set_test_env(
+        IMAGES_TO_PUSH,
+        BUILD_CONTEXT=os.getenv("BUILD_CONTEXT"),
+        TEST_TRIGGER=test_trigger_job,
+    )
+
+def get_conclusion_stage_image_object(initial_stage_image_object):
```
saimidu, line 234: Please add a docstring to this function.

src/safety_report_generator.py
```diff
@@ -0,0 +1,139 @@
...
+                self.vulns_dict[pkg['key']] = {
+                    "package": pkg['key'],
+                    "scan_status": "SUCCEEDED",
+                    "installed": pkg['version'],
+                    "vulnerabilities": [{"vid": "N/A", "advisory": "N/A", "reason_to_ignore":"N/A", "spec":"N/A"}],
+                    "date":self.timestamp
+                }
+    
+    def process_report(self):
+        for (k, v) in self.vulns_dict.items():
```
saimidu, line 94: What are `k` and  `v` in this case? What is the key for this dict, and what is the value assigned to that key?

src/safety_report_generator.py
```diff
@@ -0,0 +1,139 @@
+from invoke.context import Context
+from datetime import datetime
+
+import json
+import os
+
+TO_BE_DECIDED="To Be Decided"
```
saimidu, line 7: Why is `TO_BE_DECIDED` a variable, while `SUCCEEDED`, `FAILED`, and `IGNORED` are all used as string literals?

src/safety_report_generator.py
```diff
@@ -0,0 +1,139 @@
...
+    
+    def insert_safe_packages_into_report(self, packages):
+        for pkg in packages:
+            if pkg['key'] not in self.vulns_dict:
+                self.vulns_dict[pkg['key']] = {
+                    "package": pkg['key'],
+                    "scan_status": "SUCCEEDED",
+                    "installed": pkg['version'],
+                    "vulnerabilities": [{"vid": "N/A", "advisory": "N/A", "reason_to_ignore":"N/A", "spec":"N/A"}],
+                    "date":self.timestamp
```
saimidu, line 90: Please run `black -l 120` on this file.

Reference: https://github.com/psf/black

src/utils.py
```diff
@@ -486,3 +487,64 @@ def set_test_env(images, images_env="DLC_IMAGES", **kwargs):
...
+    :param image_uri:
+    :return: <list> list of safety check IDs to ignore
+    """
+    framework = ("mxnet" if "mxnet" in image_uri else
+                 "pytorch" if "pytorch" in image_uri else
+                 "tensorflow")
+    job_type = ("training" if "training" in image_uri else
+                "inference-eia" if "eia" in image_uri else
+                "inference-neuron" if "neuron" in image_uri else
+                "inference")
```
saimidu, line 503: Can this be made more future proof so that we don't have to keep adding more `if-else` conditions as more device types are added here?

src/utils.py
```diff
@@ -486,3 +487,64 @@ def set_test_env(images, images_env="DLC_IMAGES", **kwargs):
...
+    python_version = "py2" if "py2" in image_uri else "py3"
+
+    IGNORE_SAFETY_IDS = {}
+    with open(f'{os.getenv("ROOT_FOLDER_PATH")}/data/ignore_ids_safety_scan.json') as f:
```
saimidu, line 507: Please use `os.path.join()` instead of string formatting to generate file paths.

src/utils.py
```diff
@@ -486,3 +487,64 @@ def set_test_env(images, images_env="DLC_IMAGES", **kwargs):
...
+        IGNORE_SAFETY_IDS = json.load(f)
+
+    return IGNORE_SAFETY_IDS.get(framework, {}).get(job_type, {}).get(python_version, {})
+    
+
+
```
saimidu, line 513: Too many spaces.

test/dlc_tests/sanity/test_safety_check_v2.py
```diff
@@ -0,0 +1,99 @@
+import json
```
saimidu, line 1: Please rename this file to `test_safety_check_in_pipelines`, and the other file to `test_safety_check_canary`.

test/dlc_tests/sanity/test_safety_check_v2.py
```diff
@@ -0,0 +1,99 @@
...
+    run(f"docker run -id "
+        f"--name {container_name} "
+        f"--entrypoint='/bin/bash' "
+        f"{image}", hide=True)
+    
+    try:
+        # Check if file exists
+        docker_exec_cmd = f"docker exec -i {container_name}"
+        safety_file_check = run(f"{docker_exec_cmd} test -f {SAFETY_FILE}", warn=True, hide=True)
+        assert safety_file_check.return_code == SAFETY_FILE_EXISTS,  f"Safety file existence test failed for {image}"
```
saimidu, line 70: The `SAFETY_FILE_EXISTS` variable is unnecessary, because `command_result.return_code == 0` is a standard expectation to check if a command was successful, as per `invoke`/`fabric` documentation.

Also, it is more concise to use `command_result.ok` when just checking if the return code is zero or non-zero. `command_result.return_code` can be used when the return code value can be meaningfully compared to get more details about the failure.

test/dlc_tests/sanity/test_safety_check_v2.py
```diff
@@ -0,0 +1,99 @@
...
+
+        file_content = run(f"{docker_exec_cmd} cat {SAFETY_FILE}", warn=True, hide=True)
+        raw_scan_result = json.loads(file_content.stdout)
+        scan_results = []
```
saimidu, line 74: Why make this a list if there is only a single item in the list by design?

test/dlc_tests/sanity/test_safety_check_v2.py
```diff
@@ -0,0 +1,99 @@
...
+        scan_results.append(
+            SafetyPythonEnvironmentVulnerabilityReport(
+                report=raw_scan_result
+            )
+        )
```
saimidu, line 79: This can be in a single line.

Please run `black -l 120` on this file to ensure that it matches formatting with the rest of the code.

Reference: https://github.com/psf/black

src/Dockerfile.multipart
```diff
@@ -0,0 +1,7 @@
+# Use the Deep Learning Container as a base Image
```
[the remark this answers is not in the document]
junpuf: The current design decision is to use 1 single docker file to host stuff that are applicable to all containers. So something like `Dockerfile.common` will work. 

Having multi-staged capability will be definitely better so we can have `Dockerfile.oss_compliance`, `Dockerfile.dependency_check` etc. And a DLC build process can pick and choose which ones to run. But we are not there yet.

src/Dockerfile.multipart
```diff
@@ -0,0 +1,7 @@
...
+ARG INITIAL_STAGE_IMAGE=""
+
+FROM $INITIAL_STAGE_IMAGE
+
+# Add any script or repo as required
+COPY safety_report.json /var/safety_report.json
```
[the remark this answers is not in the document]
junpuf: the location for the file should be `/opt/aws/dlc/info`

src/image.py
```diff
@@ -69,34 +72,67 @@ def collect_installed_packages_information(self):
...
+        self.pre_build_configuration()
+        print(f"self.context {self.context}")
+
+        ## Start building the image
+        if self.context:
+            with open(self.context.context_path, "rb") as context_file:
+                print("within context")
+                self.docker_build(fileobj=context_file, custom_context=True)
+                self.context.remove()  
+        else:
```
tejaschumbalkar, line 115: Is there a case where the `self.docker_build()` is invoked?

src/image.py
```diff
@@ -69,34 +72,67 @@ def collect_installed_packages_information(self):
...
+        
+        ## Conduct some preprocessing before building the image
+        self.pre_build_configuration()
+        print(f"self.context {self.context}")
+
+        ## Start building the image
+        if self.context:
+            with open(self.context.context_path, "rb") as context_file:
+                print("within context")
+                self.docker_build(fileobj=context_file, custom_context=True)
```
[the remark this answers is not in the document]
tejaschumbalkar: good point

src/image.py
```diff
@@ -121,39 +157,50 @@ def build(self):
                 else:
                     response.append(str(line))
 
-            self.context.remove()
+        self.log = response
+        print(f"self.log {self.log}")
+        self.build_status = constants.SUCCESS
+        #TODO: return required?
```
tejaschumbalkar, line 163: nit: comment can be deleted

src/image.py
```diff
@@ -121,39 +157,50 @@ def build(self):
...
+        if self.summary["image_size"] > self.info["image_size_baseline"] * 1.20:
+            response.append("Image size baseline exceeded")
+            response.append(f"{self.summary['image_size']} > 1.2 * {self.info['image_size_baseline']}")
+            response += self.collect_installed_packages_information()
+            self.build_status = constants.FAIL_IMAGE_SIZE_LIMIT
+        else:
+            self.build_status = constants.SUCCESS
+        self.log = response
+        print(f"self.log {self.log}")
+        #TODO: return required?
```
tejaschumbalkar, line 181: nit: comment can be deleted

src/image.py
```diff
@@ -121,39 +157,50 @@ def build(self):
...
+            if line.get("stream") is not None:
+                response.append(line["stream"])
+            else:
+                response.append(str(line))
+
+        self.summary["status"] = constants.STATUS_MESSAGE[self.build_status]
+        self.summary["end_time"] = datetime.now()
+        self.summary["ecr_url"] = self.ecr_url
+        self.log = response
+        #TODO: return required?
```
tejaschumbalkar, line 205: nit: comment can be deleted

src/image_builder.py
```diff
@@ -154,109 +156,198 @@ def image_builder(buildspec):
...
-        image_object = DockerImage(
+        
+        #Create initial stage docker object
+        initial_stage_image_object = DockerImage(
             info=info,
             dockerfile=image_config["docker_file"],
             repository=image_repo_uri,
             tag=image_tag,
             to_build=image_config["build"],
+            stage=constants.INITIAL_STAGE,
```
tejaschumbalkar, line 167: Should be `PREVIOUS_STAGE` as mentioned earlier

src/image_builder.py
```diff
@@ -154,109 +156,198 @@ def image_builder(buildspec):
...
+
+    #initial stage standard images build
+    FORMATTER.banner("Standard Build")
+    build_images(standard_images)
+
+    #initial stage example images build
+    FORMATTER.banner("Example Build")
+    build_images(example_images)
+       
+    #Conclusion stage build
```
tejaschumbalkar, line 203: This stage must run after the `standard_images` to complete the build of standard images. The same must be done after building `example_images`

src/image_builder.py
```diff
@@ -154,109 +156,198 @@ def image_builder(buildspec):
...
+
+def get_conclusion_stage_image_object(initial_stage_image_object):
+    # Check if this is only required for mainline
+    # if build_context == "MAINLINE":
+    conclusion_stage_image_object = None
+    conclusion_stage_image_object = ConclusionStageImage(
+        info=initial_stage_image_object.info,
+        dockerfile=os.path.join(os.sep, os.getenv("ROOT_FOLDER_PATH"), "src", "Dockerfile.multipart"),
+        repository=initial_stage_image_object.repository,
+        tag=initial_stage_image_object.tag,
```
tejaschumbalkar, line 242: optionally we can also have a additional tag that will be applied to the `conclusion` stage image so that we can easily distinguish between a single/mulitstage stage DLC

src/image_builder.py
```diff
@@ -154,109 +156,198 @@ def image_builder(buildspec):
...
+    # the FORMATTER.progress(THREADS) function call also waits until all threads have completed
+    FORMATTER.progress(THREADS)
+
+#### TODO: Remove this entire method when https://github.com/boto/boto3/issues/1592 is resolved ####
+def get_dummy_boto_client():
+    # In absence of this method, the behaviour documented in https://github.com/boto/boto3/issues/1592 is observed.
+    # If this function is not added, boto3 fails because boto3 sessions are not thread safe.
+    # However, once a dummy client is created, it is ensured that the calls are thread safe.
+    import boto3
+    return boto3.client("secretsmanager", region_name=os.getenv('REGION'))
```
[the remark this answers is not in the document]
tejaschumbalkar: +1

src/metrics.py
```diff
@@ -40,6 +40,8 @@ def push_image_metrics(self, image):
             "device_type": image.device_type,
             "python_version": image.python_version,
             "image_type": image.image_type,
+            "image_stage": image.stage,
+            "push_required": str(image.to_push)
```
tejaschumbalkar, line 44: metrics of only the images that are qualified for a push should be uploaded? In that case what's the usecase for `push_required`?

test/dlc_tests/sanity/test_safety_check_v2.py
```diff
@@ -0,0 +1,99 @@
...
+                        failed_count += 1
+                        LOGGER.info(
+                            report_log_template.format(
+                                status="FAILED",
+                                pkg=report_item.package,
+                                installed=report_item.installed,
+                                vulnerabilities = report_item.vulnerabilities,
+                            )
+                        )
+        assert failed_count == 0, f"{failed_count} package/s failed safety test for {image} !!!"
```
tejaschumbalkar, line 96: PR description states `It will raise an exception if there is any vulnerability found in the package and mark the build as failed.` Which approach are we following here? Are we marking the test or the build as failed?

src/Dockerfile.multipart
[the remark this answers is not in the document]
arjkesh: +1 to `Docker.common`

src/Dockerfile.multipart
```diff
@@ -0,0 +1,7 @@
+# Use the Deep Learning Container as a base Image
+ARG INITIAL_STAGE_IMAGE=""
```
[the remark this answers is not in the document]
arjkesh: `PRE_PUSH_IMAGE` could work

src/conclusion_stage_image.py
```diff
@@ -0,0 +1,68 @@
...
+        ARTIFACTS.update(
+                    {
+                        "dockerfile": {
+                            "source": f"Dockerfile.multipart",
+                            "target": "Dockerfile",
+                        }
+                    }
+                )
+        
+        artifact_root = os.path.join(os.sep, os.getenv("ROOT_FOLDER_PATH"), "src") + "/"
```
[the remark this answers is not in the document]
arjkesh: No, don't add '/' to the path with os.path.join - `os.path.join('foo', 'bar', '/')` would generate `'/'` as a path. @shantanutrip - you should be able to remove the trailing` '/'` here - artifact_root will be used in a subsequent os.path.join statement

buildspec.yml
```diff
@@ -12,6 +12,7 @@ phases:
       - $(aws ecr get-login --no-include-email --region $AWS_DEFAULT_REGION)
       - pip install -r src/requirements.txt
       - bash src/setup.sh $FRAMEWORK
+      - export ROOT_FOLDER_PATH=$(pwd)
```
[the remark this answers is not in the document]
arjkesh: I don't think `$HOME` would necessary be the right env variable here, since this file is executed on CB. `CODEBUILD_SRC_DIR` should already be set, and you should be able to leverage that for most of your needs

src/utils.py
```diff
@@ -486,3 +487,64 @@ def set_test_env(images, images_env="DLC_IMAGES", **kwargs):
...
+    framework = ("mxnet" if "mxnet" in image_uri else
+                 "pytorch" if "pytorch" in image_uri else
+                 "tensorflow")
+    job_type = ("training" if "training" in image_uri else
+                "inference-eia" if "eia" in image_uri else
+                "inference-neuron" if "neuron" in image_uri else
+                "inference")
+    python_version = "py2" if "py2" in image_uri else "py3"
+
+    IGNORE_SAFETY_IDS = {}
```
arjkesh, line 506: Can you style this snake case?

test/dlc_tests/sanity/test_safety_check_v2.py
arjkesh: No, don't make a distinction based on canary/pipelines for this test - that's the job of markers. Just be descriptive about what this test actually does - you can call it `test_safety_file` or something that describes what it does.

test/dlc_tests/sanity/test_safety_check_v2.py
```diff
@@ -0,0 +1,99 @@
...
+class SafetyPythonEnvironmentVulnerabilityReport:
+    report: List[SafetyPackageVulnerabilityReport]
+
+    def __post_init__(self):
+        self.report = [SafetyPackageVulnerabilityReport(**i) for i in self.report]
+
+
+
+@pytest.mark.model("N/A")
+@pytest.mark.skipif(
```
arjkesh, line 53: This is not necessary, because the test is currently not marked as canary, so it will not run anyway.

src/safety_report_generator.py
arjkesh, line 49: nit: rename `vid` to `vulnerability_id` for clarity

src/conclusion_stage_image.py
```diff
@@ -0,0 +1,68 @@
...
+        the safety_report which is then baked into the image. 
+        """
+        ## Call the pre_build_configuration steps from the parent class
+        super(ConclusionStageImage, self).pre_build_configuration()
+        ## Generate safety scan report for the first stage image and add the file to artifacts
+        first_stage_image_uri = self.build_args['INITIAL_STAGE_IMAGE']
+        processed_image_uri = first_stage_image_uri.replace('.','-').replace('/','-').replace(':','-')
+        image_name = self.name
+        tarfile_name_for_context = f"{processed_image_uri}-{image_name}"
+        storage_file_path = f"{os.getenv('ROOT_FOLDER_PATH')}/src/{tarfile_name_for_context}_safety_report.json"
```
arjkesh, line 40: use os.path.join on this one - also instead of calling `os.getenv('ROOT_FOLDER_PATH') everywhere, you can create a utility function and use CODEBUILD_SRC_DIR when available, and have some fallback option when not available. Example:

```
def get_root_folder_path():
    root_dir_pattern = re.compile(r"^(\S+deep-learning-containers)")
    pwd = os.getcwd()
    return os.getenv('CODEBUILD_SRC_DIR', root_dir_pattern.match(pwd).group(1))
```

arjkesh (commented):
Please swap out `print` statements with `logging` statements if you're trying to write log info to the console

src/conclusion_stage_image.py
```diff
@@ -0,0 +1,68 @@
...
+from image import DockerImage
+from context import Context
+from utils import generate_safety_report_for_image
+
+import os
+
+class ConclusionStageImage(DockerImage):
+    """
+    Class designed to handle the ConclusionStageImages
+    """
```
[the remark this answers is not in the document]
saimidu: It makes sense that it is the "last stage" of the docker build, but this code seems like it can be reused for other kinds of multi-stage builds. Besides that, we don't know for sure if this will always be the last stage of a docker image build forever. We may add more layers in the future.

The issue here is that a small decision like that would result in the code either becoming confusing because "Conclusion" is no longer the last stage, or having to be revamped pretty heavily to change the names of classes and objects related to this. I would like to make this future proof by not making the assumption that the safety report layer will always be the "concluding" stage.

src/image_builder.py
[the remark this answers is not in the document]
shantanutrip: I think that there is a lot of confusion regarding this issue, like how its reproduced etc. Also, I think the comment that I have made looks slightly wrong in the very first go.

In the [link](https://github.com/boto/boto3/issues/1592) that I have shared in the comments, the user just uses `boto3.client` and gets the error. Further, on reading [this](https://github.com/boto/botocore/issues/1246#issuecomment-632444392) comment, I think I could say the following: 
(1) The clients are also not completely thread safe. 
(2) If we use a single session and clients for that session run on different threads, we might run the risk of facing this error.

I think the underlying concept is that when we use just clients, they internally go and create a `default session` ([code](https://github.com/boto/boto3/blob/fed71d95263093f9d9aeed3650ebe875c9e5dc60/boto3/__init__.py#L87)). In future, if we use another client, I assume it uses the same default session that is initialized for the very first time.

src/image.py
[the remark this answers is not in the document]
shantanutrip: Thanks, will do.

src/image_builder.py
```diff
@@ -154,109 +156,198 @@ def image_builder(buildspec):
...
+        conclusion_stage_image_object = get_conclusion_stage_image_object(initial_stage_image_object)
 
-        # the FORMATTER.progress(THREADS) function call also waits until all threads have completed
-        FORMATTER.progress(THREADS)
+        FORMATTER.separator()
 
-        for image in example_images:
-            THREADS[image.name] = executor.submit(image.build)
+        INITIAL_STAGE_IMAGES.append(initial_stage_image_object)
+        if conclusion_stage_image_object is not None:
```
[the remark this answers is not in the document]
shantanutrip: I will make the change. I think that has been left by mistake.

src/conclusion_stage_image.py
shantanutrip: Thanks, will do. Follow up question: I can't find `CODEBUILD_SRC_DIR` in the list of already set env variables in the CodeBuild image

test/dlc_tests/sanity/test_safety_check_v2.py
shantanutrip: We are making the sanity tests as failed. I will make the desired change in the desctiption.

src/safety_report_generator.py
arjkesh: +1 - we should be consistent in styling of these statuses as well. I'm fine if you want to replace with "TBD" since that's a fairly universal acronym.

Also should be spaces in between var, = and value assigned. This would be fixed by running `black -l 120` as @saimidu suggested.

src/conclusion_stage_image.py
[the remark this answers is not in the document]
saimidu: Sure. That should be good. Please also make add this explanation to the docstring for the class, so that it is readily apparent what the class is meant to do.

src/image_builder.py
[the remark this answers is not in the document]
saimidu: Sure, but what is the advantage of creating clients here in each thread even though they remain unused? Wouldn't each client be creating Sessions of its own in parallel with all other threads?

src/image_builder.py
```diff
@@ -154,109 +156,198 @@ def image_builder(buildspec):
...
+            THREADS[image.name] = executor.submit(image.build)
+    # the FORMATTER.progress(THREADS) function call also waits until all threads have completed
+    FORMATTER.progress(THREADS)
+
+#### TODO: Remove this entire method when https://github.com/boto/boto3/issues/1592 is resolved ####
+def get_dummy_boto_client():
+    # In absence of this method, the behaviour documented in https://github.com/boto/boto3/issues/1592 is observed.
+    # If this function is not added, boto3 fails because boto3 sessions are not thread safe.
+    # However, once a dummy client is created, it is ensured that the calls are thread safe.
+    import boto3
```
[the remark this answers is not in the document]
saimidu: `boto3` is a module that is common enough that we can reasonably expect it to always be available in the environment when performing builds (to be confirmed by checking `src/requirements.txt`.

We can always remove `boto3` later if it is unused. Most Python IDEs highlight unused imports so that developers can easily discover them and decide whether to remove those imports.

buildspec.yml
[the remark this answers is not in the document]
shantanutrip: Removed. Followed Arjuna's suggestions.

src/constants.py
```diff
@@ -71,3 +75,6 @@
 ECS_TESTS = "ecs"
 EKS_TESTS = "eks"
 ALL_TESTS = ["sagemaker", "ec2", "eks", "ecs"]
+
+API_CLIENT_TIMEOUT = 600
```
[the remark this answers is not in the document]
shantanutrip: Added

src/image.py
shantanutrip: I haven't seen any yet. It was already existing in the code, so I thought of not playing with it.

src/conclusion_stage_image.py
```diff
@@ -0,0 +1,68 @@
...
+                    {
+                        "dockerfile": {
+                            "source": f"Dockerfile.multipart",
+                            "target": "Dockerfile",
+                        }
+                    }
+                )
+        
+        artifact_root = os.path.join(os.sep, os.getenv("ROOT_FOLDER_PATH"), "src") + "/"
+        return Context(ARTIFACTS, context_path=f'build/{tarfile_name}.tar.gz',artifact_root=artifact_root)
```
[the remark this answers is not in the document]
shantanutrip: Done

src/conclusion_stage_image.py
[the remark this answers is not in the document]
shantanutrip: Done

src/constants.py
```diff
@@ -31,6 +31,10 @@
 # Left and right padding between text and margins in output
 PADDING = 1
 
+#Docker build stages
+INITIAL_STAGE="initial"
+CONCLUSION_STAGE="conclusion"
```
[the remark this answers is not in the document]
shantanutrip: Done

src/image.py
```diff
@@ -121,39 +157,50 @@ def build(self):
                 else:
                     response.append(str(line))
 
-            self.context.remove()
+        self.log = response
+        print(f"self.log {self.log}")
```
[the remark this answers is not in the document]
shantanutrip: I have removed this log display line. Also, I have changed the way in which logs are now displayed and made it such that it doesn't take up a lot of space.

src/image.py
```diff
@@ -69,34 +72,67 @@ def collect_installed_packages_information(self):
...
+            self.build_args["INITIAL_STAGE_IMAGE"] = self.ecr_url
+
+        if self.info.get("extra_build_args"):
+            self.build_args.update(self.info.get("extra_build_args"))
+        
+        if self.info.get("labels"):
+            self.labels.update(self.info.get("labels"))
+        
+        print(f"self.build_args {self.build_args}")
+        print(f"self.labels {self.labels}")
```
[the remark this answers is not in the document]
shantanutrip: I will remove this.

src/image.py
```diff
@@ -69,34 +72,67 @@ def collect_installed_packages_information(self):
         docker_client.containers.prune()
         return command_responses
 
+    def pre_build_configuration(self):
```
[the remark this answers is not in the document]
shantanutrip: Done

src/conclusion_stage_image.py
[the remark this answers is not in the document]
shantanutrip: Done

src/image_builder.py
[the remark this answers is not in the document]
shantanutrip: Done.

src/image.py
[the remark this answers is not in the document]
shantanutrip: Done.

src/safety_report_generator.py
shantanutrip: Done

src/safety_report_generator.py
shantanutrip: Done.

src/safety_report_generator.py
shantanutrip: Done.

src/safety_report_generator.py
shantanutrip: Done.

src/image_builder.py
shantanutrip: Done.

src/image_builder.py
shantanutrip: Done.

src/image_builder.py
shantanutrip: Done.

src/image_builder.py
shantanutrip: Done.

src/image_builder.py
shantanutrip: Done.

src/image_builder.py
shantanutrip: Done.

src/image_builder.py
shantanutrip: Done.

src/image_builder.py
shantanutrip: Done.

src/safety_report_generator.py
shantanutrip: I have changed the names now.

src/safety_report_generator.py
shantanutrip: Made `TBD` string literal

src/safety_report_generator.py
shantanutrip: Done

src/utils.py
shantanutrip: Done

src/utils.py
shantanutrip: Done

test/dlc_tests/sanity/test_safety_check_v2.py
shantanutrip: Done.

test/dlc_tests/sanity/test_safety_check_v2.py
shantanutrip: Thanks. Done.

test/dlc_tests/sanity/test_safety_check_v2.py
shantanutrip: Changed. List was not required.

test/dlc_tests/sanity/test_safety_check_v2.py
shantanutrip: Done

src/Dockerfile.multipart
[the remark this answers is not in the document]
shantanutrip: Done

src/constants.py
```diff
@@ -71,3 +75,6 @@
 ECS_TESTS = "ecs"
 EKS_TESTS = "eks"
 ALL_TESTS = ["sagemaker", "ec2", "eks", "ecs"]
+
+API_CLIENT_TIMEOUT = 600
+MAX_WORKER_COUNT_FOR_PUSHING_IMAGES = 3
```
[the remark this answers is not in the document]
shantanutrip: There is no specific reason. Just that pushing a lot of images altogether leads to unnecessary delay, so added this.

src/image.py
```diff
@@ -69,34 +72,67 @@ def collect_installed_packages_information(self):
...
             return self.build_status
+        
+        ## Conduct some preprocessing before building the image
+        self.pre_build_configuration()
+        print(f"self.context {self.context}")
+
+        ## Start building the image
+        if self.context:
+            with open(self.context.context_path, "rb") as context_file:
+                print("within context")
```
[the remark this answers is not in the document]
shantanutrip: Done.

src/image.py
```diff
@@ -69,34 +72,67 @@ def collect_installed_packages_information(self):
...
+                self.docker_build(fileobj=context_file, custom_context=True)
+                self.context.remove()  
+        else:
+            print("out of context")
```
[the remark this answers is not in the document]
shantanutrip: Done.

src/image.py
shantanutrip: Done.

src/image.py
shantanutrip: Done.

src/image_builder.py
shantanutrip: Done.

src/conclusion_stage_image.py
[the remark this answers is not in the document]
tejaschumbalkar: @shantanutrip Can you explain the reason behind `I agree that we should remove the assumption that this is the last stage of the docker build.` ? This docker layer is meant to be the final stage in the build process as per the design discussions. I agree that we can add more layers on top of the common layer but what is the usecase for that?

Ideally, we want the common components to run at the very last stage of the build. Eg, we want to run ECR scan/safety report/oss compliance licenses on the final layer of the docker build so that there are no new components in the docker added after this final layer.

src/image_builder.py
tejaschumbalkar: I dont think there is will be an issue since we will be adding a new tag and not modifying an existing tag.

src/image_builder.py
tejaschumbalkar: Can we run the CommonStage after the respective builds for standard/example image is complete. This will reduce the build time for the job, especially when there will be more images to build when SM vs DIY is introducted

src/image_builder.py
shantanutrip: Done.

src/image_builder.py
shantanutrip: I am not very sure how it reduces the build time for the job. Overall, the number of images being built and pushed will remain the same, right?

src/metrics.py
shantanutrip: The images that are qualified for push are just the Common Stage images. The time to process the Common Stage images is not same as the time taken to process Pre_push images. In fact, the time taken to process CommonStage images is quite small as opposed to Pre_push images. Therefore, I thought of segregating the buckets by adding extra dimension so that we can monitor things correctly.

src/image_builder.py
shantanutrip: Add `-`

src/constants.py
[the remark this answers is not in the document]
tejaschumbalkar: Would this configuration needed as we are trying to push the images in parallel?

src/image.py
tejaschumbalkar: Existing code as in on the PR branch or the master, if PR branch then lets just remove it, probably it might have been added for experimental purposes.

src/image_builder.py
tejaschumbalkar: Lets update the new build process as discussed in the PR description as well

src/utils.py
shantanutrip: I have made it future proof to the extent possible.  According to my understanding the job_type cannot be made completely future proof because inference type images have not been further divided into inference-eia and inference-neuron in the buildspec files.

src/constants.py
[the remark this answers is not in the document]
shantanutrip: `MAX_WORKER_COUNT_FOR_PUSHING_IMAGES` just states that whenever we push images in parallel, we do not push more than 3 images.

src/image.py
shantanutrip: Sure, I will remove it.

src/image_builder.py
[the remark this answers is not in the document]
shantanutrip: Done

src/image_builder.py
shantanutrip: Done

src/metrics.py
shantanutrip: Done

src/conclusion_stage_image.py
shantanutrip: Done

src/image.py
```diff
@@ -70,6 +73,22 @@ def is_child_image(self):
...
+        """
+        return self._corresponding_common_stage_image
+
+    @corresponding_common_stage_image.setter
+    def corresponding_common_stage_image(self, docker_image_object):
+        """
+        Sets the value for the corresponding_common_stage_image variable.
+        """
+        if self.to_push:
+            raise ValueError("Corresponding common stage image can only exist if the image is non-pushable")
```
saimidu, line 89: Why can this only exist if the image is non-pushable?

saimidu (approved):

src/common_stage_image.py
```diff
@@ -0,0 +1,69 @@
...
+    def generate_common_stage_context(self, safety_report_path, tarfile_name="common-stage-file"):
+        """
+        For CommonStageImage, build context is built once the safety report is generated. This is because
+        the Dockerfile.common uses this safety report to COPY the report into the image.
+        """
+        artifacts = {
+            "safety_report": {"source": safety_report_path, "target": "safety_report.json",},
+            "dockerfile": {
+                "source": os.path.join(
+                    os.sep, get_root_folder_path(), "miscellaneous_dockerfiles", "Dockerfile.common",
```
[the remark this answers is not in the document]
shantanutrip: Removed

src/common_stage_image.py
```diff
@@ -0,0 +1,69 @@
...
+                ),
+                "target": "Dockerfile",
+            },
+        }
+
+        artifact_root = os.path.join(os.sep, get_root_folder_path(), "src")
+        return Context(artifacts, context_path=f"build/{tarfile_name}.tar.gz", artifact_root=artifact_root,)
```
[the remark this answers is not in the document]
shantanutrip: Removed

src/image.py
shantanutrip: Changes as per conversation.

src/common_stage_image.py
```diff
@@ -0,0 +1,69 @@
...
+        )
+        self.context = self.generate_common_stage_context(storage_file_path, tarfile_name=tarfile_name_for_context)
+
+    def generate_common_stage_context(self, safety_report_path, tarfile_name="common-stage-file"):
+        """
+        For CommonStageImage, build context is built once the safety report is generated. This is because
+        the Dockerfile.common uses this safety report to COPY the report into the image.
+        """
+        artifacts = {
+            "safety_report": {"source": safety_report_path, "target": "safety_report.json",},
```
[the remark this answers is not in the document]
shantanutrip: Removed

src/common_stage_image.py
```diff
@@ -0,0 +1,69 @@
...
+        """
+        # Call the update_pre_build_configuration steps from the parent class
+        super(CommonStageImage, self).update_pre_build_configuration()
+        # Generate safety scan report for the first stage image and add the file to artifacts
+        pre_push_stage_image_uri = self.build_args["PRE_PUSH_IMAGE"]
+        processed_image_uri = pre_push_stage_image_uri.replace(".", "-").replace("/", "-").replace(":", "-")
+        image_name = self.name
+        tarfile_name_for_context = f"{processed_image_uri}-{image_name}"
+        storage_file_path = os.path.join(
+            os.sep, get_root_folder_path(), "src", f"{tarfile_name_for_context}_safety_report.json",
```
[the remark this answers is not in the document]
shantanutrip: Removed

data/ignore_ids_safety_scan.json
```diff
@@ -0,0 +1,138 @@
...
+        "training": {
+            "_comment":"py2 is deprecated",
+            "py2": {
+                "38449":"for shipping pillow<=6.2.2 - the last available version for py2",
+                "38450":"for shipping pillow<=6.2.2 - the last available version for py2",
+                "38451":"for shipping pillow<=6.2.2 - the last available version for py2",
+                "38452":"for shipping pillow<=6.2.2 - the last available version for py2",
+                "35015":"for shipping pycrypto<=2.6.1 - the last available version for py2"
+            },
+            "py3": {
```
tejaschumbalkar, line 12: Should this contain the python minor version as well? What happens when a DLC has 2 python version support?

src/image_builder.py
```diff
@@ -155,109 +159,259 @@ def image_builder(buildspec):
...
+        COMMON_STAGE_IMAGES.append(common_stage_image_object)
+        FORMATTER.separator()
 
     FORMATTER.banner("DLC")
-    FORMATTER.title("Status")
 
-    THREADS = {}
+    # Standard images must be built before example images
+    # Example images will use standard images as base
+    # Common images must be built at the end as they will consume respective standard and example images
```
tejaschumbalkar, line 188: This needs to be reworded

src/image_builder.py
```diff
@@ -155,109 +159,259 @@ def image_builder(buildspec):
...
+    standard_images = [image for image in PRE_PUSH_STAGE_IMAGES if "example" not in image.name.lower()]
+    example_images = [image for image in PRE_PUSH_STAGE_IMAGES if "example" in image.name.lower()]
+    ALL_IMAGES = PRE_PUSH_STAGE_IMAGES + COMMON_STAGE_IMAGES
+    IMAGES_TO_PUSH = [image for image in ALL_IMAGES if image.to_push and image.to_build]
```
tejaschumbalkar, line 192: Can the `IMAGES_TO_PUSH` be `COMMON_STAGE_IMAGES` ?

src/image.py
```diff
@@ -143,12 +145,9 @@ def build(self):
...
         # Start building the image
-        if self.context:
-            with open(self.context.context_path, "rb") as context_file:
-                self.docker_build(fileobj=context_file, custom_context=True)
-                self.context.remove()
-        else:
-            self.docker_build()
+        with open(self.context.context_path, "rb") as context_file:
+            self.docker_build(fileobj=context_file, custom_context=True)
+            self.context.remove()
```
saimidu: Please undo this change, because we might have images being built in the future that do not have any files to be included through build context, even if we do not right now.

saimidu (commented):
Approved.

src/image_builder.py
shantanutrip: Synced up

src/image_builder.py
shantanutrip: Synced up

data/ignore_ids_safety_scan.json
shantanutrip: Synced up

src/image.py
```diff
@@ -260,9 +266,43 @@ def push_image(self):
...
+            self.summary["pushed_uris"] = []
+        self.summary["pushed_uris"].append(f"{self.repository}:{tag}")
+        response.append(f"Completed Push for {self.repository}:{tag}")
         self.log.append(response)
 
         LOGGER.info(f"DOCKER PUSH LOGS: \n {self.get_tail_logs_in_pretty_format(2)}")
-        LOGGER.info(f"Completed Push for {self.name}")
+        return self.build_status
+
+    def push_image_with_additional_tags(self):
```
tejaschumbalkar, line 278: Why do we need at separate function to push a additonal tag? Why cant be this done in `push_image`

src/image_builder.py
```diff
@@ -229,7 +228,11 @@ def process_images(pre_push_image_list, pre_push_image_type="Pre-push"):
     pre push images and then builds it. After the pre-push images have been built, it extracts the 
     corresponding common stage images for the pre-push images and builds those common stage images.
     After the common stage images have been built, it finds outs the docker images that need to be 
-    pushed and pushes them according.
+    pushed and pushes them accordingly.
+
+    Note that the common stage images should always be built after the pre-push images of a
+    particular kind. This is because the Common stage images use are built on respctive
```
tejaschumbalkar, line 234: nit: `respctive` spell

src/image.py
tejaschumbalkar: Seems that we need to tag the docker image first and push to ECR later.

tejaschumbalkar:
Changes looks good. The ECR images are tagged as expected. Let's comment the safety functionality as discussed.

tejaschumbalkar:
Soft Approved!

tejaschumbalkar:
PR is good to merge. Thanks for adding the functionality to our build system @shantanutrip !

shantanutrip (approved):

PR #1220, opened 2021-07-09
[build] Fix MXNet 1.8 dockerfile paths

saimidu:
*Issue #, if available:*

## PR Checklist

## Reviewer Checklist

*Description:*
The addition of a new dockerfile for `mxnet/inference/1.8/py3/Dockerfile.neuron` causes the helper function to find dockerfiles to fail, as the CPU/GPU dockerfiles for inference are located in `mxnet/inference/1.8.0/py3/`. This PR fixes that failure.

*Tests run:*

*DLC image/dockerfile:*

*Additional context:*


tejaschumbalkar (approved):

junpuf (approved):

arjkesh:
Is there a reason to move all Dockerfiles into a major.minor scheme instead of moving neuron into major.minor.patch?

In some of our label sanity tests, we check for the presence of all DLC major versions for a framework by walking through directories with major.minor*. So in the example where 1.8.1 is released and we need to update the DLC major version, and we do it in place, this will cause the test to fail. So either the test needs to be updated, or we should keep the files in a 1.8.0 dir.

saimidu:
We only update the DLC Release Major Version when there are backward incompatible changes between the Patch Version, such as between PyTorch `1.5.0` and `1.5.1`. It is more convenient to store those dockerfiles separately, rather than add exceptions to the test logic.
However, in the normal case, we never update the DLC Release Major Version between Patch Version updates on the framework, such as between TensorFlow `2.3.0`, `2.3.1`, and `2.3.2`. In these cases, the latest patch version is the only version that is relevant, which means that we do not need the older framework patch version dockerfiles to still exist.
Besides, we can still use the `major.minor.patch` version folders for dockerfiles, where needed. This PR only addresses the problem of finding dockerfiles corresponding to a particular image when there are both `major.minor` as well as `major.minor.patch` folders for the same framework version, such as having both `1.8` and `1.8.0` in the folder tree.

arjkesh:
> We only update the DLC Release Major Version when there are backward incompatible changes between the Patch Version, such as between PyTorch `1.5.0` and `1.5.1`. It is more convenient to store those dockerfiles separately, rather than add exceptions to the test logic.
> 
> However, in the normal case, we never update the DLC Release Major Version between Patch Version updates on the framework, such as between TensorFlow `2.3.0`, `2.3.1`, and `2.3.2`. In these cases, the latest patch version is the only version that is relevant, which means that we do not need the older framework patch version dockerfiles to still exist.
> 
> Besides, we can still use the `major.minor.patch` version folders for dockerfiles, where needed. This PR only addresses the problem of finding dockerfiles corresponding to a particular image when there are both `major.minor` as well as `major.minor.patch` folders for the same framework version, such as having both `1.8` and `1.8.0` in the folder tree.

The main question I had was what is the reasoning behind favoring the major.minor dir? I'm fine with it if it's meant to house the latest framework version, but we should have a policy in place for what to do, because it can be confusing to users to see multiple directories, some with major.minor, some with major.minor.patch, etc.

PR #1241, opened 2021-07-20
[build]|[test] Adds Web scraper Module and handles automations related to ECR scan.

shantanutrip:
*Issue #, if available:*

## PR Checklist

## Reviewer Checklist

*Description:*
The following PR consists of the Web Scraper that is required to scratch the data related to each CVE. In addition to that, it add the allowlists for MxNet 1.8 - training and inference images and handles the automations revolving around security.


scraped_data.json
```diff
@@ -0,0 +1,4 @@
+[
```
shantanutrip, line 1: This is how the output JSON would look like. This file will be deleted in the final merge.

web_scraper/web_scraper/items.py
```diff
@@ -0,0 +1,12 @@
+# Define here the models for your scraped items
```
shantanutrip, line 1: Framework Code- Automatically created

web_scraper/web_scraper/middlewares.py
```diff
@@ -0,0 +1,103 @@
+# Define here the models for your spider middleware
```
shantanutrip, line 1: Framework Code- Automatically created

web_scraper/web_scraper/pipelines.py
```diff
@@ -0,0 +1,13 @@
+# Define your item pipelines here
```
shantanutrip, line 1: Framework Code- Automatically created

web_scraper/web_scraper/settings.py
```diff
@@ -0,0 +1,88 @@
+# Scrapy settings for web_scraper project
```
shantanutrip, line 1: Framework Code- Automatically created

web_scraper/web_scraper/spiders/__init__.py
```diff
@@ -0,0 +1,4 @@
+# This package will contain the spiders of your Scrapy project
```
shantanutrip, line 1: Framework Code- Automatically created

scraped_data.json
saimidu: Is it possible to write to this file in such a way that there is good indentation on the JSON text?

Something like:
```
[
    {
        "URL": "...",
        "status_tables": [
            {
                "packages": [...],
                "release_states": [...]
            }
        ],
        "notes": [...]
    }
]
```

shantanutrip: Sure, there should be. I will take a look into it.

tejaschumbalkar: Can we format this file with json indentation to understand the file format better for review

web_scraper/web_scraper/spiders/cve_spiders.py
```diff
@@ -0,0 +1,92 @@
+import scrapy
+import os
+
+
+class QuotesSpider(scrapy.Spider):
```
junpuf, line 5: It looks like this class is suitable for fetch data from the type of web page with url starts with `https://ubuntu.com/security/CVE-`. If we pass in any other url for example `https://www.youtube.com`, then this class will not work for its purpose. This means that each spider class will be responsible for one type of webpage (which is a good idea i believe). Here are some recommendations because of the 1:1 relationship of the spider class and the type of webpage.

1. Using a class level docstring to explain that this is used to retrieve information from `https://ubuntu.com/security/CVE-` webpages. Also, from the docstring, state the output file format.
2. After getting the `SCRAPE_URL_LIST` from the environment variable, validate the url format.

On a side note: Could you comment on why this class is called QuotesSpider as opposed to something like UbuntuCVEPageSpider?

web_scraper/web_scraper/pipelines.py
junpuf: So these files are auto-generated and we shouldn't change it right?

web_scraper/web_scraper/spiders/cve_spiders.py
```diff
@@ -0,0 +1,92 @@
...
+    name = "cve" ## Spider name should be unique
```
junpuf, line 6: Maybe something more descriptive such as `ubuntu_cve_page_spider`? There are different types of CVEs we are likely going to deal with in the future, so we may also have something like `amazon_linux_cve_page_spider`, `pypi_cve_page_spider` etc.

web_scraper/web_scraper/spiders/cve_spiders.py
junpuf: After think about this again, we could also reuse this class to crawl other webpages as well, we are likely need a mapping like below in that case.

```
{
      "https://ubuntu.com/security/CVE-": parse_ubuntu_cve_page,
       ....
}
```

web_scraper/README.md
```diff
@@ -0,0 +1,23 @@
...
+To run the code, following steps can be taken:
+
+**Note: all the commands given below should be run from this directory i.e. `~/deep-learning-containers/web_scraper` itself.**
+
+1. Set the following environment variable:
+`export SCRAPE_URL_LIST="https://ubuntu.com/security/CVE-2016-1585 https://ubuntu.com/security/CVE-2021-29973"`. <br> The `SCRAPE_URL_LIST` environment variable consists of the list of space seperated CVE URLs to scrape. 
+
+2. In the deep-learning-containers root folder, remove the file scraped_data using the following command `rm -r ../scraped_data.json`.
+
+3. Start the crawl by using the following command `scrapy crawl cve -o ../scraped_data.json`
```
junpuf, line 20: is it possible to send input to the spider from the command line interface rather than using environment?

scraped_data.json
shantanutrip: Done

web_scraper/README.md
shantanutrip: There is a way to pass arguments. I am trying to figure out a way to directly call the crawler class from within the script. That would take away the need of all the command line operations.

web_scraper/web_scraper/pipelines.py
shantanutrip: Yes

web_scraper/web_scraper/spiders/cve_spiders.py
shantanutrip: This is a very good thing to do. Will do it.

web_scraper/test_script.py
```diff
@@ -0,0 +1,15 @@
...
+from scrapy.crawler import CrawlerRunner
+## crochet plays with Twisted's Reactors to avoid ReactorNotRestartable error
+from crochet import setup 
+setup()
+
+def run_spider(spiderClass):
+    crawler = CrawlerRunner()
+    crawler.crawl(spiderClass)
+
+######################## How to run #####################
```
shantanutrip, line 11: @junpuf @saimidu This is a test script to show how one could call the crawler from within the script.

junpuf: great! this will allow us to use existing testing workflow for this as well.

junpuf (commented):
@shantanutrip any thought about this comment? https://github.com/aws/deep-learning-containers/pull/1241#discussion_r673703385

test/test_utils/security.py
```diff
@@ -63,6 +64,53 @@ def construct_allowlist_from_ecr_scan_result(self, vulnerability_list):
...
+        We want to first sort the innermost list of dicts based on the "name" of each dict and then we sort the
+        otermost dict based on keys i.e. package_name1 and package_name2
+        :param input_dict: dict(key, list(dict)), represents vulnerability_list
+        :return: dict, input_dict sorted in a custom way
+        """
+        copy_dict = copy.deepcopy(input_dict)
+        for _,list_of_dict in copy_dict.items():
+            list_of_dict.sort(key=lambda dict_element:dict_element["name"])
+        od = collections.OrderedDict(sorted(copy_dict.items()))
+        return dict(od)
```
SergTogul, line 107: Not sure I understand this. You sort `list_of_dict` and think that's enough. 
Then you sort copy_dict but convert it to `OrderedDict` to preserve the sorted order.
But after that you convert `OrderedDict` into a regular dict. 

So I think you should either don't use OrderedDict at all or convert everything into OrderedDict.

testspec.yml
```diff
@@ -22,6 +22,7 @@ phases:
       - pip install scheduler/.
       - echo Running pytest $TEST_TYPE tests on $DLC_IMAGES...
       - export PYTHONPATH=$PYTHONPATH:$(pwd)/src
+      - export UPGRADE_REPO_NAME=trshanta-repo
```
SergTogul, line 25: Do we still need it?

web_scraper/README.md
```diff
@@ -0,0 +1,23 @@
+# Webscraper
```
SergTogul, line 1: Do we really need it inhere? Maybe we could move all that code into dlc package? To keep only directly related code in the repo?

test/dlc_tests/sanity/test_security_check.py
```diff
@@ -54,6 +66,297 @@ def run_scan(ecr_client, image):
...
+    """
+    This method sets the value for MINIMUM_SEV_THRESHOLD
+
+    :param image: str Image URI for which threshold has to be set
+    """
+    global MINIMUM_SEV_THRESHOLD
+    if is_image_covered_by_allowlist_feature(image):
+        MINIMUM_SEV_THRESHOLD = "MEDIUM"
+        return
+    MINIMUM_SEV_THRESHOLD = "HIGH"
```
saimidu, line 289: How do global variables work in the context of multithreaded pytest executions for multiple images? Are the global variables within each script managed separately for each thread?

test/dlc_tests/sanity/test_security_check.py
```diff
@@ -54,6 +66,297 @@ def run_scan(ecr_client, image):
...
 
+def get_new_image_uri(image):
+    """
+    Returns the new image uri for the image being tested. After running the apt commands, the
+    new image will be uploaded to the ECR based on the new image uri.
+
+    :param image: str
+    :param new_image_uri: str
+    """
+    repository_name = os.getenv("UPGRADE_REPO_NAME")
```
saimidu, line 77: What is `UPGRADE_REPO_NAME`? Please add a comment explaining what it does.

Does it exist temporarily until some supporting infrastructure is set up for this code to work as intended?

test/dlc_tests/sanity/test_security_check.py
```diff
@@ -54,6 +66,297 @@ def run_scan(ecr_client, image):
         raise TimeoutError(f"ECR Scan is still in {scan_status} state. Exiting.")
 
 
+def get_new_image_uri(image):
```
saimidu, line 69: Please move all the helper functions into a separate file, so that we can keep only tests here. This test file has gotten too long, now, and seems to mostly be just helper code.

I would recommend moving these to either `test_utils/__init__.py` or `test_utils/security.py`.

test/dlc_tests/sanity/test_security_check.py
```diff
@@ -54,6 +66,297 @@ def run_scan(ecr_client, image):
...
+    """
+    Returns the new image uri for the image being tested. After running the apt commands, the
+    new image will be uploaded to the ECR based on the new image uri.
+
+    :param image: str
+    :param new_image_uri: str
+    """
+    repository_name = os.getenv("UPGRADE_REPO_NAME")
+    ecr_account = f"{os.getenv('ACCOUNT_ID')}.dkr.ecr.{os.getenv('REGION')}.amazonaws.com"
```
saimidu, line 78: When the `REGION` env variable is unset, this workflow will completely fail. Please default to a known preset value for environment variables if doing so is legal: `region = os.getenv("REGION", test_utils.DEFAULT_REGION)`.

When the `ACCOUNT_ID` env variable is unset, this workflow will completely fail. Please use `sts_client.get_caller_identity()` to obtain the account ID of the IAM role/user credentials configured.
```
sts_client = boto3.client("sts", region_name=region)
account_id = sts_client.get_caller_identity().get("Account")
```

The string being constructed is the ECR registry, rather than the ECR account. This method of constructing the registry string will fail for other partitions of AWS regions. Please use the following code to get the registry in a region-agnostic manner using pre-existing APIs:
```
endpoint_data = _botocore_resolver().construct_endpoint("ecr", region)
registry = "{}.dkr.{}".format(account, endpoint_data["hostname"])
```

test/dlc_tests/sanity/test_security_check.py
```diff
@@ -54,6 +66,297 @@ def run_scan(ecr_client, image):
...
+    upgraded_image_tag = "-".join(image.replace("/", ":").split(":")[1:]) + "-up"
```
saimidu, line 79: Is this meant to reupload the image to a known ECR repo to be scanned and validated for the updated packages?

Please do not use string manipulation with `.join`, `.replace`, and `.split` to create the image tag in a single line of code, as it is confusing, and takes time to understand. Please break this up into a clear set of parts, and then join those parts in the way you need it to be.

For example:
```
image_registry, image_repository_and_tag = image_uri.split("/")
image_repository, image_tag = image_repository_and_tag.split(":")
```
or even better, since you don't need the original image registry at all:
```
image_repository, image_tag = test_utils.get_repository_and_tag_from_image_uri(image_uri)
```

test/dlc_tests/sanity/test_security_check.py
```diff
@@ -54,6 +66,297 @@ def run_scan(ecr_client, image):
...
+    :param image: str
+    :param new_image_uri: str
+    """
+    ctx = Context()
+    docker_run_cmd = f"docker run -id --entrypoint='/bin/bash' {image}"
+    container_id = ctx.run(f"{docker_run_cmd}", hide=True, warn=True).stdout.strip()
+    apt_command = "apt-get update && apt-get upgrade"
+    docker_exec_cmd = f"docker exec -i {container_id}"
+    attempt_count = 0
+    apt_ran_successfully_flag = True
```
saimidu, line 98: If the default state of this flag is `True`, then any execution that exits from the while loop in an in incorrect way could leave the flag in its default state. The flag would state that the upgrade ran successfully, while the loop has in fact never succeeded. This could happen especially in cases where errors are ignored and outputs are hidden, as done with `ctx.run(..., hide=True, warn=True)` below.

Please start by setting `apt_ran_successfully_flag` to `False`.

test/dlc_tests/sanity/test_security_check.py
```diff
@@ -54,6 +66,297 @@ def run_scan(ecr_client, image):
...
+                f"Stdout is {run_output.stdout} \n"
+                f"Stderr is {run_output.stderr} \n"
+                f"Failed status is {run_output.exited}"
+            )
+            sleep(2 * 60)
+        elif run_output.ok:
+            break
+        if attempt_count == 10:
+            if not run_output.ok:
+                apt_ran_successfully_flag = False
```
saimidu, line 115: Why does this need to be attempted repeatedly?

web_scraper/README.md
saimidu: I think it would actually be a good idea to have this be a part of the GitHub code, since it doesn't need to use any private credentials, and would be easier to debug.

test/test_utils/__init__.py
```diff
@@ -1344,6 +1344,19 @@ def run_cmd_on_container(container_name, context, cmd, executable="bash", warn=F
...
+def uniquify_list_of_dict(list_of_dict):
+    """
+    Takes list_of_dict as an input and returns a list of dict such that each dict is only present
+    once in the returned list. Runs an operation that is similar to list(set(input_list)). However,
+    for list_of_dict, it is not possible to run the operation directly. 
+
+    :param list_of_dict: List(dict)
+    :return: List(dict)
+    """
+    list_of_string = [json.dumps(dict_element) for dict_element in list_of_dict]
```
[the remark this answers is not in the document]
shantanutrip: Good catch, handled that using `sort_keys=True` in `json.dumps`. Thanks!!

test/test_utils/security.py
shantanutrip: Great suggestion, made the changes accordingly. Thanks!!

testspec.yml
shantanutrip: Yeah, this is needed till we have the temporary infra in place for V1.

mxnet/buildspec.yml
```diff
@@ -37,31 +37,37 @@ context:
...
+    device_type: &DEVICE_TYPE cpu
+    python_version: &DOCKER_PYTHON_VERSION py3
+    tag_python_version: &TAG_PYTHON_VERSION py37
+    os_version: &OS_VERSION ubuntu16.04
+    tag: !join [ *VERSION, "-", *DEVICE_TYPE, "-", *TAG_PYTHON_VERSION, "-", *OS_VERSION ]
+    docker_file: !join [ docker/, *SHORT_VERSION, /, *DOCKER_PYTHON_VERSION, /Dockerfile., *DEVICE_TYPE ]
+    context:
+      <<: *TRAINING_CONTEXT
+      apt-upgrade-list:
+        source: !join [ docker/, *SHORT_VERSION, /, *DOCKER_PYTHON_VERSION, /apt-upgrade-list.txt]
```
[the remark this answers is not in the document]
shantanutrip: Handled the case by adding the processor name to make the files look like `apt-upgrade-list-neuron.txt`. Great catch, thanks!!

mxnet/inference/docker/1.8/py3/Dockerfile.cpu
```diff
@@ -52,6 +52,11 @@ RUN apt-get update \
  && apt-get clean \
  && rm -rf /var/lib/apt/lists/*
 
+COPY apt-upgrade-list.txt /tmp/apt-upgrade-list.txt
+RUN apt-get update \
+    && apt-get install --only-upgrade $(cat /tmp/apt-upgrade-list.txt) \
+    && rm -rf /tmp/apt-upgrade-list.txt
```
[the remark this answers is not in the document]
shantanutrip: Done, thanks!!

mxnet/inference/docker/1.8/py3/Dockerfile.cpu
```diff
@@ -52,6 +52,11 @@ RUN apt-get update \
  && apt-get clean \
  && rm -rf /var/lib/apt/lists/*
 
+COPY apt-upgrade-list.txt /tmp/apt-upgrade-list.txt
```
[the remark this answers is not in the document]
shantanutrip: Done, thanks!!

mxnet/inference/docker/1.8/py3/cu110/Dockerfile.gpu
```diff
@@ -50,6 +50,11 @@ RUN apt-get update \
  && apt-get clean \
  && rm -rf /var/lib/apt/lists/*
 
+COPY apt-upgrade-list.txt /tmp/apt-upgrade-list.txt
```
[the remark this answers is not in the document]
shantanutrip: Done, thanks!!

mxnet/training/docker/1.8/py3/Dockerfile.cpu
```diff
@@ -52,6 +52,11 @@ RUN apt-get update \
  && apt-get clean \
  && rm -rf /var/lib/apt/lists/*
 
+COPY apt-upgrade-list.txt /tmp/apt-upgrade-list.txt
```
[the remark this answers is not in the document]
shantanutrip: Done, thanks!!

mxnet/training/docker/1.8/py3/cu110/Dockerfile.gpu
```diff
@@ -66,6 +66,11 @@ RUN apt-get update \
  && apt-get clean \
  && rm -rf /var/lib/apt/lists/*
 
+COPY apt-upgrade-list.txt /tmp/apt-upgrade-list.txt
```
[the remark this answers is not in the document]
shantanutrip: Done, thanks!!

test/dlc_tests/sanity/test_security_check.py
```diff
@@ -4,19 +4,31 @@
...
+    CVESeverity,
+    ScanVulnerabilityList,
+    ECRScanFailureException,
+    get_ecr_scan_allowlist_path,
 )
 from src.config import is_ecr_scan_allowlist_feature_enabled
 
+LOGGER = logging.getLogger(__name__)
+LOGGER.setLevel(logging.INFO)
+LOGGER.addHandler(logging.StreamHandler(sys.stderr))
```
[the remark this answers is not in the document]
shantanutrip: Rectified, thanks!!

test/dlc_tests/sanity/test_security_check.py
```diff
@@ -54,6 +66,297 @@ def run_scan(ecr_client, image):
...
+
+def is_image_covered_by_allowlist_feature(image):
+    """
+    This method checks if the allowlist feature has been enabled for the image
+
+    :param image: str, Image URI
+    """
+    if any(substring in image for substring in ["example","neuron"]):
+        return False
+    for framework in LOWER_THRESHOLD_IMAGES.keys():
```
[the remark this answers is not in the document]
shantanutrip: Thanks, rectified!!

test/dlc_tests/sanity/test_security_check.py
```diff
@@ -54,6 +66,297 @@ def run_scan(ecr_client, image):
...
+        if framework in image:
+            if any(version in image for version in LOWER_THRESHOLD_IMAGES[framework]):
```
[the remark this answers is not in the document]
shantanutrip: Great suggestion. Implemented accordingly, thanks!!

test/dlc_tests/sanity/test_security_check.py
shantanutrip: I read that `pytest -n auto` uses multiprocessing to run the tests. In that case, change to Global variables in one process is not reflected in the other. However, to be on the safer side, I have removed the usage of global variables altogether.

test/dlc_tests/sanity/test_security_check.py
shantanutrip: Yes, it exists temporarily until some supporting infrastructure is set up. I have added the comment, thanks!!

test/dlc_tests/sanity/test_security_check.py
shantanutrip: Thanks for the suggestions. Made the changes.

test/dlc_tests/sanity/test_security_check.py
shantanutrip: Great suggestion!! Incorporated!!

test/dlc_tests/sanity/test_security_check.py
```diff
@@ -54,6 +66,297 @@ def run_scan(ecr_client, image):
         raise TimeoutError(f"ECR Scan is still in {scan_status} state. Exiting.")
 
 
+def get_new_image_uri(image):
+    """
+    Returns the new image uri for the image being tested. After running the apt commands, the
+    new image will be uploaded to the ECR based on the new image uri.
```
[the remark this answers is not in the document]
shantanutrip: Changed the function name.

test/dlc_tests/sanity/test_security_check.py
shantanutrip: Started by setting `apt_ran_successfully_flag` to `False`, thanks!!

test/dlc_tests/sanity/test_security_check.py
shantanutrip: When a command or application is updating the system or installing a new software (i.e. using apt get update), it locks the dpkg file (Debian package manager). Since we have multiple processes running for the tests, there are cases when one of the process locks the dpkg file. In this scenario, we get error: ‘E: Could not get lock /var/lib/dpkg/lock’ while running apt-get update. That is why we need multiple tries to ensure that it succeeds in one of the tries. More info is seen here: https://itsfoss.com/could-not-get-lock-error/

test/test_utils/security.py
```diff
@@ -187,3 +262,280 @@ def get_ecr_scan_allowlist_path(image_uri):
...
+    )
+    newly_found_non_fixable_list = {}
+    if newly_found_non_fixable_vulnerabilites:
+        newly_found_non_fixable_list = newly_found_non_fixable_vulnerabilites.vulnerability_list
+    message_body = {
+        "edited_files": edited_files,
+        "fixable_vulnerabilities": fixable_list,
+        "non_fixable_vulnerabilities": newly_found_non_fixable_list,
+    }
+    _invoke_lambda(function_name="trshanta-ECR-AS", payload_dict=message_body)
```
saimidu, line 512: Please deploy the final lambda, and add its name as a global constant in `test_utils.security` or something.

test/test_utils/security.py
```diff
@@ -187,3 +262,280 @@ def get_ecr_scan_allowlist_path(image_uri):
...
+    s3_filename_for_fixable_list = s3_filename_for_allowlist.replace(
+        "allowlist.json", "fixable-vulnerability-list.json"
+    )
+    s3_filename_for_non_fixable_list = s3_filename_for_allowlist.replace(
+        "allowlist.json", "non-fixable-vulnerability-list.json"
+    )
+    save_details = []
+    save_details.append((s3_filename_for_fixable_list, failure_routine_summary["fixable_vulnerabilities"]))
+    save_details.append((s3_filename_for_non_fixable_list, failure_routine_summary["non_fixable_vulnerabilities"]))
+    _save_lists_in_s3(save_details, s3_bucket_name)
```
saimidu, line 540: review note: If you are pushing these files to S3 anyway, an alternative approach is to just put an Event Trigger on the Lambda Function, with the event set to be the addition of a new file to the S3 bucket at some expected folder (or just anywhere in the bucket).

However, I don't know the time bounds on S3's Strong Consistency, now, so I think this approach is good as is.

test/test_utils/security.py
```diff
@@ -187,3 +262,280 @@ def get_ecr_scan_allowlist_path(image_uri):
...
+
+def get_ecr_registry(account, region):
+    """
+    Get prefix of ECR image URI
+    :param account: Account ID
+    :param region: region where ECR repo exists
+    :return: AWS ECR registry
+    """
+    endpoint_data = _botocore_resolver().construct_endpoint("ecr", region)
+    return "{}.dkr.{}".format(account, endpoint_data["hostname"])
```
saimidu, line 299: nit: I think this is more of an `ecr` util than a `security` util.

test/test_utils/security.py
```diff
@@ -187,3 +262,280 @@ def get_ecr_scan_allowlist_path(image_uri):
...
+
+
+def _invoke_lambda(function_name, payload_dict={}):
+    """
+    Asyncronously Invokes the passed lambda.
+
+    :param function_name: str, name of the lambda function
+    :param payload_dict: dict, payload to be sent to the lambda
+    """
+    lambda_client = boto3.client("lambda", region_name=os.getenv("REGION"))
```
saimidu, line 378: Please change this to:
```suggestion
    lambda_client = boto3.client("lambda", region_name=test_utils.DEFAULT_REGION)
```
or whatever region the Lambda Function is meant to be deployed to.

The Lambda Function is not replicated across regions, right? There needs to be only one Lambda Function for this, and you can just set the region to be that.

In a hypothetical scenario, if the workflow were being executed in `us-east-1`, for some reason, we don't want it to look for the Lambda Function in a region that it doesn't exist.

testspec.yml
saimidu, line 25: Let's remove this env variable now.

test/test_utils/security.py
```diff
@@ -187,3 +262,280 @@ def get_ecr_scan_allowlist_path(image_uri):
...
+    """
+    Creates a container for the image being tested. Runs apt update and upgrade on the container
+    and the commits the container as new_image_uri. This new image is then pushed to the ECR. 
+
+    :param image: str
+    :param new_image_uri: str
+    """
+    ctx = Context()
+    docker_run_cmd = f"docker run -id --entrypoint='/bin/bash' {image}"
+    container_id = ctx.run(f"{docker_run_cmd}", hide=True, warn=True).stdout.strip()
```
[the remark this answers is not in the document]
shantanutrip: Done, thanks!!

test/test_utils/security.py
```diff
@@ -187,3 +262,280 @@ def get_ecr_scan_allowlist_path(image_uri):
...
+                f"Stderr is {run_output.stderr} \n"
+                f"Failed status is {run_output.exited}"
+            )
+            sleep(2 * 60)
+        elif run_output.ok:
+            apt_ran_successfully_flag = True
+            break
+        if attempt_count == 10:
+            break
+    if apt_ran_successfully_flag == False:
```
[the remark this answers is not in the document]
shantanutrip: Done, thanks!!

test/test_utils/security.py
```diff
@@ -187,3 +262,280 @@ def get_ecr_scan_allowlist_path(image_uri):
...
+                f"Could not run apt update and upgrade. \n"
+                f"Stdout is {run_output.stdout} \n"
+                f"Stderr is {run_output.stderr} \n"
+                f"Failed status is {run_output.exited}"
+            )
+            sleep(2 * 60)
+        elif run_output.ok:
+            apt_ran_successfully_flag = True
+            break
+        if attempt_count == 10:
```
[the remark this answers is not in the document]
shantanutrip: Done, thanks!!

test/test_utils/security.py
```diff
@@ -187,3 +262,280 @@ def get_ecr_scan_allowlist_path(image_uri):
...
+            break
+    if apt_ran_successfully_flag == False:
+        raise ValueError(
```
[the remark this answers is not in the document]
shantanutrip: `RuntimeError` would be more appropriate. Thanks for catching it. Changed to `RuntimeError`.

test/test_utils/security.py
```diff
@@ -187,3 +262,280 @@ def get_ecr_scan_allowlist_path(image_uri):
...
+            f"Could not run apt update and upgrade on image: {image}. \n"
+            f"Stdout is {run_output.stdout} \n"
+            f"Stderr is {run_output.stderr} \n"
+            f"Failed status is {run_output.exited}"
+        )
+    ctx.run(f"docker commit {container_id} {new_image_uri}", hide=True, warn=True)
+    ctx.run(f"docker rm -f {container_id}", hide=True, warn=True)
+    ctx.run(f"docker push {new_image_uri}", hide=True, warn=True)
```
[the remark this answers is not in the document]
shantanutrip: Done, thanks for pointing this out.

test/test_utils/security.py
```diff
@@ -187,3 +262,280 @@ def get_ecr_scan_allowlist_path(image_uri):
...
+    container_id = ctx.run(f"{docker_run_cmd}", hide=True, warn=True).stdout.strip()
+    apt_command = "apt-get update && apt-get upgrade"
+    docker_exec_cmd = f"docker exec -i {container_id}"
+    attempt_count = 0
+    apt_ran_successfully_flag = False
+    # When a command or application is updating the system or installing a new software, it locks the dpkg file (Debian package manager).
+    # Since we have multiple processes running for the tests, there are cases when one of the process locks the dpkg file
+    # In this scenario, we get error: ‘E: Could not get lock /var/lib/dpkg/lock’ while running apt-get update
+    # That is why we need multiple tries to ensure that it succeeds in one of the tries.
+    # More info: https://itsfoss.com/could-not-get-lock-error/
```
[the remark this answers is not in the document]
shantanutrip: Thankyou!!

test/dlc_tests/sanity/test_security_check.py
```diff
@@ -54,6 +65,62 @@ def run_scan(ecr_client, image):
...
+    if is_image_covered_by_allowlist_feature(image):
+        return "MEDIUM"
+    return "HIGH"
+
+
+def fetch_other_vulnerability_lists(image, ecr_client, minimum_sev_threshold):
+    """
+    For a given image it fetches all the other vulnerability lists except the vulnerability list formed by the
+    ecr scan of the current image. In other words, for a given image it fetches upgraded_image_vulnerability_list and
+    image_scan_allowlist.
```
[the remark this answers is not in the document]
shantanutrip: These are great suggestions. I have incorporated the 3rd suggestion in this PR. For the first 2 suggestions, I will make changes in the follow-up PR. Thanks for detailed descriptions.

dlc_developer_config.toml
```diff
@@ -31,10 +31,10 @@ do_build = true
 ### On by default
 sanity_tests = true
   safety_check_test = false
-  ecr_scan_allowlist_feature = false
-ecs_tests = true
-eks_tests = true
-ec2_tests = true
+  ecr_scan_allowlist_feature = true
```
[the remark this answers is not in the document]
shantanutrip: I think it will need a deeper look. I will make the changes in the follow-up PR. Thanks.

test/dlc_tests/sanity/test_security_check.py
```diff
@@ -85,71 +152,69 @@ def test_ecr_scan(image, ecr_client, sts_client, region):
...
-        remaining_vulnerabilities = ecr_image_vulnerability_list - image_scan_allowlist
-
-        assert not remaining_vulnerabilities, (
-            f"The following vulnerabilities need to be fixed on {image}:\n"
-            f"{json.dumps(remaining_vulnerabilities.vulnerability_list, indent=4)}"
+    if is_image_covered_by_allowlist_feature(image):
+        upgraded_image_vulnerability_list, image_scan_allowlist = fetch_other_vulnerability_lists(
+            image, ecr_client, minimum_sev_threshold
+        )
+        s3_bucket_name = "trshanta-bucket"
```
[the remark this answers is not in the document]
shantanutrip: Done, thanks!!

test/test_utils/security.py
```diff
@@ -64,6 +69,54 @@ def construct_allowlist_from_ecr_scan_result(self, vulnerability_list):
...
+            uniquified_list.sort(key=lambda dict_element: dict_element["name"])
+            copy_dict[key] = uniquified_list
+        return dict(sorted(copy_dict.items()))
+
+    def save_vulnerability_list(self, path):
+        if self.vulnerability_list:
+            sorted_vulnerability_list = self.sort_dictionary_in_custom_way(self.vulnerability_list)
+            with open(path, "w") as f:
+                json.dump(sorted_vulnerability_list, f, indent=4)
+
```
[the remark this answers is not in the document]
shantanutrip: Thanks, made the changes!!

test/test_utils/__init__.py
```diff
@@ -1344,6 +1344,19 @@ def run_cmd_on_container(container_name, context, cmd, executable="bash", warn=F
...
+    Takes list_of_dict as an input and returns a list of dict such that each dict is only present
+    once in the returned list. Runs an operation that is similar to list(set(input_list)). However,
+    for list_of_dict, it is not possible to run the operation directly. 
+
+    :param list_of_dict: List(dict)
+    :return: List(dict)
+    """
+    list_of_string = [json.dumps(dict_element, sort_keys=True) for dict_element in list_of_dict]
+    unique_list_of_string = list(set(list_of_string))
+    list_of_dict_to_return = [json.loads(str_element) for str_element in unique_list_of_string]
```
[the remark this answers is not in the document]
shantanutrip: Thankyou :)

test/test_utils/__init__.py
```diff
@@ -1344,6 +1344,19 @@ def run_cmd_on_container(container_name, context, cmd, executable="bash", warn=F
         f"docker exec --user root {container_name} {executable} -c '{cmd}'", hide=True, warn=warn, timeout=60
     )
 
+def uniquify_list_of_dict(list_of_dict):
```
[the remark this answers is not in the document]
shantanutrip: Thanks for sharing the link. Yes it is :)

test/test_utils/security.py
```diff
@@ -64,6 +69,54 @@ def construct_allowlist_from_ecr_scan_result(self, vulnerability_list):
...
+        """
+        if self.vulnerability_list:
+            return [
+                vulnerability
+                for package_vulnerabilities in self.vulnerability_list.values()
+                for vulnerability in package_vulnerabilities
+            ]
+        return None
+
+    def sort_dictionary_in_custom_way(self, input_dict):
```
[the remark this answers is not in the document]
shantanutrip: This is a great suggestion. I have changed the name accordingly. Thanks!!

test/test_utils/security.py
```diff
@@ -64,6 +69,54 @@ def construct_allowlist_from_ecr_scan_result(self, vulnerability_list):
...
+
+        :return: List(dict)
+        """
+        if self.vulnerability_list:
+            return [
+                vulnerability
+                for package_vulnerabilities in self.vulnerability_list.values()
+                for vulnerability in package_vulnerabilities
+            ]
+        return None
```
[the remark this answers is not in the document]
shantanutrip: Thanks for the wonderful suggestion. Made the change!!

test/test_utils/security.py
```diff
@@ -127,6 +180,28 @@ def __sub__(self, other):
         difference.construct_allowlist_from_ecr_scan_result(missing_vulnerabilities)
         return difference
 
+    def __add__(self, other):
+        """
+        Does Union between ScanVulnerabilityList objects
+
+        :param other: Another ScanVulnerabilityList object
+        :return: Union of vulnerabilites exisiting in self and other
+        """
```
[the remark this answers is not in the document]
shantanutrip: Thankyou!!

test/test_utils/security.py
```diff
@@ -187,3 +262,280 @@ def get_ecr_scan_allowlist_path(image_uri):
...
+    """
+    fixable_ecr_image_scan_vulnerabilites = ecr_image_vulnerability_list - upgraded_image_vulnerability_list
+    fixable_allowlist_vulnerabilites = image_allowlist - upgraded_image_vulnerability_list
+    vulnerabilities_fixable_by_upgrade = None
+    if fixable_ecr_image_scan_vulnerabilites and fixable_allowlist_vulnerabilites:
+        vulnerabilities_fixable_by_upgrade = fixable_ecr_image_scan_vulnerabilites + fixable_allowlist_vulnerabilites
+    elif fixable_ecr_image_scan_vulnerabilites:
+        vulnerabilities_fixable_by_upgrade = fixable_ecr_image_scan_vulnerabilites
+    elif fixable_allowlist_vulnerabilites:
+        vulnerabilities_fixable_by_upgrade = fixable_allowlist_vulnerabilites
```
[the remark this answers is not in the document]
shantanutrip: Sure, I will address this in the follow up PR. Thanks!!

test/test_utils/security.py
```diff
@@ -187,3 +262,280 @@ def get_ecr_scan_allowlist_path(image_uri):
...
+    )
+    newly_found_non_fixable_vulnerabilites = upgraded_image_vulnerability_list - image_allowlist
+    fixable_list = {}
+    if vulnerabilities_fixable_by_upgrade:
+        fixable_list = vulnerabilities_fixable_by_upgrade.vulnerability_list
+    apt_upgrade_list_filename = f"apt-upgrade-list-{test_utils.get_processor_from_image_uri(image)}.txt"
+    s3_filename_for_apt_upgrade_list = s3_filename_for_allowlist.replace("allowlist.json", apt_upgrade_list_filename)
+    original_filepath_for_apt_upgrade_list = "/".join(
+        original_filepath_for_allowlist.split("/")[:-1] + [apt_upgrade_list_filename]
+    )
```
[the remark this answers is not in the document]
shantanutrip: Done, thanks for the suggestion.

test/test_utils/security.py
```diff
@@ -187,3 +262,280 @@ def get_ecr_scan_allowlist_path(image_uri):
...
+    :param old_filpath: str, path of original file
+    :param new_packages: list[str], consists of list of packages
+    :param new_filpath: str, path of new file that will have the results of union
+    :param s3_bucket_name: string, name of the s3 bucket
+    """
+    file1 = open(old_filepath, "r")
+    lines = file1.readlines()
+    current_packages = [line.strip() for line in lines]
+    package_list = current_packages
+    new_packages.sort()
```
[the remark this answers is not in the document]
shantanutrip: While writing the code, I was thinking that it would look better if the new packages got appended to the very end of the already existing packages. However, on reading the comment, I felt it would be better to implement the way in which you suggested.

I have made the changes as you suggested. Thanks!!

test/test_utils/security.py
```diff
@@ -187,3 +262,280 @@ def get_ecr_scan_allowlist_path(image_uri):
...
+    for new_package in new_packages:
+        if new_package in package_list:
+            continue
+        package_list.append(new_package)
+    package_list = [f"{package_name}\n" for package_name in package_list]
+    file1.close()
+    run(f"rm -rf {new_filepath}")
+    file2 = open(new_filepath, "w")
+    file2.writelines(package_list)
+    file2.close()
```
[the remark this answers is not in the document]
shantanutrip: Done, thanks!!

test/test_utils/security.py
shantanutrip: I think it could have been done, however, as per the implementation, the PR would be created only when the code is running in canary context. In other words, the contents will be uploaded in the s3 file everytime, however, the lambda should not be invoked everytime. So, I even I think that this approach is good for the time being.

test/test_utils/security.py
shantanutrip: Done, thanks!!

test/test_utils/security.py
shantanutrip: Great suggestion. Made changes, thanks!!

testspec.yml
shantanutrip: Removed, thanks!!

test/test_utils/security.py
```diff
@@ -187,3 +262,280 @@ def get_ecr_scan_allowlist_path(image_uri):
...
+def get_new_image_uri_for_uploading_upgraded_image_to_ecr(image):
+    """
+    Returns the new image uri for the image being tested. After running the apt commands, the
+    new image will be uploaded to the ECR based on the new image uri.
+
+    :param image: str
+    :param new_image_uri: str
+    """
+    ## UPGRADE_REPO_NAME is a temporary name for the ecr repository where the ecr upgraded version of the image is uploaded to run ecr scan on it.
+    new_repository_name = os.getenv("UPGRADE_REPO_NAME")
```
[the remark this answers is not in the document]
shantanutrip: Done, thanks!!

test/test_utils/security.py
shantanutrip: Done, thanks!!

test/test_utils/__init__.py
```diff
@@ -100,6 +100,9 @@
 
 SAGEMAKER_EXECUTION_REGIONS = ["us-west-2", "us-east-1", "eu-west-1"]
 
+UPGRADE_ECR_REPO_NAME = f"""upgraded-image-repo-{boto3.client("sts").get_caller_identity().get("Account")}"""
+ECR_SCAN_HELPER_BUCKET = f"""ecr-scan-helper-{boto3.client("sts").get_caller_identity().get("Account")}"""
```
saimidu, line 104: Please use `boto3.client("sts", region_name=test_utils.DEFAULT_REGION)` in order to avoid using the global STS endpoint.

test/test_utils/security.py
```diff
@@ -189,13 +191,9 @@ def __add__(self, other):
         """
         flattened_vulnerability_list_self = self.get_flattened_vulnerability_list()
         flattened_vulnerability_list_other = other.get_flattened_vulnerability_list()
-        if not flattened_vulnerability_list_self and not flattened_vulnerability_list_other:
+        all_vulnerabilites = flattened_vulnerability_list_self + flattened_vulnerability_list_other
+        if len(all_vulnerabilites) == 0:
```
saimidu, line 195: nit: You can use `if not all_vulnerabilities:`.

test/test_utils/__init__.py
```diff
@@ -100,6 +100,9 @@
 
 SAGEMAKER_EXECUTION_REGIONS = ["us-west-2", "us-east-1", "eu-west-1"]
 
+UPGRADE_ECR_REPO_NAME = f"""upgraded-image-repo-{boto3.client("sts").get_caller_identity().get("Account")}"""
```
saimidu, line 103: An ECR repository is an account-specific resource, and therefore doesn't need the account ID to be a part of its name. We only need the account ID to be in the name of S3 buckets because S3 buckets are global resources across all AWS accounts.
```suggestion
UPGRADE_ECR_REPO_NAME = "upgraded-image-ecr-scan-repo"
```

saimidu:
LGTM. Please make the minor changes recommended in the final review.

saimidu (commented):
LGTM. Please revert temp config changes and mxnet buildspec changes, after which we can go ahead and merge.

test/test_utils/__init__.py
```diff
@@ -350,6 +353,14 @@ def is_time_for_canary_safety_scan():
     current_utc_time = time.gmtime()
     return current_utc_time.tm_hour == 16 and (0 < current_utc_time.tm_min < 20)
 
+def is_time_for_invoking_ecr_scan_failure_routine_lambda():
+    """
+    Canary tests run every 15 minutes.
+    Using a 20 minutes interval to make tests run only once a day around 9 am PST (10 am during winter time).
+    """
+    current_utc_time = time.gmtime()
+    return current_utc_time.tm_hour == 16 and (0 < current_utc_time.tm_min < 20)
```
saimidu, line 362: Please address this comment in a follow-up PR, so that we can unblock the merge of this PR and make this feature usable.

You can merge this function with the `is_time_for_canary_safety_scan` function, and just make it a more generic time-of-day gate function, with a configurable time-of-day when you want to allow some repetitive activity to be allowed, in both the safety check as well as ecr scan.

test/test_utils/__init__.py
```diff
@@ -350,6 +353,14 @@ def is_time_for_canary_safety_scan():
     current_utc_time = time.gmtime()
     return current_utc_time.tm_hour == 16 and (0 < current_utc_time.tm_min < 20)
 
+def is_time_for_invoking_ecr_scan_failure_routine_lambda():
```
shantanutrip, line 356: Eventually make this function and  `is_time_for_canary_safety_scan():` same

saimidu (approved):

PR #1258, opened 2021-08-04
[tensorflow] Update TF2.5 images to only upgrade systemd

arjkesh:

### Description
Avoid upgrading every package, only upgrade systemd

### Tests run
Run PR tests

### DLC image/dockerfile
TF2.5 dockerfiles

### Additional context

### DIFF
--- a/tensorflow/inference/docker/2.5/py3/Dockerfile.cpu
+++ b/tensorflow/inference/docker/2.5/py3/Dockerfile.cpu
@@ -29,7 +29,7 @@ ENV DEBIAN_FRONTEND=noninteractive
 
 # nginx + njs
 RUN apt-get update \
- && apt-get -y upgrade \
+ && apt-get -y upgrade --only-upgrade systemd \
  && apt-get -y install --no-install-recommends \
     curl \
     gnupg2 \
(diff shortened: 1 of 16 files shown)


tensorflow/inference/docker/2.5/py3/Dockerfile.cpu
```diff
@@ -29,7 +29,8 @@ ENV DEBIAN_FRONTEND=noninteractive
 
 # nginx + njs
 RUN apt-get update \
- && apt-get -y upgrade \
```
jeet4320, line 32: is running `apt-get -y upgrade` not updating `systemd`

arjkesh: It is, but it is also upgrading many packages under the hood. We want to maintain consistency in our build system as much as possible from build to build, so we're only updating packages we need to upgrade.

shantanutrip:
I think we can get this merged. We have confirmed that the PR has the same issue that we see on the mainline. So, there is no need to keep the PR waiting. We can merge it and the failure can be taken care of in a totally different PR.

shantanutrip (approved):

arjkesh:
TF2 sanity test failures are for unrelated reason (instance setup for telemetry test). I'm going to merge this in so that we have the correct image building. We'll also need to follow up on systemd issues for TF1

PR #1269, opened 2021-08-11
[dev][test] Add a config option to run RC integration tests & SM local tests in PR

arjkesh:

### Description
- Split "sagemaker_tests" config option into "sagemaker_remote_tests" and "sagemaker_local_tests"
- Provide option to run RC tests on "sagemaker_remote_tests"
 
### Tests run
Ran tests in CI to ensure that RC tests can run in the sagemaker jobs

### DLC image/dockerfile
N/A

### Additional context
This provides option for developers to run RC integration tests in CI before merge, or when updating RC integration tests. It also allows devs to switch on/off sagemaker-local tests independent of the remote tests.

### DIFF
--- a/dlc_developer_config.toml
+++ b/dlc_developer_config.toml
@@ -32,9 +32,9 @@ efa_tests = false
 sagemaker_local_tests = false
 
 # SM remote test valid values:
-# "off" --> do not trigger sagemaker remote tests
-# "default --> run default sagemaker remote tests
+# "off" --> do not trigger sagemaker remote tests (default)
+# "standard" --> run standard sagemaker remote tests from test/sagemaker_tests
 # "rc" --> run release_candidate_integration tests
-sagemaker_remote_tests = "default"
+sagemaker_remote_tests = "standard"
 
 use_scheduler = false
--- a/dlc_developer_config.toml
+++ b/dlc_developer_config.toml
@@ -13,7 +13,7 @@ benchmark_mode = false
 
 [build]
 # Frameworks for which you want to disable both builds and tests
-skip_frameworks = ["huggingface_pytorch", "huggingface_tensorflow", "mxnet"]
+skip_frameworks = []
 # Set to false in order to remove datetime tag on PR builds
 datetime_tag = true
 # Note: Need to build the images at least once with datetime_tag = false
@@ -28,9 +28,8 @@ eks_tests = true
 ec2_tests = true
 
 ### Off by default
-sagemaker_local_tests = true
-# valid values are "off", "default", "release_candidate"
-# if you place any value `val` where `bool(val) == True`, we will warn you, but assume default
+sagemaker_local_tests = false
+# valid values are "off", "default", "rc"
 sagemaker_remote_tests = "off"
 
 efa_tests = false
--- a/dlc_developer_config.toml
+++ b/dlc_developer_config.toml
@@ -30,7 +30,11 @@ ec2_tests = true
 ### Off by default
 efa_tests = false
 sagemaker_local_tests = false
-# valid values are "off", "default", "rc"
-sagemaker_remote_tests = "off"
+
+# SM remote test valid values:
+# "off" --> do not trigger sagemaker remote tests
+# "default --> run default sagemaker remote tests
+# "rc" --> run release_candidate_integration tests
+sagemaker_remote_tests = "default"
 
 use_scheduler = false
--- a/dlc_developer_config.toml
+++ b/dlc_developer_config.toml
@@ -28,10 +28,10 @@ eks_tests = true
 ec2_tests = true
 
 ### Off by default
-sagemaker_local_tests = false
-# valid values are "", "default", "release_candidate"
+sagemaker_local_tests = true
+# valid values are "off", "default", "release_candidate"
 # if you place any value `val` where `bool(val) == True`, we will warn you, but assume default
-sagemaker_remote_tests = "release_candidate"
+sagemaker_remote_tests = "off"
 
 efa_tests = false
 
--- a/dlc_developer_config.toml
+++ b/dlc_developer_config.toml
@@ -21,14 +21,19 @@ datetime_tag = true
 do_build = true
 
 [test]
-# On by default
+### On by default
 sanity_tests = true
 ecs_tests = true
 eks_tests = true
 ec2_tests = true
 
-# Off by default
+### Off by default
 sagemaker_tests = false
+  # Setting release_candidate_tests to `true` will disable SM tests and run RC tests in SM test jobs.
+  # SM local tests will still run.
+  # *sagemaker_tests must be set to `true` if setting release_candidate_tests to `true`.*
+  release_candidate_tests = false
+
 efa_tests = false
 
 use_scheduler = false
--- a/dlc_developer_config.toml
+++ b/dlc_developer_config.toml
@@ -35,6 +35,6 @@ sagemaker_local_tests = false
 # "off" --> do not trigger sagemaker remote tests (default)
 # "standard" --> run standard sagemaker remote tests from test/sagemaker_tests
 # "rc" --> run release_candidate_integration tests
-sagemaker_remote_tests = "rc"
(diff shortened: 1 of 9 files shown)


dlc_developer_config.toml
```diff
@@ -21,14 +21,19 @@ datetime_tag = true
...
+### On by default
 sanity_tests = true
 ecs_tests = true
 eks_tests = true
 ec2_tests = true
 
-# Off by default
-sagemaker_tests = false
+### Off by default
+sagemaker_tests = true
```
SergTogul, line 31: Maybe it will be more useful/clear to have sort of a switch here?
sagemaker_tests = [ "no" | "default" | "release_candidate" ]

SergTogul (approved):

dlc_developer_config.toml
arjkesh: I like this idea a lot - let me scope this out. What about  "" | "default" | "release_candidate" - sort of the same pattern we use for partner_developer

dlc_developer_config.toml
```diff
@@ -21,14 +21,19 @@ datetime_tag = true
...
+  # Setting release_candidate_tests to `true` will disable SM tests and run RC tests in SM test jobs.
```
jeet4320, line 32: can be `will disable SM remote tests`?

arjkesh: sure, will add that

dlc_developer_config.toml
SergTogul: It's not clear what empty string means (""). I'd prefer to see something more self explaining / oblivious. 
What about "off" ? 
And later on we'll refactor *_tests from true/false to "on"/"off"

arjkesh: "off" it is - i also allow "" & false - but I expect default to be "off"

dlc_developer_config.toml
arjkesh: went with @SergTogul suggestion and changed this around. will resolve this for now

src/start_testbuilds.py
```diff
@@ -70,28 +70,22 @@ def run_test_job(commit, codebuild_project, images_str=""):
...
-    if not benchmark_mode:
-        if test_type == constants.ECS_TESTS and ecs_tests_enabled:
+    if not config.is_benchmark_mode_enabled():
+        if test_type == constants.ECS_TESTS and config.is_ecs_test_enabled():
             return True
-        if test_type == constants.EKS_TESTS and eks_tests_enabled:
+        if test_type == constants.EKS_TESTS and config.is_eks_test_enabled():
             return True
-        if test_type == constants.SANITY_TESTS and sanity_tests_enabled:
+        if test_type == constants.SANITY_TESTS and config.is_eks_test_enabled():
```
SergTogul, line 88: You can add a check for is_sm_local_test_enabled() here as well.

src/start_testbuilds.py
```diff
@@ -127,11 +121,15 @@ def main():
...
-                # Trigger sagemaker local test jobs when there are changes in sagemaker_tests
-                # sagemaker local test is not supported in benchmark dev mode
-                if test_type == "sagemaker" and not is_benchmark_mode_enabled():
-                    test_job = f"dlc-pr-{test_type}-local-test"
-                    run_test_job(commit, test_job, images_str)
+            # Trigger sagemaker local test jobs when there are changes in sagemaker_tests
+            # sagemaker local test is not supported in benchmark dev mode
+            if (
+                test_type == "sagemaker"
+                and not config.is_benchmark_mode_enabled()
```
SergTogul, line 128: You won't need it if you handle "sm_local_case" in is_test_job_enabled()

test/test_utils/__init__.py
```diff
@@ -243,6 +243,11 @@ def is_benchmark_dev_context():
     return is_benchmark_mode_enabled()
 
 
+def is_rc_test_context():
+    sm_remote_tests_val = get_sagemaker_remote_tests_config_value()
```
SergTogul, line 247: Why not use is_sm_remote_test_enabled()?

src/config.py
```diff
@@ -31,3 +38,87 @@ def is_benchmark_mode_enabled():
...
+    allowed_config_values = get_allowed_sagemaker_remote_tests_config_values()
+    if sm_config_value not in allowed_config_values:
+        LOGGER.warning(
+            f"Unrecognized sagemaker_remote_tests config {sm_config_value}. "
+            f"Please ensure it is one of {allowed_config_values}, or your tests may not get triggered as expected."
+        )
+    return sm_config_value
+
+
+def is_sm_remote_test_enabled():
```
SergTogul, line 98: Frankly seems overcomplicated. I like [this approach.](https://github.com/aws/deep-learning-containers/pull/1269/files#diff-5d5511865c124cd43c10b4a8f75f026de33c01632acdee7f0fbc849ae5498eb1R248)
Smthng like
`
if value not in ["default", "rc"]:
        LOGGER.warning("unrecognized SM remote test type. Disabling execution.")
        return false
return true
`

src/config.py
```diff
@@ -31,3 +40,69 @@ def is_benchmark_mode_enabled():
...
+    # Sanitize value in case of extra whitespace or inconsistent capitalization
+    if isinstance(sm_remote_tests_value, str):
+        sm_remote_tests_value = sm_remote_tests_value.lower().strip()
+
+    if sm_remote_tests_value == AllowedSMRemoteConfigValues.OFF.value:
+        return False
+    # Support "true" so as not to break existing workflows
+    if sm_remote_tests_value is True:
+        return True
+    if sm_remote_tests_value in (AllowedSMRemoteConfigValues.DEFAULT.value, AllowedSMRemoteConfigValues.RC.value):
```
SergTogul, line 102: You can add True here and delete the [previous if](https://github.com/aws/deep-learning-containers/pull/1269/files#diff-24a3fe5c783755412cb9de83238b4dfd0699345bf50dcadc3ec23c6749361592R100). 

nit: it's more efficient to use set in cases like that.

src/config.py
```diff
@@ -31,3 +40,69 @@ def is_benchmark_mode_enabled():
...
+    Check to see if sm remote test is enabled by config
+    """
+    sm_remote_tests_value = get_sagemaker_remote_tests_config_value()
+    allowed_values = [cfg_opt.value for cfg_opt in AllowedSMRemoteConfigValues]
+
+    # Sanitize value in case of extra whitespace or inconsistent capitalization
+    if isinstance(sm_remote_tests_value, str):
+        sm_remote_tests_value = sm_remote_tests_value.lower().strip()
+
+    if sm_remote_tests_value == AllowedSMRemoteConfigValues.OFF.value:
```
SergTogul, line 97: We don't even need that check. You return False by default.

arjkesh: We should add this check so that we don't warn when this is set to a valid "off" value

arjkesh: How about this - I'll get rid of the condition, but add a condition to "warn" if it's not set to OFF

src/config.py
arjkesh: Sure, we can do that

src/config.py
SergTogul: Yes. You right. 
Sure that works also

SergTogul (dismissed):

SergTogul (approved):

SergTogul (approved):

tejaschumbalkar (approved):

dlc_developer_config.toml
```diff
@@ -21,14 +21,20 @@ datetime_tag = true
...
 ec2_tests = true
 
-# Off by default
-sagemaker_tests = false
+### Off by default
 efa_tests = false
+sagemaker_local_tests = false
+
+# SM remote test valid values:
+# "off" --> do not trigger sagemaker remote tests (default)
```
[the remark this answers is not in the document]
arjkesh: don't want to explicitly call this out as we'll deprecate "true" for this - but I didn't want to break any existing workflow, so I still allow "true"

PR #1361, opened 2021-09-18
[config] Remove efa_tests config option and include as sagemaker_remote_tests option

arjkesh:

### Description
- EFA tests are run on SM remote job in PR - so they should not be a separate option. Refactoring this to avoid confusion when it comes to the separate implementation of sm remote options and the efa option.
- Update PR template to indicate the changes, as well as the tests that should be run prior to new framework merge

### DIFF
--- a/dlc_developer_config.toml
+++ b/dlc_developer_config.toml
@@ -31,13 +31,13 @@ eks_tests = true
 ec2_tests = true
 
 ### Off by default
-efa_tests = false
 sagemaker_local_tests = false
 
 # SM remote test valid values:
 # "off" --> do not trigger sagemaker remote tests (default)
 # "standard" --> run standard sagemaker remote tests from test/sagemaker_tests
 # "rc" --> run release_candidate_integration tests
+# "efa" --> run efa sagemaker tests
 sagemaker_remote_tests = "off"
 
 use_scheduler = false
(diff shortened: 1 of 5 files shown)


jeet4320 (approved):

dlc_developer_config.toml
```diff
@@ -31,13 +31,13 @@ eks_tests = true
...
 
 ### Off by default
-efa_tests = false
 sagemaker_local_tests = false
 
 # SM remote test valid values:
 # "off" --> do not trigger sagemaker remote tests (default)
 # "standard" --> run standard sagemaker remote tests from test/sagemaker_tests
 # "rc" --> run release_candidate_integration tests
+# "efa" --> run efa sagemaker tests
```
shantanutrip, line 40: Thanks for adding this.

dlc_developer_config.toml
```diff
@@ -31,13 +31,13 @@ eks_tests = true
 ec2_tests = true
```
shantanutrip, line 31: Is there a way such that we can add a config variable to run the tests on the PR even if they have been disabled for the PR context?

arjkesh: Do you mean like a universal "on" switch? Currently the only things that are off are the sm local and sm remote tests. Since sm remote tests come in several different configurations, this could get confusing for users. I would suggest not adding this type of option for that reason.

shantanutrip: Sure. Thanks.

shantanutrip (approved):

PR #1374, opened 2021-09-23
[Autogluon] Remove test_autogluon_local_vision_object_detection_gpu - component is not fully supported and planned for deprecation

gradientsky:


test/sagemaker_tests/autogluon/training/integration/local/test_single_machine_training.py
```diff
@@ -66,6 +66,7 @@ def test_autogluon_local_vision_gpu(docker_image, sagemaker_local_session, insta
 @pytest.mark.processor("gpu")
 @pytest.mark.model("autogluon")
 @pytest.mark.skip_cpu
+@pytest.mark.skip(reason="Component is not fully supported and planned for deprecation")
```
arjkesh, line 69: if we are skipping, and we don't plan on using this again, can we just remove the test?

arjkesh (approved):

PR #1493, opened 2021-11-04
bump up TF to 2.6.2

hsl89:
tf2.6 -> tf2.6.2


tensorflow/buildspec.yml
```diff
@@ -84,30 +84,3 @@ images:
       /Dockerfile., *DEVICE_TYPE ]
```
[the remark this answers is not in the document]
hsl89: that makes sense. updated it

satishpasumarthi:
Aren't the wheels supposed to be manylinux wheels?

hsl89:
@satishpasumarthi I don't see why they should be compatible with different linux when we just want them to work in DLC

josephevans:
> @satishpasumarthi I don't see why they should be compatible with different linux when we just want them to work in DLC

These also need to work in DLAMI, on AL2, which requires manylinux for libc compatibility.

shantanutrip (commented):
Looks like there is an increase in the image size. You might want to change the image baseline value to `7800` instead of `7738` in the buildspecs.

tensorflow/buildspec.yml
```diff
@@ -56,7 +56,7 @@ images:
   BuildTensorflowGpuCu112Py38TrainingDockerImage:
     <<: *TRAINING_REPOSITORY
     build: &TENSORFLOW_GPU_TRAINING_PY3 false
-    image_size_baseline: &IMAGE_SIZE_BASELINE 7738
+    image_size_baseline: &IMAGE_SIZE_BASELINE 7800
```
shantanutrip, line 59: Need to change it for the example image as well. You could make `7880` for both, gpu and example gpu images

shantanutrip, line 59: Hey, sorry for inconvenience, can you please convert it to `7880`

arjkesh (approved):

tensorflow/buildspec-2-6.yml
```diff
@@ -0,0 +1,114 @@
...
+    tag_python_version: &TAG_PYTHON_VERSION py38
+    cuda_version: &CUDA_VERSION cu112
+    os_version: &OS_VERSION ubuntu20.04
+    tag: !join [ *VERSION, "-", *DEVICE_TYPE, "-", *TAG_PYTHON_VERSION, "-", *CUDA_VERSION,
+      "-", *OS_VERSION, "-example" ]
+    docker_file: !join [ docker/, *SHORT_VERSION, /, *DOCKER_PYTHON_VERSION, /example,
+      /Dockerfile., *DEVICE_TYPE ]
+    context:
+      <<: *TRAINING_CONTEXT
+  # TODO: Uncomment when Inference images will be released
```
arjkesh, line 87: can you update here that we have not updated inference images for 2.6.2?

shantanutrip: Done

arjkesh (approved):

PR #1521, opened 2021-11-16
Tf inference 2.7 mod

hsl89:
update tensorflow_model_server binary urls.


arjkesh:
@hsl89 there are several conflicts with master here - can you resolve these? Please don't change the buildspec, just the dockerfile changes are necessary

hsl89:
> @hsl89 there are several conflicts with master here - can you resolve these? Please don't change the buildspec, just the dockerfile changes are necessary

Do you mean to keep the buildspec from master?

arjkesh:
> > @hsl89 there are several conflicts with master here - can you resolve these? Please don't change the buildspec, just the dockerfile changes are necessary
> 
> Do you mean to keep the buildspec from master?

yes

arjkesh (approved):

arjkesh (changes requested):
I see that model server still prints out version is rc1. Adding this request changes to block accidental merge

hsl89:
> I see that model server still prints out version is rc1. Adding this request changes to block accidental merge

This is possibly a bug in upstream https://github.com/tensorflow/serving/blob/r2.7/tensorflow_serving/model_servers/version.h#L30

arjkesh (approved):

PR #1532, opened 2021-11-20
Add support for multistage dockerfiles for e3/sagemaker

arjkesh:

### Description
- Add "target" parsing for buildspec to support multistage dockerfiles
- Also provide one example (hidden in .dir) for reference


src/image_builder.py
```diff
@@ -170,6 +170,7 @@ def image_builder(buildspec):
             stage=constants.PRE_PUSH_STAGE,
             context=context,
             additional_tags=[image_tag],
+            target=image_config.get("target")
```
tejaschumbalkar, line 173: Where is the target configured for `e3`

tejaschumbalkar (commented):
How will the backward incompatibility handled given the 2 images managed in a single dockerfile?

src/image_builder.py
arjkesh: Target doesn't have to be explicitly mapped to "e3" or "sagemaker" - it just corresponds to a build target in a multistage dockerfile. If I had a target called "e3" for instance, I'd just specifly `target: e3` in my buildspec

tejaschumbalkar (approved):

PR #1704, opened 2022-02-21
[Tensorflow][Test][ec2] Fix Habana Tensorflow EC2 tests

shantanutrip:

### Description
Tensorflow ec2 tests run for a very long time and thus do not succeed. This PR is formed to handle that issue.

### Tests run
**NOTE: If you are creating a PR for a new framework version, please ensure success of the standard, rc, and efa sagemaker remote tests by updating the dlc_developer_config.toml file:**

- [ ] Revision A: `sagemaker_remote_tests = "standard"`
- [ ] Revision B: `sagemaker_remote_tests = "rc"`
- [ ] Revision C: `sagemaker_remote_tests = "efa"`

- [ ] `sagemaker_local_tests = true`

## Label Checklist
- [ ] I have added the project label for this PR (*<project_name>* or "Improvement")

## PR Checklist
- [x] I've prepended PR tag with frameworks/job this applies to : [mxnet, tensorflow, pytorch] | [ei/neuron/graviton] | [build] | [test] | [benchmark] | [ec2, ecs, eks, sagemaker]
- [ ] If the PR changes affects SM test, I've modified dlc_developer_config.toml in my PR branch by setting sagemaker_tests = true and efa_tests = true
- [ ] If this PR changes existing code, the change fully backward compatible with pre-existing code. (Non backward-compatible changes need special approval.)

## Pytest Marker Checklist

#### EIA/NEURON/GRAVITON Testing Checklist
* When creating a PR:
- [ ] I've modified `dlc_developer_config.toml` in my PR branch by setting `ei_mode = true`, `neuron_mode = true` or `graviton_mode = true`

#### Benchmark Testing Checklist
* When creating a PR:
- [ ] I've modified `dlc_developer_config.toml` in my PR branch by setting `benchmark_mode = true`


test/test_utils/ec2.py
```diff
@@ -452,6 +458,44 @@ def get_ec2_instance_tags(instance_id, region=DEFAULT_REGION, ec2_client=None):
...
+    :param s3_location: str, s3 location where the logs are supposed to be uploaded.
+    :param required_log_ending: str, The string that is desired to be present at the end of the logs
+    :param loop_time: int, seconds for which we would wait for the tests to execute on ec2 instance
+    :param loop_sleep_time: int, interval at which the logs would be uploaded to the s3 bucket
+    :param log_location_within_ec2: Location within ec2 instance where the logs are being witten.
+    """
+    connection.run(execution_command, hide=True, timeout=connection_timeout, asynchronous=True)
+    start_time = int(time.time())
+    loop_count = 0
+    while int(time.time()) - start_time <= loop_time:
```
SergTogul, line 488: I'd suggest to add here something like ` or not last_line_of_log.endswith(required_log_ending)` into while condition. That will let you to stop execution immediately without waiting for the whole loop_time.

Plus maybe we need some sort of a check for hanged container. Something like "if no new lines appear during 15-30 minutes" - we drop the job and fail the test.

test/test_utils/ec2.py
```diff
@@ -452,6 +458,44 @@ def get_ec2_instance_tags(instance_id, region=DEFAULT_REGION, ec2_client=None):
...
+        time.sleep(loop_sleep_time)
+        loop_count += 1
+        connection.run(f"aws s3 cp {log_location_within_ec2} {s3_location}")
```
SergTogul, line 491: Have you considered [CloudWatch](https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/WhatIsCloudWatchLogs.html) specifically designed for cases like that one?

test/test_utils/ec2.py
shantanutrip: This is a good idea. I will do it. Thanks!!

test/test_utils/ec2.py
shantanutrip: We are planning to share the logs with the Habana team using the s3 buckets. So this would solve both the purposes.

test/test_utils/ec2.py
```diff
@@ -452,6 +458,44 @@ def get_ec2_instance_tags(instance_id, region=DEFAULT_REGION, ec2_client=None):
     return {tag["Key"]: tag["Value"] for tag in response.get("Tags")}
 
 
+def execute_asynchronus_testing_using_s3_bucket(
+    connection,
+    execution_command,
+    connection_timeout,
+    s3_location,
```
SergTogul, line 479: I'd suggest to remove a bucket name from here. Otherwise we'll end up with a bunch of buckets those we don't really control. Create 1 bucket in each environment. E.g. dlc-test-asynch-execution-<acc_number>. And 
a. let people to create directories inside of that bucket;
b. implement a logic similar to [this](https://github.com/aws/deep-learning-containers/blob/6d2b86727e79b2cb4e5ba66456f8d5369b952934/test/test_utils/pytest_cache.py#L153)

I personally would prefer (b). It will give us more control. 

BTW! Don't forget to setup a retention policy for that bucket. "Delete after 1 week / 1 month" or something like that.

test/test_utils/ec2.py
```diff
@@ -452,6 +458,73 @@ def get_ec2_instance_tags(instance_id, region=DEFAULT_REGION, ec2_client=None):
...
+                line_count == line_count_list[-1]
+                for line_count in line_count_list[:-number_of_previous_line_counts_to_check]
+            ):
+                # If last 3 runs lead to sam line number then it demonstrates no progress and 
+                # hence we stop.
+                LOGGER.error("No Progress Reported!!")
+                break
+        last_line_of_log = get_last_line_of_s3_file(s3_location, local_filename)
+        LOGGER.info(f"Uploaded file to {s3_location} for {loop_count} number of times")
+    assert last_line_of_log.endswith(required_log_ending), f"Test failed!! Check {s3_location}"
```
SergTogul, line 525: It's arguable but I think you should throw an exception here. E.g. `ValueError("Last row is not as expected")` or something. In some sense it's a runtime error. It's not something you test. 
But even if you keep assert, could you add some explanations to the error message. E.g. "Test failed because the last row is not as expected: 
last row in the log file - {};
expected - {}.
Full log - {s3_location}"

test/test_utils/ec2.py
```diff
@@ -452,6 +458,73 @@ def get_ec2_instance_tags(instance_id, region=DEFAULT_REGION, ec2_client=None):
...
+
+
+def execute_asynchronus_testing_using_s3_bucket(
+    connection,
+    execution_command,
+    connection_timeout,
+    s3_location,
+    required_log_ending,
+    loop_time=2.5 * 3600,
+    loop_sleep_time=5 * 60,
```
SergTogul, line 482: Do you really need such flexibility? I'd suggest to remove it from function parameters. 
Imagine somebody will pass `loop_sleep_time=1`. The test will always fail because there will be no log update during 3 seconds.

And `loop_time` should be a bit less or equal to proper test job timeout parameter. Otherwise you'll have a hard time investigating tests timeout failures

test/test_utils/ec2.py
```diff
@@ -452,6 +458,73 @@ def get_ec2_instance_tags(instance_id, region=DEFAULT_REGION, ec2_client=None):
...
+    :param s3_location: str, s3 location where the logs are supposed to be uploaded.
+    :param required_log_ending: str, The string that is desired to be present at the end of the logs
+    :param loop_time: int, seconds for which we would wait for the tests to execute on ec2 instance
+    :param loop_sleep_time: int, interval at which the logs would be uploaded to the s3 bucket
+    :param log_location_within_ec2: Location within ec2 instance where the logs are being witten.
+    """
+    connection.run(execution_command, hide=True, timeout=connection_timeout, asynchronous=True)
+    start_time = int(time.time())
+    loop_count = 0
+    local_filename = s3_location.split("//")[-1].replace("/", "-")
```
SergTogul, line 503: Do you really need it?
you have a default filename in `get_last_line_of_s3_file()`. You don't have to pass your onw

test/test_utils/ec2.py
```diff
@@ -452,6 +458,73 @@ def get_ec2_instance_tags(instance_id, region=DEFAULT_REGION, ec2_client=None):
...
+        line_count_list.append(number_of_lines_in_log_file)
+        number_of_previous_line_counts_to_check = 3
+        if len(line_count_list) >= number_of_previous_line_counts_to_check:
+            if all(
+                line_count == line_count_list[-1]
+                for line_count in line_count_list[:-number_of_previous_line_counts_to_check]
+            ):
+                # If last 3 runs lead to sam line number then it demonstrates no progress and 
+                # hence we stop.
+                LOGGER.error("No Progress Reported!!")
```
SergTogul, line 521: Could you add here: "No progress reported during X minutes. Job most likely hanged so stopping the execution".

test/test_utils/ec2.py
shantanutrip: I have created a bucket that uses the account number so that we do not have to feel the pain of running it in different accounts.

test/test_utils/ec2.py
shantanutrip: I have removed `loop_sleep_time` as it was too granular. The 'connection_timeout` and `loop_time` are different. `connection_timeout` is the generic fabric connection timeout and `loop_time` is the amount of time for which we would run the loop and is more flexible and can be changed based on tests.

test/test_utils/ec2.py
shantanutrip: I have actually put that to prevent any stepping that might happen when we run multiple tests as multiple processes.

test/test_utils/ec2.py
shantanutrip: Done.

test/test_utils/ec2.py
shantanutrip: Done.

SergTogul (commented):
LGTM

SergTogul (approved):

PR #1809, opened 2022-04-11
[tensorflow][inference] Add test to validate TF serving versioning

arjkesh:

### Description
- Add test to enforce TF serving versioning scheme


test/dlc_tests/sanity/test_pre_release.py
```diff
@@ -141,6 +141,49 @@ def test_ubuntu_version(image):
...
+        image)
+
+    ctx = Context()
+    container_name = get_container_name("tf-serving-version", image)
+    start_container(container_name, image, ctx)
+    output = run_cmd_on_container(
+        container_name, ctx, "tensorflow_model_server --version", executable="bash"
+    )
+    assert (
+        re.match(rf"TensorFlow Model Server: {tag_framework_version}(\D+)?", output.stdout),
```
arjkesh, line 179: I have noticed that sometimes the model server version is not intuitive. For instance, for TF2.7, it looks like this

"TensorFlow ModelServer: 2.7.0-rc1+dev.sha.no_git"

@hsl89 - is there a way we can reliably predict what this string will be? Why does the version sometimes have "rc" specified?

hsl89: @arjkesh , the version is determined by this file https://github.com/tensorflow/serving/blob/2.8.0/tensorflow_serving/model_servers/version.h
the entire string might not be predictable, because it depends on how many rc they have. I think a work around is to grep the major.minor.patch version

arjkesh: @hsl89 is there a way to reliably predict which RC is the latest one, or ensure that the version of the released DLC is not built off of an RC?

I don't want to run into a situation where we have a binary with RC1, for instance, but the release version is built off of RC2

tejaschumbalkar (commented):
LGTM

tejaschumbalkar (approved):

PR #2432, opened 2022-11-17
Split out PyTorch buildspec files, remove out-of-support buildspecs

arjkesh:

### Description
- Remove out of support buildspec files, split inference PT buildspecs into separate buildspecs
- update test cuda paths to read from env variable or dev config, instead of trying to construct buildspec paths


test/test_utils/__init__.py
```diff
@@ -1623,40 +1623,18 @@ def get_python_version_from_image_uri(image_uri):
...
-            raise ValueError(f"Framework version {framework_version} was not matched.")
 
-    buildspec_path = os.path.join(dlc_path, framework_path, f"{buildspec}.yml")
-    if not os.path.exists(buildspec_path):
-        raise ValueError('Could not construct a valid buildspec path.')
+def construct_buildspec_path(dlc_path):
+    """
+    Construct a relative path to the buildspec yaml file by checking for config override or env variable.
+    """
+    buildspec_file_path = config.get_buildspec_override_from_testjob() or os.getenv("FRAMEWORK_BUILDSPEC_FILE")
```
arjkesh, line 1631: I imagine this may have trouble in mainline context if this env variable is not set.

test/dlc_tests/sanity/test_pre_release.py
```diff
@@ -696,6 +696,7 @@ def test_pip_check(image):
 
 @pytest.mark.usefixtures("sagemaker", "huggingface")
 @pytest.mark.model("N/A")
+@pytest.mark.skipif(not is_dlc_cicd_context(), reason="This relies on buildspec env variables being properly set")
 def test_cuda_paths(gpu):
```
arjkesh, line 700: seems like a candidate for quick-checks. Do we need to run this in mainline at all? PR context seems enough - and I'm not convinced we need to enforce this convention anymore

arjkesh: If we also need this in mainline, then we also need to pass FRAMEWORK_BUILDSPEC_FILE to the subsequent jobs in mainline context. We can do this with secondary artifacts or other methods. But to me this seems a bit much for accommodating one test

test/test_utils/__init__.py
arjkesh: note: if preferred, we can probably just remove this function altogether since it is only used in test cuda paths, and put these lines directly into the test.

test/dlc_tests/sanity/test_pre_release.py
```diff
@@ -739,19 +741,10 @@ def test_cuda_paths(gpu):
         # Use the pyX version as opposed to the pyXY version if pyXY path does not exist
         short_python_version = python_version[:3]
 
-    # Check buildspec for cuda version
-    buildspec = "buildspec"
-    if is_tf_version("1", image):
-        buildspec = "buildspec-tf1"
-    if "trcomp" in image:
-        buildspec = "buildspec-trcomp"
-    if "sagemaker-lite" in image:
```
kenny-ezirim, line 748: Are we no longer supporting sagemaker lite?

test/test_utils/__init__.py
```diff
@@ -1623,40 +1623,18 @@ def get_python_version_from_image_uri(image_uri):
...
+
+    if not buildspec_file_path:
+        raise RuntimeError("$FRAMEWORK_BUILDSPEC_FILE is unset and not defined in developer toml. Config will override env variable, but please define.")
+
+    return os.path.join(dlc_path, buildspec_file_path)
```
kenny-ezirim, line 1636: I believe it is safer to check full path os.path.join(dlc_path, buildspec_file_path) and raise an exception if does not exist. If does exist, then return. Should be fine if the check is done after the function construct_buildspec_path returns.

arjkesh:
Closing in favor of #2634

PR #2437, opened 2022-11-17
Install python3.9 in test AMI at runtime

saimidu:

### Description
In EC2 and SM Local Test instances, install Python 3.9 at test preparation step instead of using AMI's default python version, so that test execution can be made agnostic to system python version.

### Tests run
**NOTE: If you are creating a PR for a new framework version, please ensure success of the standard, rc, and efa sagemaker remote tests by updating the dlc_developer_config.toml file:**

- [x] Revision A: `sagemaker_remote_tests = "standard"`
- [ ] Revision B: `sagemaker_remote_tests = "rc"`
- [ ] Revision C: `sagemaker_remote_tests = "efa"`

- [x] `sagemaker_local_tests = true`

## Label Checklist
- [ ] I have added the project label for this PR (*<project_name>* or "Improvement")

## PR Checklist
- [ ] I've prepended PR tag with frameworks/job this applies to : [mxnet, tensorflow, pytorch] | [ei/neuron/graviton] | [build] | [test] | [benchmark] | [ec2, ecs, eks, sagemaker]
- [ ] If the PR changes affects SM test, I've modified dlc_developer_config.toml in my PR branch by setting sagemaker_tests = true and efa_tests = true
- [ ] If this PR changes existing code, the change fully backward compatible with pre-existing code. (Non backward-compatible changes need special approval.)

## Pytest Marker Checklist

#### EIA/NEURON/GRAVITON Testing Checklist
* When creating a PR:
- [ ] I've modified `dlc_developer_config.toml` in my PR branch by setting `ei_mode = true`, `neuron_mode = true` or `graviton_mode = true`

#### Benchmark Testing Checklist
* When creating a PR:
- [ ] I've modified `dlc_developer_config.toml` in my PR branch by setting `benchmark_mode = true`


saimidu:
Tests failed with previously unseen errors when using py39 in [955980d](https://github.com/aws/deep-learning-containers/pull/2437/commits/955980d68416353b55c54e2b61f694b38f6e52a1).

Trying now with py38.

saimidu:
Tests failed with same errors when using py38 in [4e9cc8c](https://github.com/aws/deep-learning-containers/pull/2437/commits/4e9cc8c1a67d5f56d9ff4fe741b8411674ed8a5f).

Trying now with py38.

saimidu:
All tests pass upon upgrading docker-compose. Starting builds for PT Inference DLCs to confirm that change doesn't break PT inference SM Local tests.

saimidu:
All SM Local tests pass on PT Inference 1.12 DLCs on [6081191](https://github.com/aws/deep-learning-containers/pull/2437/commits/6081191b58cb20009f15ad48871796ebeaf881f2)

test/sagemaker_tests/autogluon/inference/requirements.txt
```diff
@@ -5,7 +5,7 @@ protobuf>=3.20,<3.21
 sagemaker>=2.31,<3
 smdebug>=1
 coverage==4.5.3
-docker-compose==1.25.4
+docker-compose
```
kace, line 8: Do we know why this was pinned in the first place?

test/sagemaker_tests/pytorch/inference/requirements.txt
```diff
@@ -12,7 +12,8 @@ PyYAML==5.4
 sagemaker>=2,<3
 six==1.12.0
 requests==2.24.0
-torchvision
+torch==1.11
+torchvision==0.12
```
kace, line 16: Does pinning here mean that we'll have to periodically update this version?

test/sagemaker_tests/pytorch/training/requirements.txt
```diff
@@ -15,8 +15,8 @@ pytest-rerunfailures==8.0
 pytest-xdist==1.28.0
 PyYAML==5.4
 requests==2.24
-torch==1.6.0
-torchvision==0.7.0
+torch==1.11
+torchvision==0.12
```
kace, line 19: Interesting that these were already pinned whereas inference wasn't. Opposite question, do we know why these are pinned?

test/test_utils/sagemaker.py
```diff
@@ -240,25 +243,34 @@ def generate_sagemaker_pytest_cmd(image, sagemaker_test_type):
     )
 
 
-def install_sm_local_dependencies(framework, job_type, image, ec2_conn, ec2_instance_ami):
+@retry(reraise=True, wait=wait_fixed(60), stop=stop_after_attempt(5))
+def install_python_in_instance(context, python_version="3.9"):
```
kace, line 247: Pyenv FTW!

test/test_utils/sagemaker.py
```diff
@@ -304,6 +316,7 @@ def execute_local_tests(image, pytest_cache_params):
     :param pytest_cache_params: parameters required for :param pytest_cache_util
     :return: None
```
kace, line 317: Nit : update the documented return

test/test_utils/sagemaker.py
```diff
@@ -326,10 +339,11 @@ def execute_local_tests(image, pytest_cache_params):
             image,
             ec2_ami_id,
             ec2_key_name,
-            region
+            region,
         )
         ec2_conn = ec2_utils.get_ec2_fabric_connection(instance_id, key_file, region)
         ec2_conn.put(sm_tests_tar_name, f"{UBUNTU_HOME_DIR}")
+        install_python_in_instance(ec2_conn, python_version="3.9")
```
kace, line 346: Remind me, what was the reasoning behind 3.9?

test/test_utils/sagemaker.py
```diff
@@ -392,9 +407,8 @@ def execute_local_tests(image, pytest_cache_params):
         if ec2_client and ec2_key_name:
             print(f"Destroying ssh Key_pair for image: {image}")
             destroy_ssh_keypair(ec2_client, ec2_key_name)
-        # return None here to prevent errors from multiprocessing.map(). Without this it returns some object by default
-        # which is causing "cannot pickle '_thread.lock' object" error
-        return None
+
+    return test_success
```
kace, line 411: Should this not be in the finally code block? Do we want to return a failure for an exception raised during testing?

test/testrunner.py
```diff
@@ -60,7 +61,13 @@ def run_sagemaker_local_tests(images, pytest_cache_params):
 
     pool_number = len(images)
     with Pool(pool_number) as p:
-        p.starmap(sm_utils.execute_local_tests, [[image, pytest_cache_params] for image in images])
+        test_results = p.starmap(sm_utils.execute_local_tests, [[image, pytest_cache_params] for image in images])
+    if not all(test_results):
```
kace, line 65: Nice!

test/test_utils/sagemaker.py
```diff
@@ -240,25 +243,34 @@ def generate_sagemaker_pytest_cmd(image, sagemaker_test_type):
...
+
+    context.run("sudo apt-get update")
+    context.run(
+        "sudo apt install -y make build-essential libssl-dev zlib1g-dev "
+        "libbz2-dev libreadline-dev libsqlite3-dev wget curl llvm "
+        "libncursesw5-dev xz-utils tk-dev libxml2-dev libxmlsec1-dev libffi-dev liblzma-dev"
+    )
+
+    context.run(f"pyenv install {python_version}")
+    context.run(f"pyenv global {python_version}")
```
kace, line 273: Should we add a validation step here to verify our pretest setup was successful?

test/sagemaker_tests/autogluon/inference/requirements.txt
saimidu: I think it was pinned because it was always pinned:
- https://github.com/aws/deep-learning-containers/pull/42
- https://github.com/aws/deep-learning-containers/pull/41

test/sagemaker_tests/pytorch/inference/requirements.txt
saimidu: I believe so, yes. Unfortunately, I wasn't able to deep dive and figure out why the torch import was failing when torch isn't pinned here. It should have been installed as a dependency of torchvision at any compatible version, but apparently that doesn't work as expected.

test/sagemaker_tests/pytorch/training/requirements.txt
saimidu: We copied these test requirements from the sagemaker-pytorch-training/inference-toolkit repositories. I believe this version pin was a consequence of that.

test/test_utils/sagemaker.py
saimidu: Python 3.10 doesn't support a number of PyTorch Ecosystem packages yet. Python 3.9 is as high as I could go.

test/test_utils/sagemaker.py
saimidu: No, the try, except, and finally blocks do not need to return.

Right now, any exceptions raised will be printed out to console, and the list of images that failed tests will be printed out after all parallel processes are joined.

test/test_utils/sagemaker.py
saimidu: Makes sense. Will do.

saimidu:
All non SM-HF-Inference SM Remote tests succeed at [5876897](https://github.com/aws/deep-learning-containers/pull/2437/commits/5876897a5648803a2a39753a5bd55d0abe54feab).

Reverting to SM Local tests and updating SM Local Test AMI to Ubuntu Base DLAMI for latest Nvidia Driver version.

test/test_utils/__init__.py
```diff
@@ -52,11 +55,10 @@ def get_ami_id_ssm(region_name, parameter_path):
     ami_id = eval(ami['Parameter']['Value'])['image_id']
     return ami_id
 
-# TODO revert change after Python 3.7+ is added to Base DLAMI 
-# UBUNTU_18_BASE_DLAMI_US_WEST_2 = get_ami_id_boto3(region_name="us-west-2", ami_name_pattern="Deep Learning Base AMI (Ubuntu 18.04) Version ??.?")
-# UBUNTU_18_BASE_DLAMI_US_EAST_1 = get_ami_id_boto3(region_name="us-east-1", ami_name_pattern="Deep Learning Base AMI (Ubuntu 18.04) Version ??.?")
-UBUNTU_18_BASE_DLAMI_US_WEST_2 = get_ami_id_boto3(region_name="us-west-2", ami_name_pattern="Deep Learning AMI GPU CUDA 11.1.1 (Ubuntu 18.04) ????????")
-UBUNTU_18_BASE_DLAMI_US_EAST_1 = get_ami_id_boto3(region_name="us-east-1", ami_name_pattern="Deep Learning AMI GPU CUDA 11.1.1 (Ubuntu 18.04) ????????")
+
+# TODO revert change after Python 3.7+ is added to Base DLAMI
```
saimidu, line 59: To delete this comment if current revision of PR works.

saimidu:
All SM Local tests succeeded with Deep Learning Base AMI.

test/test_utils/ec2.py
```diff
@@ -904,3 +909,43 @@ def post_process_mxnet_ec2_performance(connection, log_location):
...
+    """
+    Install sagemaker local test dependencies
+    :param context: Invoke Context / Fabric Connection object
+    :param python_version: str python version to install, such as 3.8, 3.9, etc.
+    :return: None
+    """
+    context.run("""ls ~/.pyenv || git clone https://github.com/pyenv/pyenv.git ~/.pyenv""")
+    # Need to configure PATH and PYENV_ROOT changes in alternative location because ~/.bashrc and ~/.profile are
+    # not used when using bash through Fabric Connection.
+    context.run("sudo chmod 666 /etc/profile.d/dlami.sh")
```
arjkesh, line 925: Is this function to explicitly install python on a dlami?

test/test_utils/ec2.py
```diff
@@ -904,3 +909,43 @@ def post_process_mxnet_ec2_performance(connection, log_location):
...
+@retry(reraise=True, wait=wait_fixed(60), stop=stop_after_attempt(5))
+def install_python_in_instance(context, python_version="3.9"):
+    """
+    Install sagemaker local test dependencies
+    :param context: Invoke Context / Fabric Connection object
+    :param python_version: str python version to install, such as 3.8, 3.9, etc.
+    :return: None
+    """
+    context.run("""ls ~/.pyenv || git clone https://github.com/pyenv/pyenv.git ~/.pyenv""")
+    # Need to configure PATH and PYENV_ROOT changes in alternative location because ~/.bashrc and ~/.profile are
```
arjkesh, line 923: instead of running several fabric commands like this, do you think it makes more sense to execute an installer shell script? This way if the installation process differs between python versions in the future, we can have versioned shell scripts to install the given version of python

arjkesh: Multiple cmds is also fine, but I believe in order to get the most of this we need to use "hide=True" because of an issue in pytest/fabric integration that will print these logs elsewhere if the commands fail

test/test_utils/ec2.py
```diff
@@ -904,3 +909,43 @@ def post_process_mxnet_ec2_performance(connection, log_location):
...
+    context.run("sudo chmod 666 /etc/profile.d/dlami.sh")
+    context.run("""echo 'export PYENV_ROOT="$HOME/.pyenv"' >> /etc/profile.d/dlami.sh""")
+    context.run(
+        """echo 'command -v pyenv >/dev/null || export PATH="$PYENV_ROOT/bin:$PATH"' >> /etc/profile.d/dlami.sh"""
+    )
+    context.run("""echo 'eval "$(pyenv init -)"' >> /etc/profile.d/dlami.sh""")
+    context.run("sudo chmod 644 /etc/profile.d/dlami.sh")
+
+    context.run("sudo apt-get update")
+    context.run(
```
arjkesh, line 934: why are we explicitly installing additional packages here? will this be consistently needed across python versions?

test/test_utils/sagemaker.py
```diff
@@ -240,27 +241,6 @@ def generate_sagemaker_pytest_cmd(image, sagemaker_test_type):
     )
 
 
-def install_sm_local_dependencies(framework, job_type, image, ec2_conn, ec2_instance_ami):
```
arjkesh, line 243: I thought this would cause the coverage doc test to fail, but it looks like it is actually failing due to a different issue (seems related to tenacity) fyi

test/test_utils/sagemaker.py
```diff
@@ -302,8 +282,9 @@ def execute_local_tests(image, pytest_cache_params):
     Run the sagemaker local tests in ec2 instance for the image
     :param image: ECR url
     :param pytest_cache_params: parameters required for :param pytest_cache_util
-    :return: None
+    :return: True if test execution was successful, else False
     """
+    test_success = False
```
arjkesh, line 287: why do we need to track execution success/failure this way as opposed to just raising exceptions?

test/test_utils/sagemaker.py
```diff
@@ -379,6 +361,7 @@ def execute_local_tests(image, pytest_cache_params):
                             f"{pytest_command} failed with error code: {res.return_code}\n"
                             f"Traceback:\n{res.stdout}"
                         )
+        test_success = True
```
arjkesh, line 364: Curious - do you know why the code below is catching every exception and just printing it instead of raising? The issue I see is that all the errors raised in the try block above, including ones we explicitly raise, are just printed out

test/sagemaker_tests/pytorch/inference/requirements.txt
arjkesh: I'm confused on this pinning here - why torch 1.11, and what happens if we try to run this on 1.12 or 1.13?

In what circumstances do these get updated?

test/test_utils/ec2.py
```diff
@@ -904,3 +909,43 @@ def post_process_mxnet_ec2_performance(connection, log_location):
         return {"Throughput": total / n}
     else:
         raise ValueError("total: {}; n: {} -- something went wrong".format(total, n))
+
+
+@retry(reraise=True, wait=wait_fixed(60), stop=stop_after_attempt(5))
+def install_python_in_instance(context, python_version="3.9"):
+    """
+    Install sagemaker local test dependencies
```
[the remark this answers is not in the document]
saimidu: ACK. Forgot to update this docstring text. WIll update to say something like "Install python on EC2 instances to create a consistent test environment agnostic to AMI used for test"

test/dlc_tests/conftest.py
```diff
@@ -444,6 +444,8 @@ def delete_s3_artifact_copy():
 
     request.addfinalizer(delete_s3_artifact_copy)
 
+    ec2_utils.install_python_in_instance(conn, python_version="3.9")
```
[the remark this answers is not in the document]
saimidu: Python 3.10 doesn't support a number of PyTorch Ecosystem packages for PT < 1.13. Python 3.9 is as high as I could go.

test/test_utils/ec2.py
saimidu: Could you elaborate on why we should use `hide=True`?

The reason for using pyenv is that it abstracts away any "special steps" to install various python versions. It should be as straightforward as just `pyenv install <python-version>` to install any version.

I can make it an action item for a future PR to use a pre-configured script, but as of now, I don't see the point.

test/test_utils/ec2.py
saimidu: I've just written the installation steps from the pyenv installation instructions as code here. These are all pyenv dependencies used to install any python version. As of now, Ubuntu does not have a deb package to install pyenv. Therefore we need to clone the repo, install its dependencies, and execute the `bin/pyenv` script to install python at the required version from source.

Haven't validated each of the additional packages to see whether they're already installed in the AMI (which could change over time anyway).

test/test_utils/sagemaker.py
saimidu: The problem with python multiprocessing is that it isn't sophisticated enough to handle Exceptions from multiprocessed functions.

The difference between implicitly returning None (by not having a return value in a function) and raising an Exception in the child function, is that the parent function that calls children in multiple processes tries to unpack the Exception returned by the child as "any other return value", and fails to do so, which leads to the weird `cannot pickle '_thread.lock' object` error.

The fix is to always return something. Unfortunately, there is no better solution I could find that would still prevent the silent failures in the SM Local tests.

test/test_utils/sagemaker.py
saimidu: This is intentional. We do not want to raise Exceptions in functions called through `map` or `starmap`.

Reason - https://github.com/aws/deep-learning-containers/pull/2437#discussion_r1029936057

test/sagemaker_tests/pytorch/inference/requirements.txt
saimidu: @arjkesh I just upgraded it to the earliest supported version as of now. This would need to be updated when tests fail due to being unable to generate the saved model to be used for SM PT inference tests.

test/test_utils/ec2.py
```diff
@@ -904,3 +909,43 @@ def post_process_mxnet_ec2_performance(connection, log_location):
         return {"Throughput": total / n}
     else:
         raise ValueError("total: {}; n: {} -- something went wrong".format(total, n))
+
+
+@retry(reraise=True, wait=wait_fixed(60), stop=stop_after_attempt(5))
+def install_python_in_instance(context, python_version="3.9"):
```
kace, line 915: Could we add a precheck to see if the python version we want is already available? As written it looks like we install blindly.

test/dlc_tests/conftest.py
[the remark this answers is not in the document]
arjkesh: OK what's the process around updating this? Is there a way we can do this that conditionally installs python? Example let's say PT 1.14 doesn't support python 3.9 anymore, but we still need to support 1.12 and 1.13 - how do we make such a change?

test/test_utils/ec2.py
arjkesh: Basically if something in this string of run commands causes a test execution to error, in the stack trace it will likely say something like "Stdout/stderr already printed" and you'll need to search through logs to find the error.

When you use hide=True, it prints the error in line with the stack trace

test/test_utils/ec2.py
arjkesh: Can we also make that clear in the docstring or function name?

test/test_utils/ec2.py
saimidu: Got it. I can set `hide=True` right now. I'd rather do the script thing later on without the urgency.

test/test_utils/ec2.py
saimidu: Makes sense, but I think it might be better to do a blind install of pyenv always, because `dist-packages` always make things more difficult for us.

kace: Works for me. I suppose the additional overhead is pretty minimal.

saimidu:
All tests succeed. Reverting config toml changes.

arjkesh (approved):

test/sagemaker_tests/pytorch/inference/requirements.txt
saimidu: Unpinned and it works. I guess installing torch as a dependency of torchvision was the problem.

test/test_utils/sagemaker.py
saimidu: Fixed it with [928fc9b](https://github.com/aws/deep-learning-containers/pull/2437/commits/928fc9b91caebbcc57770f99e63065ef90bc6cf4)

kace (approved):

saimidu:
Force-merging this PR.

PR #2562, opened 2023-01-05
Add inference specific canary logic

arjkesh:

### Description
- Add inference canary logic; Add pull test for training/inference respectively

### DIFF
--- a/test/dlc_tests/sanity/test_canary_integration.py
+++ b/test/dlc_tests/sanity/test_canary_integration.py
@@ -17,10 +17,24 @@ from test.test_utils import (
 @pytest.mark.usefixtures("sagemaker")
 @pytest.mark.skipif(not is_pr_context(), reason=PR_ONLY_REASON)
 @pytest.mark.model("N/A")
-def test_canary_images_pullable(region):
+def test_canary_images_pullable_training(region):
     """
     Sanity test to verify canary specific functions
     """
+    _run_canary_pull_test(region, "training")
+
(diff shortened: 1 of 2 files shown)


test/dlc_tests/sanity/test_canary_integration.py
```diff
@@ -17,10 +17,24 @@
 @pytest.mark.usefixtures("sagemaker")
 @pytest.mark.skipif(not is_pr_context(), reason=PR_ONLY_REASON)
 @pytest.mark.model("N/A")
-def test_canary_images_pullable(region):
+def test_canary_images_pullable_training(region):
```
kenny-ezirim, line 20: I don't think any change is required in this module. Reasons:
1. We can simply obtain the image_type within the parse_canary_images, using the get_image_type (both defined in __init__.py)
2. Both tests will be executed, except if we are calculating the specific function for our canary.
3. It is much cleaner and less changes to make if we determine the image type in the parse_canary_images function, even if change is needed in the future.

I will suggest rolling back changes in the test_canary_integration.py and retrieve image_type within  the parse_canary_images function.

arjkesh: This is a test that relies on sanity test execution of image pulls. As such, it does not have access to the image type env variables. In order to test both training and inference, a change is required to these tests, otherwise tests will just default to training

kenny-ezirim (approved):

PR #2691, opened 2023-02-08
Update PT 1.11 inference DLCs

arjkesh:

### Description
- Rebuild PT 1.11 images, address sanity test failures
- Pin toolkit to appropriate version for PT 1.11
- Add nvgpu dependency which is a requirement for torchserve

### Tests run
- [x] Tests sanity/sagemaker/ec2/ecs/eks ([66306a8](https://github.com/aws/deep-learning-containers/pull/2691/commits/66306a8680658e15b20c9da4281dbbcab9fe29ed))
- [x] Test suggested changes to security conditions ([371fcf3](https://github.com/aws/deep-learning-containers/pull/2691/commits/371fcf3a0ca99f6795e34c550ea95d338f77ec75)) - note: forgot to disable training, please ignore training sanity output


test/test_utils/security.py
```diff
@@ -19,7 +19,7 @@
 @dataclass
```
arjkesh, line 19: ran autoformatter on this file

test/test_utils/security.py
```diff
@@ -483,6 +516,18 @@ def construct_allowlist_from_allowlist_formatted_vulnerabilities(self, allowlist
             self.vulnerability_list[package_name].append(vulnerability)
         return self.vulnerability_list
 
+    def allow_ubuntu_severity_override(self, vulnerability_obj):
```
arjkesh, line 519: actual change to content

test/test_utils/security.py
```diff
@@ -492,14 +537,20 @@ def construct_allowlist_from_ecr_scan_result(self, ecr_format_vulnerability_list
...
-                allowlist_format_vulnerability_object = AllowListFormatVulnerabilityForEnhancedScan(**ecr_format_vulnerability)
+                allowlist_format_vulnerability_object = AllowListFormatVulnerabilityForEnhancedScan(
+                    **ecr_format_vulnerability
+                )
                 vulnerable_package_object = VulnerablePackageDetails(**vulnerable_package)
                 allowlist_format_vulnerability_object.set_package_details_and_name(vulnerable_package_object)
-                if CVESeverity[allowlist_format_vulnerability_object.cvss_v3_severity] < self.minimum_severity:
+                if CVESeverity[
+                    allowlist_format_vulnerability_object.cvss_v3_severity
+                ] < self.minimum_severity or self.allow_ubuntu_severity_override(allowlist_format_vulnerability_object):
```
arjkesh, line 547: actual change to content

test/dlc_tests/sanity/test_ecr_scan.py
```diff
@@ -120,106 +120,6 @@ def conduct_preprocessing_of_images_before_running_ecr_scans(image, ecr_client,
     return image
 
 
-@pytest.mark.usefixtures("sagemaker")
```
arjkesh, line 123: removing this, no need in keeping anything that is marked as obselete

pytorch/inference/docker/1.11/py3/cu115/Dockerfile.gpu
```diff
@@ -213,12 +213,13 @@ RUN pip install --no-cache-dir -U \
...
     "cryptography>3.2" \
     enum-compat==0.0.3 \
-    ipython==8.1.0 \
+    "ipython>=8.10,<9" \
     numpy==1.22.2 \
     "opencv-python>=4.6,<5" \
     packaging \
     "Pillow>=9.3.0" \
     "pyyaml>=5.4,<5.5" \
+    "nvgpu==0.9.0" \
```
arjkesh, line 222: SM image is not release for cu115, so not making SM changes here, just ec2 changes

pytorch/inference/docker/1.11/py3/Dockerfile.cpu
```diff
@@ -177,10 +177,13 @@ RUN pip install --upgrade pip --trusted-host pypi.org --trusted-host files.pytho
  && pip install --no-cache-dir -U ${PT_INFERENCE_URL} ${PT_TORCHVISION_URL} ${PT_TORCHAUDIO_URL} ${PT_TORCHDATA_URL}
 
 RUN pip uninstall -y model-archiver multi-model-server \
- && pip install captum \
```
arjkesh, line 180: captum is now brought in via the TS dependency file

pytorch/inference/docker/1.11/py3/Dockerfile.cpu
```diff
@@ -177,10 +177,15 @@ RUN pip install --upgrade pip --trusted-host pypi.org --trusted-host files.pytho
...
  && pip install torchserve==${TS_VERSION} \
  && pip install torch-model-archiver==${TS_VERSION}
 
+# Install TorchServe pypi dependencies directly from their requirements.txt file
+# NOTE: This also brings in unnecessary cpu dependencies like nvgpu
```
arjkesh, line 184: note, we are installing based on TS requirements, but nvgpu will be brought into the cpu image

test/dlc_tests/sanity/test_ecr_scan.py
xncqr: Understood, this looks to be the old scanner.

xncqr (approved):
As future thought, I believe we need to look at all the pinning we have and verify we actually need it as it does cause more maintenance work when it comes to CVE patching. Looks good. Approving for Inference.

test/dlc_tests/sanity/test_ecr_scan.py
kace: Ack

test/test_utils/security.py
kace: I recommend against doing the `< self.minimum_severity` comparison here and in the `allow_ubuntu_severity_override` method.  Developers would need to know to make changes in both places if there were changes to this logic.

I'd prefer something like the below where we choose the source of the CVE severity in the one-liner and do the comparison in a single place.
```
# Use vendor severity if Ubuntu else use the default severity
cve_severity = CVESeverity[vulnerability_obj.severity] if allowlist_format_vulnerability_object.source == "UBUNTU_CVE" else allowlist_format_vulnerability_object.cvss_v3_severity
if cve_severity < self.minimum_severity:
...
```

We could also move the CVE severity selection logic out of this function and into a getter. This would simplify incorporating other vendors as well.

test/test_utils/security.py
kace: What's preventing us from running a formatter over the entire project? Let's do that!

test/test_utils/security.py
arjkesh: Thanks for the review. I do not think we should combine this into one long conditional statement, as this looks clunky and potentially more confusing to engineers, and if more conditions are added we can expect a more clunky statement. I also am not in favor of assigning cve_severity in a variable as suggested, because it conflates cvss v3 score and vendor severity, as well as changes the logical flow. If the issue is clarity/consistency, I can encapsulate the old condition (cvss v3 score) as well. 

I'm not sure I understand how wrapping the CVESeverity dictionary call in a getter function will simplify incorporating other vendors, but I can generalize the ubuntu method to be a general vendor method in case there are other vendors where we want to follow the same protocol.

test/test_utils/security.py
arjkesh: Best done in a formatter-only PR which also adds in linter/formatting tests to ensure that standards are met going forward. Whole repo is out of scope for this PR.

kace: Agreed. Definitely out of scope for this PR.

test/test_utils/security.py
kace: Works for me. My motivation here was to separate the severity selection from the comparison logic under the assumption that we'd likely see the prior become more complex.

On second thought, it'd probably be best to wait until we update our security policy before anticipating such changes.

kace (approved):

kace (approved):

agunapal:
Supported CUDA versions in PyTorch 1.11 were 10.2 and 11.3 and it was the same in TorchServe. In this PR, the files being updated  are for CUDA 11.3 and 11.5. Shouldn't there be one with CUDA 10.2 as well? cc @lxning

arjkesh:
Hi @agunapal thanks for reviewing. We currently do not have a cu10 image for PT 1.11.

One thing I wanted to confirm is that it is whether or not we need to install nvgpu in the cpu image - currently we are installing everything from the requirements file from torchserve, but if we do not need this package in a cpu environment, we can remove it

agunapal:
@arjkesh Thanks for confirming. The PR looks good to me. Lets keep it for cpu as well. I see that we are importing unconditionally in some files.  I can take up an action item to resolve it in a future release.

agunapal (approved):
Looks good to me

PR #2830, opened 2023-03-30
Add formatting checks to DLC repo

arjkesh:

### Description
- Standardize formatting across all files (use black formatting from https://github.com/psf/black)
- Run black -l 100 on all python files in repo
- Add pyproject.toml to add default configuration
- Following guidelines from https://black.readthedocs.io/en/stable/guides/introducing_black_to_your_project.html and https://black.readthedocs.io/en/stable/faq.html#is-black-safe-to-use

### Tests run
- Add new quick checks test to ensure that formatting is met (test_formatting_and_linting.py)

ToDo for future - add pylint checks (out of scope for now)

I will follow up with another commit to add a .git-blame-ignore-revs file once this is merged in (see https://github.com/aws/deep-learning-containers/pull/2830/files) to avoid overwriting git blame history


Lokiiiiii (dismissed):

test/dlc_tests/sanity/quick_checks/test_formatting_and_linting.py
```diff
@@ -0,0 +1,27 @@
...
+    Check that black style formatting exits gracefully.
+    """
+    # Look up the path until deep-learning-containers is in our base directory
+    dlc_base_dir = os.getcwd()
+    while "deep-learning-containers" not in os.path.basename(dlc_base_dir):
+        dlc_base_dir = os.path.split(dlc_base_dir)[0]
+
+    ctx = Context()
+
+    ctx.run(f"black --check {dlc_base_dir}", hide=True)
```
kace, line 27: Iirc `black` logs errors to stderr. It might be worthwhile to include this output in our test logging by changing the `hide` kwarg value. Reason being that different versions of `black` can sometimes have differences in formatting requirements.

I have personal experience where `black` failed in pipelines and passed locally for this exact reason. It can be very confusing and having any sort of error output from the tool would provide some context for a formatting check failure.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -19,6 +19,9 @@ Note: If merging this PR should also close the associated Issue, please also add
 
 **Additionally, please run the sagemaker local tests in at least one revision:**
 - [ ] `sagemaker_local_tests = true`
+  
+### Formatting
+I have run `black -l 100` on my code (formatting tool: https://black.readthedocs.io/en/stable/getting_started.html)
```
kace, line 24: Is this meant to be a checkbox? If so, it's missing.

test/dlc_tests/sanity/quick_checks/test_yml_config.py
```diff
@@ -34,24 +34,25 @@ def _release_images_yml_verifier(image_type, excluded_image_type):
...
+    while "deep-learning-containers" not in os.path.basename(dlc_base_dir):
         dlc_base_dir = os.path.split(dlc_base_dir)[0]
 
     release_images_yml_file = os.path.join(dlc_base_dir, f"release_images_{image_type}.yml")
 
     # Define exclude regex
     exclude_pattern = re.compile(rf"{excluded_image_type}", re.IGNORECASE)
 
-    with open(release_images_yml_file, "r") as release_imgs_handle:
+    with open(release_images_yml_file, "r", encoding="utf-8") as release_imgs_handle:
```
kace, line 47: [nit] Why stipulate encoding here? My understanding is that `utf-8` is the Python default.

kace, line 47: [nit] Why stipulate encoding here? My understanding is that `utf-8` is the Python default.

pyproject.toml
```diff
@@ -0,0 +1,2 @@
+[tool.black]
+line-length = 100
```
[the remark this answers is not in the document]
arjkesh: pylint default is 100, and PEP8 line length is 79 chars. I went with the longer line default because we have typically been using 120 chars

test/dlc_tests/sanity/quick_checks/test_formatting_and_linting.py
arjkesh: hide=True is necessary to keep output with pytest stacktrace when running tests in parallel. It is an edge case with invoke/pytest behavior, you'll see that throughout our codebase.

It is not as valuable, imo to have the output print somewhere high up in the logs as opposed to along with the stack trace. So as implemented, I think it it will be easier to work with.

test/dlc_tests/sanity/quick_checks/test_yml_config.py
arjkesh: I believe encoding is actually defaulted to system default. See https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/unspecified-encoding.html - 

"It is better to specify an encoding when opening documents. Using the system default implicitly can create problems on other operating systems. See https://peps.python.org/pep-0597/"

.github/PULL_REQUEST_TEMPLATE.md
arjkesh: It prob makes more sense as a checkbox. Added.

kace (approved):

kace (approved):

arjkesh:
Closing this in favor of https://github.com/aws/deep-learning-containers/pull/2891/files

PR #3325, opened 2023-09-11
Update release_images_inference.yml

lanking520:
## Description
Add LMI 0.23.0 DeepSpeed release patching

### DIFF
--- a/release_images_inference.yml
+++ b/release_images_inference.yml
@@ -1,131 +1,17 @@
 ---
 release_images:
   1:
-    framework: "tensorflow"
-    version: "2.13.0"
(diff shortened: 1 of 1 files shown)


tosterberg:
We will either need to check with Arjuna on deleting their release pipeline yaml entries being completed already, or we can just add the DJL related changes as number 11 in the yaml file. Keeping it internal, the latter will be easier to approve quickly.

lanking520:
@tosterberg just do appending for now

tosterberg (approved):

arjkesh (approved):

PR #3391, opened 2023-10-12
[PyTorch][EC2][test] Update testTorchdata

sallyseok:

**Note**: 

### Description

- Torchdata has paused development and is missing the latest release tag for v0.7.0 (which corresponds to PT2.1.0), so use main branch instead. 
- Checkout the latest commit on main as of 10/12/23. The test script most likely will not be updated after this, but pinning the commit just in case.
- I've manually tested that the test on this commit works on inference and training PT2.0.1 EC2 images, which will be the oldest supported images after PT2.1.0 release. 

### DIFF
--- a/test/dlc_tests/container_tests/bin/pytorch_tests/testTorchdata
+++ b/test/dlc_tests/container_tests/bin/pytorch_tests/testTorchdata
@@ -10,6 +10,8 @@ TORCHDATA_TEST_SCRIPT="data/test/test_remote_io.py"
 
 echo "Fetch Torchdata S3 IO test script."
 TORCHDATA_RELEASE_TAG=$(python -c "import torchdata as td; print(td.__version__.split('+')[0])")
+# checkout latest commit on main branch as of 10/12/23, as it has been tested to work on PT 2.0.1 and PT 2.1.0. 
+# torchdata is no longer being maintained, so the test script most likely will not be updated after this.
 git clone $TORCHDATA_REPO \
     && pushd data \
     && git checkout ce075397428c90be312f87b60d4701b2e3ff3700 \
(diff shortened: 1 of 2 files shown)


test/dlc_tests/container_tests/bin/pytorch_tests/testTorchdata
```diff
@@ -12,7 +12,7 @@ echo "Fetch Torchdata S3 IO test script."
 TORCHDATA_RELEASE_TAG=$(python -c "import torchdata as td; print(td.__version__.split('+')[0])")
 git clone $TORCHDATA_REPO \
     && pushd data \
-    && git checkout tags/v${TORCHDATA_RELEASE_TAG} -b ${TORCHDATA_RELEASE_TAG} \
+    && git checkout ce075397428c90be312f87b60d4701b2e3ff3700 \
```
roywei, line 15: let's add a comment on why we are fixing the commit

roywei (approved):

test/dlc_tests/container_tests/bin/pytorch_tests/testTorchdata
sallyseok: done

xncqr: Question. Why are we locking to a specific hash, instead of matching the branch to that of the PT version we are dealing with? Has this been tested against all supported versions of PyTorch?

sallyseok: @xncqr The release tag for v0.7.0 is missing (only the RC tag exists), so the original code doesn't work with PT2.1.0. We decided to just use a fixed commit on main that is working, and yes I tested it against the beta/RC image for PT2.1 ec2 inference (training isn't available yet), as well as 2.0.1. I believe 1.13.1 will be deprecated with the release of 2.1.0?

xncqr: PT 1.13 is still supported and and would be good to verify this test will works with that version.

PR #3470, opened 2023-10-30
fix: Removing libwebp6 in SAI images

chrstfu:

**Note**: 

### Description

Removing libwebp6 in all SAI images as it seems to be causing test failures (reverting https://github.com/aws/deep-learning-containers/pull/3349)

This was originally added to handle a CVE, but the base image has been patched. The integ tests have passed when the update was removed according to @arjunkes - and if it were needed, the sanity checks would have failed.


mabunday:
Can you add some more context here please? What was the rational for adding it in the first place? Is removing it really the right thing to do?

chrstfu:
The PR check wasn't working - closing and reopening to trigger the workflow

PR #3511, opened 2023-11-10
Automate EKS Host Patching

sirutBuasai:

**Note**: 

### Description
- move helper functions to helper.sh
- add patch_hosts.sh and host_patching_buildspec.yml for autopatching scripts 
- remove mxnet from template


eks_infrastructure/patch_hosts.sh
```diff
@@ -0,0 +1,26 @@
...
+source ./helper.sh
+
+# Parse parameters from build_param.json config file
+EKS_CLUSTERS=($(jq -r '.eks_clusters[]' build_param.json))
+CONTEXTS=($(jq -r '.contexts[]' build_param.json))
+EKS_VERSION=$(jq -r '.eks_version' build_param.json)
+CLUSTER_AUTOSCALAR_IMAGE_VERSION=$(jq -r '.cluster_autoscalar_image_version' build_param.json)
+
+# Create a cluster if the context is predeploy and operation is other than create
+if [ "${CONTEXTS}" = "BETA" ] && [ "${OPERATION}" != "create" ]; then
```
tejaschumbalkar, line 14: Either we keep this condition, in which case a staging cluster needs to be added to buildspec. If we do not want to test the upgrade operation on staging cluster, this condition and the delete condition at the end of the file can be removed.

I am inclined on not having staging cluster is because when a new cluster is created the AMI is used as the latest, hence performing upgrade nodegroup has no effect.

eks_infrastructure/patch_hosts.sh
```diff
@@ -0,0 +1,26 @@
...
+  echo "Create a cluster inorder to perform operations other than creating cluster"
+  create_cluster
+fi
+
+# Upgrade hosts to latest ami
+upgrade_nodegroup
```
tejaschumbalkar, line 20: Optionally, We can perform a validation if the cluster exists before performing this operation

eks_infrastructure/patch_hosts.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+#/ Usage: ./patch_hosts.sh
+
+set -ex
+source ./helper.sh
+
+# Parse parameters from build_param.json config file
+EKS_CLUSTERS=($(jq -r '.eks_clusters[]' build_param.json))
```
tejaschumbalkar, line 8: Can we use a separate template similar to build_param.json file that includes the upgrade operation and all the clusters. This file is not supposed to be changed, specially the operation field

The build_param.json is meant to be changed for different operations, in which case this upgrade job will not serve its purpose.

eks_infrastructure/patch_hosts.sh
```diff
@@ -0,0 +1,26 @@
...
+CONTEXTS=($(jq -r '.contexts[]' build_param.json))
+EKS_VERSION=$(jq -r '.eks_version' build_param.json)
+CLUSTER_AUTOSCALAR_IMAGE_VERSION=$(jq -r '.cluster_autoscalar_image_version' build_param.json)
+
```
tejaschumbalkar, line 12: Add validation for operation, if upgrade operation is not found, skip the job
Or
If we are creating a duplicate file similar to build_param.json, let's not include operation field, we can only keep cluster names and other parameters if required

eks_infrastructure/patch_hosts.sh
sirutBuasai: I agree with no having this stage instead. I meant to delete that in the previous commit that I deleted the input logic but forgot to delete the staging logic.

eks_infrastructure/patch_hosts.sh
sirutBuasai: [upgrade_nodegroup](https://github.com/aws/deep-learning-containers/pull/3511/files#diff-00242329c764422d7007b79f0cf43e67bf3b2357483edf916512a0160a910470R60) already checks if the cluster already exists.

eks_infrastructure/patch_hosts.sh
sirutBuasai: Will do.

eks_infrastructure/patch_hosts.sh
sirutBuasai: there is no operation variable in this file.

eks_infrastructure/helper.sh
```diff
@@ -0,0 +1,124 @@
+#!/bin/bash
```
tejaschumbalkar, line 1: was this file pushed with +x chmod permissions?

eks_infrastructure/patch_hosts.json
```diff
@@ -0,0 +1,11 @@
+{
+    "contexts": [
+        "PR",
+        "MAINLINE"
+    ],
+    "eks_clusters": [
+        "dlc-pytorch",
+        "dlc-tensorflow"
+    ],
+    "eks_version": "1.25"
```
tejaschumbalkar, line 10: lets remove `eks_version` from this file. We should refer the value of `eks_version` from `build_param.json` since it is a source of truth of of the latest eks version.

eks_infrastructure/patch_hosts.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+#/ Usage: ./patch_hosts.sh
+
+set -ex
+source ./helper.sh
+
+# Parse parameters from build_param.json config file
+EKS_CLUSTERS=($(jq -r '.eks_clusters[]' patch_hosts.json))
+CONTEXTS=($(jq -r '.contexts[]' patch_hosts.json))
+EKS_VERSION=$(jq -r '.eks_version' patch_hosts.json)
```
tejaschumbalkar, line 10: make changes per the previous comment

eks_infrastructure/patch_hosts.template
```diff
@@ -0,0 +1,14 @@
...
+{
+    "contexts":[
+        "PR"
+        "MAINLINE"
+    ],
+    "eks_clusters":[
+        "dlc-pytorch"
+        "dlc-tensorflow"
+    ],
+    "eks_version": "1.25"
```
tejaschumbalkar, line 13: remove `eks_version` from this file as well

eks_infrastructure/patch_hosts.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
tejaschumbalkar, line 1: was this file pushed with +x chmod permissions?

eks_infrastructure/helper.sh
sirutBuasai: no. it shouldn't be though cuz we are just sourcing it and not executing it.

eks_infrastructure/patch_hosts.sh
sirutBuasai: yes, this file has executable permission.

tejaschumbalkar (approved):
LGTM

PR #4509, opened 2025-01-24
Revert temp boto3 fix and pin awscli/boto3

sallyseok:

**Note**: 

### Description
- This will upgrade the old awscli/boto3 (and thus botocore and s3transfer) versions in our build image to the current latest versions (as of 1/27/2025), which are compatible with each other and comply with sagemaker package's requirements
- Also removed the fabric/invoke pins since they're no longer needed

### Tests run
- PT inference 2.5 SM: [de66a53](https://github.com/aws/deep-learning-containers/pull/4509/commits/de66a536f38e82125c923e6cbf2bef1db20c3d8d)
- PT inference 2.4 EC2: [6ba5fab](https://github.com/aws/deep-learning-containers/pull/4509/commits/6ba5fabfec2cd7a878ad78bacd37d711a2f9a94f)
- PT inference arm64 2.5 EC2: [1745342](https://github.com/aws/deep-learning-containers/pull/4509/commits/1745342dd9b1d86a01dfd086bbb736c06ca09b82)
- PT training 2.5 SM: [dea00ac](https://github.com/aws/deep-learning-containers/pull/4509/commits/dea00ac17c9fc149a84f3eb7d2f38ff7107d1454)
- TF inference 2.18 EC2 (non-autopatch): [82ec321](https://github.com/aws/deep-learning-containers/pull/4509/commits/82ec321b5bb3a17d3ca47b05b0d0338ec2ca362c)


src/requirements.txt
```diff
@@ -1,12 +1,12 @@
+awscli==1.37.6
+boto3==1.36.6
```
shantanutrip, line 2: Do we want to pin to these versions? Was it the expected behavior earlier?

sallyseok: If we don’t include the pins here, the versions will stay as the old versions in our build container since they’re already installed (which eventually leads to incompatibility). As an alternative, to force these to always upgrade to latest we can change the pip install line to `pip install --upgrade -r src/requirements.txt`, but I think that would be too invasive. Seeing that most of the packages in src/requirements.txt already have pins, I decided on following that convention.

sallyseok: So yeah the previous behavior was that boto3 and awscli (and fabric) would stay as whatever versions were installed in the build image

shantanutrip: Looks like we had `boto3>=1.33` in the `requirements.txt`. I know that this >= condition was of no use due to boto3 being already present. However, looks like the intended behavior was to probably upgrade boto3 to later versions.

It might fail due to SM dependency that needs a higher version. Do you think we can probably modify it to `boto3>=1.36` to prevent future updates?

sallyseok: If SM requires a higher version in the future, it will actually just upgrade boto3 without failing since we install SM later in test requirements

shantanutrip: Ack

shantanutrip (approved):

PR #5463, opened 2025-11-07
Migrate vLLM Ray Serve Container

junpuf:
a sample PR build and test workflow

### DIFF
--- a/.github/workflows/pr-vllm-rayserve.yml
+++ b/.github/workflows/pr-vllm-rayserve.yml
@@ -2,6 +2,8 @@ name: PR - vLLM RayServe
 
 on:
   pull_request:
+    branches: 
+      - main
     paths:
       - "docker/**"
 


junpuf:
> We'll most likely need cuda compat script as well. not sure how vllm has been working thus far. https://github.com/aws/deep-learning-containers/blob/eb524f7c0737b007cf06d4fd36f67de246cc8d8f/sglang/build_artifacts/start_cuda_compat.sh

Not running any SageMaker test yet

.github/workflows/pr-vllm-rayserve.yml
```diff
@@ -0,0 +1,322 @@
...
+      - name: Pull image
+        run: |
+          docker pull $IMAGE_URI
+      
+      - name: Checkout vLLM Tests
+        uses: actions/checkout@v5
+        with:
+          repository: vllm-project/vllm
+          ref: v0.10.2
+          path: vllm_source
```
sirutBuasai, line 114: I wonder if there's a way to DRY these steps. These are going to be used repeatedly across multiple stages

junpuf: later we can refactor common patterns into callable workflows or other things

scripts/telemetry/deep_learning_container.py
```diff
@@ -0,0 +1,395 @@
...
+
+def parse_args():
+    """
+    Parsing function to parse input arguments.
+    Return: args, which containers parsed input arguments.
+    """
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--framework",
+        choices=["tensorflow", "mxnet", "pytorch", "base", "vllm"],
```
sirutBuasai, line 242: We can probably fix this list to `["tensorflow", "pytorch", "base", "vllm", "sglang"]`

scripts/telemetry/deep_learning_container.py
```diff
@@ -0,0 +1,395 @@
...
+        help="framework of container image.",
+        required=True,
+    )
+    parser.add_argument(
+        "--framework-version", help="framework version of container image.", required=True
+    )
+    parser.add_argument(
+        "--container-type",
+        choices=["training", "inference", "general"],
```
sirutBuasai, line 251: Also side note and unrelated to this PR. Currently `vllm` and `sglang` are classified as `general`. We should change these to `inference`.

scripts/telemetry/deep_learning_container.py
junpuf: actually this telemetry things doesn't work, cuz the template replacement stuff doesn't exist. will need to fix it separately

scripts/telemetry/deep_learning_container.py
junpuf: the overall telemetry integration needs to be fixed.

PR #5521, opened 2025-11-29
Refactor vllm sm endpoint test to use pytest

sallyseok:
- Updated the vllm sm endpoint test file to match the sglang version using pytest instead of a direct executable
- Also added some constants used across both files

### DIFF
--- a/test/test_utils/constants.py
+++ b/test/test_utils/constants.py
@@ -1,3 +1,4 @@
 DEFAULT_REGION = "us-west-2"
 SAGEMAKER_ROLE = "SageMakerRole"
+# https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_ProductionVariant.html
 INFERENCE_AMI_VERSION = "al2-ami-sagemaker-inference-gpu-3-1"


test/sglang/sagemaker/test_sm_endpoint.py
```diff
@@ -63,14 +46,13 @@ def model_id(request):
 
 @pytest.fixture(scope="function")
 def instance_type(request):
-    # Return the model_id given by the test parameter
+    # Return the instance_type given by the test parameter
```
sirutBuasai, line 49: ah good catch thanks!

test/test_utils/constants.py
```diff
@@ -1 +1,3 @@
 DEFAULT_REGION = "us-west-2"
+SAGEMAKER_ROLE = "SageMakerRole"
+INFERENCE_AMI_VERSION = "al2-ami-sagemaker-inference-gpu-3-1"
```
sirutBuasai, line 4: Maybe link this in the comment for future AMI name ref? https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_ProductionVariant.html

sirutBuasai (approved):

PR #5533, opened 2025-12-04
fix v2 telemetry

junpuf:
Fixed v2 telemetry setup by converting the script into a template and substitute variables during docker build. Verified with local testing like below:

```
404426647817.dkr.ecr.us-west-2.amazonaws.com/ci:vllm-0.11.2-gpu-py312-cu129-ubuntu22.04-ec2-pr-5533
https://aws-deep-learning-containers-d8c0d063.s3.us-west-2.amazonaws.com/dlc-containers-i-0566a1a9029f8cdaa.txt?x-instance-id=i-0566a1a9029f8cdaa&x-framework=vllm&x-framework_version=0.11.2&x-py_version=3.12.12&x-container_type=general
7e20109039605d6274b0e538013ef341c4b5c659ea98059a80bcb85061be35ec
```

However, the GHA testing is failing because unable to access IMDS within the GHA process, although IMDSv2 is accessible from the CodeBuild job itself.  

Potential workaround for having telemetry testing maybe using long-running EKS cluster.

### DIFF
--- a/docker/sglang/Dockerfile
+++ b/docker/sglang/Dockerfile
@@ -1,8 +1,5 @@
 # Declare the argument as default to use as input
 # base image: https://hub.docker.com/r/lmsysorg/sglang/tags
-ARG FRAMEWORK=sglang
-ARG FRAMEWORK_VERSION=0.5.5
-ARG CONTAINER_TYPE=general
 ARG BASE_IMAGE=lmsysorg/sglang:v0.5.5-cu129-amd64
 
 # Use input argument as base image
--- a/docker/vllm/Dockerfile
+++ b/docker/vllm/Dockerfile
@@ -1,8 +1,5 @@
 # Declare the argument as default to use as input
 # base image: https://hub.docker.com/r/vllm/vllm-openai/tags
-ARG FRAMEWORK=vllm
-ARG FRAMEWORK_VERSION=0.11.0
-ARG CONTAINER_TYPE=general
 ARG BASE_IMAGE=vllm/vllm-openai:v0.11.0
 
 # Use input argument as base image


docker/sglang/Dockerfile
```diff
@@ -32,13 +35,23 @@ WORKDIR /
...
 RUN chmod +x /usr/local/bin/deep_learning_container.py \
+  && sed -e "s/{{FRAMEWORK}}/${FRAMEWORK}/g" \
+    -e "s/{{FRAMEWORK_VERSION}}/${FRAMEWORK_VERSION}/g" \
+    -e "s/{{CONTAINER_TYPE}}/${CONTAINER_TYPE}/g" \
+    /tmp/bash_telemetry.sh.template >/usr/local/bin/bash_telemetry.sh \
   && chmod +x /usr/local/bin/bash_telemetry.sh \
-  && chmod +x /usr/local/bin/start_cuda_compat.sh
+  && chmod +x /usr/local/bin/start_cuda_compat.sh \
+  && rm /tmp/bash_telemetry.sh.template \
+  && uv pip install --system botocore \
```
junpuf, line 50: `botocore` is needed to made telemetry work, verified that this would only install botocore and 1 dependency of it, no existing dependencies from sglang are modified.

.github/workflows/pr-vllm.yml
```diff
@@ -118,6 +120,9 @@ jobs:
           docker buildx build --progress plain \
             --build-arg CACHE_REFRESH="$(date +"%Y-%m-%d")" \
             --build-arg BASE_IMAGE="vllm/vllm-openai:v${{ env.VLLM_VERSION }}" \
+            --build-arg CONTAINER_TYPE="${{ env.CONTAINER_TYPE }}" \
```
Yadan-Wei, line 123: I see ARG CONTAINER_TYPE is defined with default value before from and then without default after. This build-arg could work fine. But do we still need these ARG defined on top of docker file since they are not used?

Yadan-Wei: I see ARG CONTAINER_TYPE is defined with default value before from and then without default after. This build-arg could work fine.

docker/sglang/Dockerfile
```diff
@@ -1,5 +1,8 @@
 # Declare the argument as default to use as input
 # base image: https://hub.docker.com/r/lmsysorg/sglang/tags
+ARG FRAMEWORK=sglang
```
Yadan-Wei: Will this be redundant?

PR #5552, opened 2025-12-12
refactor workflows

junpuf:
1. made a image build action.
2. refactor the duplicating test commands into separate scripts.


.github/workflows/pr-sglang.yml
```diff
@@ -164,7 +143,7 @@ jobs:
           echo "IMAGE_URI=${IMAGE_URI}" >> ${GITHUB_OUTPUT}
 
   sglang-local-benchmark-test:
-    needs: [set-test-environment, build-sglang-image]
```
sirutBuasai, line 167: This change means that we will test local-benchmark-test on prod image if the build didn't happen btw. Do we want to always run the test?

My thought process for implementing it this way was that al these upstream tests only need to run if we're building a new image. If we're using a prod image, then we can assume that these tests passed

.github/workflows/pr-sglang.yml
sirutBuasai, line 1: do we want to organize the env part to be the same format as `pr-vllm` file?

junpuf: probably not necessary at this point, we can keep that free form for now

.github/workflows/pr-sglang.yml
junpuf: Hmm, make sense to me, i can add it back

sirutBuasai (approved):

PR #6044, opened 2026-05-05
feat: wire vLLM SageMaker entrypoint to standard-supervisor

Lokiiiiii:
### Summary
Adds standard-supervisor (from model-hosting-container-standards) to the vLLM SageMaker launch chain, aligning the DLC with the vLLM open source SageMaker stage which already uses standard-supervisor as its entrypoint wrapper.

### Why this is necessary
1. Parity with vLLM open source

The upstream vLLM project's sagemaker-entrypoint.sh already launches via standard-supervisor.

2. Process supervision and auto-recovery

Without standard-supervisor, if the vLLM process crashes, the container exits immediately. With it, the process is managed by supervisord which:

Automatically restarts the vLLM process on unexpected exits (configurable via PROCESS_AUTO_RECOVERY, default: true)
Retries up to N times before giving up (configurable via PROCESS_MAX_START_RETRIES, default: 3)
Provides structured exit codes so SageMaker can distinguish between transient failures and permanent ones

3. Dynamic dependency installation

standard-supervisor (v0.1.15+) automatically installs dependencies from requirements.txt before starting the server. This enables customers to bundle custom dependencies with their model artifacts without needing a custom container. 

Controlled by:
STANDARD_AUTO_INSTALL_REQ (default: true) — enable/disable
STANDARD_PIP_ARGS — override with explicit pip arguments

### Testing
Local validation 
CI will run: sanity tests, security tests, telemetry tests, upstream vLLM tests (GPU), and a real SageMaker endpoint deployment test

### DIFF
--- a/docker/vllm/Dockerfile
+++ b/docker/vllm/Dockerfile
@@ -36,7 +36,7 @@ RUN uv pip install --system \
   "pillow>=12.1.1" \
   "xgrammar>=0.1.32" \
   "PyJWT>=2.12.0" \
-  "model-hosting-container-standards>=0.1.14,<1.0.0" \
+  "model-hosting-container-standards>=0.1.15,<1.0.0" \
   "pyasn1>=0.6.3" \
   && uv cache clean
 
--- a/docker/vllm/Dockerfile.amzn2023
+++ b/docker/vllm/Dockerfile.amzn2023
@@ -345,7 +345,7 @@ RUN uv pip install --no-cache-dir \
   "xgrammar>=0.1.32" \
   "PyJWT>=2.12.0" \
   "cbor2>=5.9.0" \
-  "model-hosting-container-standards>=0.1.14,<1.0.0"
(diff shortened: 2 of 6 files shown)


docker/vllm/Dockerfile
```diff
@@ -36,7 +36,7 @@ RUN uv pip install --system \
   "pillow>=12.1.1" \
   "xgrammar>=0.1.32" \
   "PyJWT>=2.12.0" \
-  "model-hosting-container-standards>=0.1.14,<1.0.0" \
+  "model-hosting-container-standards>=0.1.15,<1.0.0" \
```
junpuf, line 39: Could you also update the AL2023 Dockerfile as well? https://github.com/aws/deep-learning-containers/blob/main/docker/vllm/Dockerfile.amzn2023#L348

junpuf (approved):

PR #6091, opened 2026-05-12
refactor: version PyTorch directory structure for multi-version support

Eren-Jeager123:
## Summary

Restructure PyTorch to support maintaining multiple versions concurrently (1-year support window). Versioned directories isolate each version's build artifacts. Workflows are version-agnostic — no new workflow files needed when adding a future version.

### 1. Versioned directory structure

```
docker/pytorch/2.11/
├── Dockerfile.cuda
├── Dockerfile.cpu
├── versions-cuda.env      # Build-time dependency versions (sourced by workflows + Dockerfiles)
├── versions-cpu.env
├── cuda/pyproject.toml
├── cuda/uv.lock
├── cpu/pyproject.toml
└── cpu/uv.lock

.github/config/image/
├── pytorch-2.11-ec2-cuda.yml       # Workflow metadata (framework, labels, prod_image)
├── pytorch-2.11-ec2-cpu.yml
├── pytorch-2.11-sagemaker-cuda.yml
└── pytorch-2.11-sagemaker-cpu.yml
```

Shared scripts (`scripts/pytorch/`) and tests (`test/pytorch/`) stay in place.

### 2. PR workflows — configs matrix (zero changes to reusable workflows)

```
gatekeeper → check-changes → build-images / unit-test / sanity / security / telemetry / single-gpu / efa
```

**Key design:** The `check-changes` job detects PyTorch versions from the diff, reads each version's config YAML, and outputs a rich `configs` matrix where each entry carries all metadata fields. Downstream test jobs iterate over this matrix and pass `matrix.*` fields directly as individual inputs to **unmodified** reusable workflows.

```yaml
# check-changes outputs:
#   configs: [{"version":"2.11", "framework":"pytorch_runtime", ..., "prod_image":"pytorch:2.11-cu130-amzn2023"}]

sanity-test:
  strategy:
    matrix:
      include: ${{ fromJson(needs.check-changes.outputs.configs) }}
  uses: ./.github/workflows/reusable-sanity-tests.yml
  with:
    image-uri: ${{ needs.build-images.result == 'success' && CI_IMAGE || PROD_IMAGE }}
    framework: ${{ matrix.framework }}
    framework-version: ${{ matrix.framework_version }}
    ...
```

This solves the GitHub Actions limitation where matrix job outputs only expose the last-finishing leg's values — metadata travels **with** the matrix, not as job outputs.

**Version detection** combines both path patterns in a single grep:
```bash
git diff --name-only origin/main...HEAD \
  | grep -oP '(?:docker/pytorch/|pytorch-)\K[0-9]+\.[0-9]+'
```

**Prod image fallback:** When only test files change (`build-change: false`), build-images is skipped and test jobs automatically fall back to the prod image:
```yaml
image-uri: ${{ needs.build-images.result == 'success' && CI_IMAGE || PROD_IMAGE }}
aws-account-id: ${{ needs.build-images.result == 'success' && CI_ACCOUNT || PROD_ACCOUNT }}
```

### 3. Autorelease workflows — simplified

- **Single `load-config` job** (merged determine-config + load-config to avoid spinning up two instances)
- **`version` dispatch input** instead of full config file path (e.g., `2.11`)
- **`framework-short-version` output** (e.g., `"2.11"`) parsed once via jq, used everywhere — no repeated `cut -d. -f1,2` in downstream jobs

### 4. Dockerfile ARG for version paths

Dockerfiles use `ARG DLC_PYTORCH_VERSION=2.11` instead of hardcoding:
```dockerfile
COPY docker/pytorch/${DLC_PYTORCH_VERSION}/cuda/pyproject.toml /tmp/build/
```

### 5. Additional improvements

- **Unit tests** as separate job (parallel with sanity/security/telemetry)
- **Removed dead matrix outputs** from build jobs (`runtime-image-uri` etc. that suffered from clobbering)
- **Trailing backslash fix** in wheel fetch/upload commands
- **upload_cached_wheels.sh** accepts Dockerfile path as parameter
- **test_versions.py** reads `DLC_PYTORCH_VERSION` env var to locate versioned env files
- **libcap security patch** (CVE-2026-33814, CVE-2026-39820, CVE-2026-33811, CVE-2026-42499)
- **Bumped actions** — checkout@v6, upload-artifact@v7, download-artifact@v8
- **Zero changes to reusable workflows** — no blast radius to other frameworks

### Adding PyTorch 2.12

1. Copy `docker/pytorch/2.11/` → `docker/pytorch/2.12/`, update version pins
2. Create 4 config files (`pytorch-2.12-{ec2,sagemaker}-{cuda,cpu}.yml`)
3. Autorelease: add one cron line + one case entry per workflow
4. PR workflows: auto-detect from diff — no changes needed

## Test plan

- [x] PR workflow detects version from `docker/pytorch/2.11/**` and `pytorch-2.11-*` changes
- [x] Configs matrix passes correct metadata per version to reusable workflows
- [x] Unit tests run as separate job per version
- [x] Reusable workflows (vLLM, SGLang, Ray, Base) work unchanged
- [x] `DLC_PYTORCH_VERSION` build-arg resolves correctly in Dockerfile COPY paths
- [x] Security scan passes (libcap patched)
- [x] Pre-commit passes (actionlint, YAML validation, python formatting)
- [ ] Prod image fallback works when build is skipped (test-only change PR)
- [ ] Autorelease `workflow_dispatch` with version input works
- [ ] Multi-version PR builds all versions in parallel


.github/workflows/autorelease-pytorch-ec2-cpu.yml
```diff
@@ -15,10 +20,32 @@ permissions:
...
-  CONFIG_FILE: ".github/config/image/pytorch-ec2-cpu.yml"
 
 jobs:
+  determine-version:
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.version.outputs.version }}
+      config-file: ${{ steps.version.outputs.config-file }}
+    steps:
+      - name: Determine PyTorch version
```
sirutBuasai: Why do we need to determine pytorch version and then construct a config file name based off that? Why dont we simply use the config file as input, or simple do 
```
case "$CRON" in
    "00 17 * * 1,3") CONF_FILE="<path_to_pt2.11_cpu>" ;;
    # "10 17 * * 1,3") CONF_FILE="<path_to_pt2.12_cpu>"  ;;
    *) echo "::error::Unknown cron: $CRON"; exit 1 ;;
esac
```

Then any usage of the actual Pytorch version field can just be derived from the config file using the output from load_config job.

.github/workflows/autorelease-pytorch-ec2-cuda.yml
```diff
@@ -16,10 +20,32 @@ permissions:
 
 env:
   FORCE_COLOR: "1"
-  CONFIG_FILE: ".github/config/image/pytorch-ec2-cuda.yml"
 
 jobs:
+  determine-version:
```
sirutBuasai: ditto

.github/workflows/autorelease-pytorch-sagemaker-cpu.yml
```diff
@@ -15,10 +20,32 @@ permissions:
 
 env:
   FORCE_COLOR: "1"
-  CONFIG_FILE: ".github/config/image/pytorch-sagemaker-cpu.yml"
 
 jobs:
+  determine-version:
```
sirutBuasai: ditto

.github/workflows/autorelease-pytorch-sagemaker-cuda.yml
```diff
@@ -15,10 +20,32 @@ permissions:
 
 env:
   FORCE_COLOR: "1"
-  CONFIG_FILE: ".github/config/image/pytorch-sagemaker-cuda.yml"
 
 jobs:
+  determine-version:
```
sirutBuasai: ditto

.github/workflows/pr-pytorch-ec2-cpu.yml
```diff
@@ -43,11 +43,70 @@ jobs:
...
+        uses: pre-commit/action@v3.0.1
+        with:
+          extra_args: --all-files
+
+      - name: Detect PyTorch version
+        id: version
+        run: |
+          VERSION=$(git diff --name-only origin/main...HEAD \
+            | grep -oP 'docker/pytorch/\K[0-9]+\.[0-9]+' \
+            | sort -u | head -1)
```
sirutBuasai: Multi-version PR silently tests only one version. so if we make a change for both PT 2.11 and PT 2.12, we would expect both jobs to run. The current logic only run one PR

.github/workflows/pr-pytorch-ec2-cpu.yml
```diff
@@ -43,11 +43,70 @@ jobs:
...
+              - "test/sanity/**"
+            telemetry-test-change:
+              - "test/telemetry/**"
+
   # ============================================================
   # Load configuration from YAML
   # ============================================================
   load-config:
-    needs: [gatekeeper]
+    needs: [gatekeeper, check-changes]
```
sirutBuasai: we should be able to parallelize the check-changes with load-config. Serializing this slows down the PR a little bit due to these two jobs being extremely small but the job startup each takes 30s

.github/workflows/pr-pytorch-ec2-cpu.yml
sirutBuasai, line 1: Take a look at https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstrategymatrix matrix job submission. This would allow us to define a list of configurations that the workflow can in sets in parallel of each other.

https://github.com/aws/deep-learning-containers/blob/main/.github/workflows/dispatch-vllm-benchmark.yml is also another good example that we have that uses matrix run mechanism

.github/workflows/pr-pytorch-ec2-cpu.yml
```diff
@@ -5,7 +5,7 @@ on:
     branches: [main]
     types: [opened, reopened, synchronize]
     paths:
-      - ".github/config/image/pytorch-ec2-cpu.yml"
+      - ".github/config/image/pytorch-*-ec2-cpu.yml"
       - ".github/workflows/pr-pytorch-ec2-cpu.yml"
       - "docker/pytorch/**"
```
sirutBuasai: this is too broad, this means that a change in cuda dockerfile will trigger cpu as well

.github/workflows/pr-pytorch-ec2-cpu.yml
```diff
@@ -164,7 +178,8 @@ jobs:
       - name: Build runtime image
         id: build-runtime
         run: |
-          source docker/pytorch/versions-cpu.env
+          VERSION="${{ needs.check-changes.outputs.pytorch-version }}"
+          source docker/pytorch/${VERSION}/versions-cpu.env
           CI_IMAGE_URI="${{ vars.CI_AWS_ACCOUNT_ID }}.dkr.ecr.${{ vars.AWS_REGION }}.amazonaws.com/ci:pytorch-cpu-runtime-pr-${{ github.event.pull_request.number }}"
```
sirutBuasai: With matrix running, we'll need to add some kind of version string to the ci image. This is because if PT 2.11 and PT 2.12 runs at the same time, we run into naming collision

.github/workflows/pr-pytorch-ec2-cuda.yml
sirutBuasai, line 1: ditto

.github/workflows/pr-pytorch-sagemaker-cpu.yml
sirutBuasai, line 1: ditto

.github/workflows/pr-pytorch-sagemaker-cuda.yml
sirutBuasai, line 1: ditto

docker/pytorch/2.11/Dockerfile.cuda
```diff
@@ -46,14 +46,14 @@ ENV UV_PROJECT_ENVIRONMENT="/opt/venv"
 RUN python${PYTHON_VERSION} -m venv /opt/venv
 ENV PATH="/opt/venv/bin:${PATH}"
 
-COPY docker/pytorch/cuda/pyproject.toml docker/pytorch/cuda/uv.lock /tmp/build/
+COPY docker/pytorch/2.11/cuda/pyproject.toml docker/pytorch/2.11/cuda/uv.lock /tmp/build/
```
sirutBuasai: If you're going to have to declare the version across the file, we should use `ARG PYTORCH_VERSION=2.11` then use `COPY .../${PYTORCH_VERSION}/...`

.github/workflows/autorelease-pytorch-ec2-cpu.yml
Eren-Jeager123: Good point I forgot we can use the one from load_config

.github/workflows/pr-pytorch-ec2-cpu.yml
Eren-Jeager123: I thought that it's rare to change more than 1 version each time previously; Ya however I will change this.

.github/workflows/autorelease-pytorch-ec2-cpu.yml
```diff
@@ -15,10 +20,29 @@ permissions:
 
 env:
   FORCE_COLOR: "1"
-  CONFIG_FILE: ".github/config/image/pytorch-ec2-cpu.yml"
 
 jobs:
+  determine-config:
```
sirutBuasai: You can probably merge this with the load-config job. Make it load-config job with two steps instead of two different jobs. This way we're not using 2 ubuntu-latest instances.

.github/workflows/autorelease-pytorch-ec2-cuda.yml
sirutBuasai, line 1: ditto

.github/workflows/autorelease-pytorch-sagemaker-cpu.yml
sirutBuasai, line 1: ditto

.github/workflows/autorelease-pytorch-sagemaker-cuda.yml
sirutBuasai, line 1: ditto

.github/workflows/pr-pytorch-ec2-cpu.yml
```diff
@@ -134,17 +135,21 @@ jobs:
...
+    strategy:
+      matrix:
+        include: ${{ fromJson(needs.detect-versions.outputs.configs) }}
+      fail-fast: false
     concurrency:
-      group: ${{ github.workflow }}-build-${{ github.event.pull_request.number }}
+      group: ${{ github.workflow }}-build-${{ matrix.version }}-${{ github.event.pull_request.number }}
       cancel-in-progress: true
     outputs:
       runtime-image-uri: ${{ steps.build-runtime.outputs.image-uri }}
```
sirutBuasai: We're determining the image uri based on the matrix value now so there's no need for this output

.github/workflows/autorelease-pytorch-ec2-cuda.yml
sirutBuasai: unnecessary backslash, not sure if this affects the workflow parsing logic but we should remove it.

.github/workflows/autorelease-pytorch-sagemaker-cuda.yml
```diff
@@ -129,17 +156,19 @@ jobs:
...
-          source docker/pytorch/versions-cuda.env
+          VERSION=$(echo "${{ needs.load-config.outputs.framework-version }}" | cut -d. -f1,2)
+          source docker/pytorch/${VERSION}/versions-cuda.env
           bash scripts/pytorch/upload_cached_wheels.sh \
             "${{ vars.WHEEL_CACHE_BUCKET }}" \
             "${CUDA_VERSION}" "${TORCH_VERSION}" "${PYTHON_VERSION}" \
             "${{ steps.build-sagemaker.outputs.image-uri }}" \
+            "docker/pytorch/${VERSION}/Dockerfile.cuda" \
             "flash-attn:${FLASH_ATTN_VERSION}" \
             "transformer-engine-torch:${TRANSFORMER_ENGINE_VERSION}" \
```
sirutBuasai: ditto

docker/pytorch/2.11/versions-cpu.env
sirutBuasai, line 1: I'm actually curious about this. What's the reason for not combining this with the config file? That way we only have one file to track all the package versions, etc. You dont necessarily to change this in this PR cuz it's out of scope, but I'm curious

.github/workflows/pr-pytorch-ec2-cpu.yml
sirutBuasai: The downside of doing matrix though is that before, if changes aren't made to the image, we dont build and simply run a prod image on specfic test. By changing it to matrix, we're determining image URI based on matrix value.

We technically dont want to lose that functionality of not having to build every single time. Can you see if it's possible to retain that functionality even with the matrix? Essentially, if the check-changes return that no build-change are made, it will point to prod image URI. In the worst case scenario where we can't figure out a sophisticated way to point to prod URI directly, we may need to add a job to pull prod image, tag it to the CI tag, then push it to CI repository so that all other jobs can reference it.

.github/workflows/autorelease-pytorch-ec2-cpu.yml
```diff
@@ -83,10 +107,12 @@ jobs:
       - name: Build runtime image
         id: build-runtime
         run: |
-          source docker/pytorch/versions-cpu.env
+          VERSION=$(echo "${{ needs.load-config.outputs.framework-version }}" | cut -d. -f1,2)
```
sirutBuasai: Instead of having to do `VERSION=$(...<version parseing>)` in each job, in the `load-config` step, why dont we just add another output name `FRAMEWORK_SHORT_VER` that does all the parsing already and the downstream job just use that? That way if there are changes to the logic, we only need to change it once.

.github/workflows/pr-pytorch-ec2-cpu.yml
Eren-Jeager123: I will read the config.prod_image and add the condition when we want to use prod image instead of building image.

docker/pytorch/2.11/versions-cpu.env
Eren-Jeager123: This file was added before I implemented the PyTorch 2.11 currency & migration by Junpu. I think maybe those are mainly being used for Dockerfile args, but I don't see any blockers to merge them to config file. Let's discuss this in the future

.github/workflows/pr-pytorch-ec2-cpu.yml
```diff
@@ -44,61 +46,18 @@ jobs:
...
-          echo "arch-type=$(jq -r '.common.arch_type // "x86"' config.json)" >> $GITHUB_OUTPUT
-          echo "contributor=$(jq -r '.common.contributor // "None"' config.json)" >> $GITHUB_OUTPUT
-          echo "customer-type=$(jq -r '.common.customer_type // ""' config.json)" >> $GITHUB_OUTPUT
-          echo "prod-image=$(jq -r '.common.prod_image' config.json)" >> $GITHUB_OUTPUT
-
-  # ============================================================
-  # Pre-commit + change detection
-  # ============================================================
-  check-changes:
+  detect-versions:
```
sirutBuasai: nit: rather than `detect-versions`, lets name it `check-changes`. The version detection is just a preliminary task rather than the main task of this job.

.github/workflows/pr-pytorch-ec2-cuda.yml
sirutBuasai, line 1: ditto on the nit

.github/workflows/pr-pytorch-sagemaker-cpu.yml
sirutBuasai, line 1: ditto on the nit

.github/workflows/pr-pytorch-sagemaker-cuda.yml
sirutBuasai, line 1: ditto on the nit

sirutBuasai (dismissed):
Another minor nit:

On all the autorelease-pytorch* and pr-pytorch* workflows, can you update the following actions if applicable?

```
actions/checkout@v* -> actions/checkout@v6
actions/upload-artifact@v* -> actions/upload-artifact@v7
actions/download-artifact@v* -> actions/download-artifact@v8
```

sirutBuasai (approved):

PR #6096, opened 2026-05-13
feat(vllm): add libsndfile, /invocations route middleware, and Voxtral ASR smoke test

Jyothirmaikottu:
## Summary

- Add SageMaker route middleware to forward /invocations to custom endpoints via CustomAttributes header
- Add Voxtral-Mini-4B transcription smoke test
- Allowlist CVE-2026-44222 (false positive — 0.20.0.dev361 includes the fix)

## Test plan

- [ ] Voxtral ASR smoke test passes
- [ ] Existing smoke tests unaffected
- [ ] Security scan passes


test/vllm/scripts/amzn2023/vllm_asr_smoke_test.sh
```diff
@@ -39,6 +39,12 @@ echo "=== Model directory: ${MODEL_DIR} ==="
 echo "=== Test fixtures ==="
 ls -lh "${FIXTURES_DIR}"
 
+if [ -f "${MODEL_DIR}/requirements.txt" ]; then
+  echo "=== Installing model dependencies ==="
+  cat "${MODEL_DIR}/requirements.txt"
+  pip install -r "${MODEL_DIR}/requirements.txt" av
```
Jyothirmaikottu: cannot load audio files without av, which is a ffmpeg based package.

docker/vllm/Dockerfile.amzn2023
```diff
@@ -166,7 +166,7 @@ ARG PYTORCH_CUDA_INDEX_BASE_URL=https://download.pytorch.org/whl
 RUN dnf install -y --setopt=install_weak_deps=False \
   tar gzip gcc gcc-c++ \
   python${PYTHON_VERSION} python${PYTHON_VERSION}-devel python${PYTHON_VERSION}-pip \
-  libibverbs-devel \
+  libibverbs-devel libsndfile \
```
Yadan-Wei: We only copy python deps installed in this stage to final stage, do you need it here or in the final stage?

docker/vllm/Dockerfile.amzn2023
Yadan-Wei, line 268: Here is the copy command I mentioned above.

test/vllm/sagemaker/test_sm_model_serving.py
```diff
@@ -0,0 +1,278 @@
+"""Config-driven SageMaker endpoint test for vLLM DLC models.
+
```
Yadan-Wei, line 2: Since we already have an endpoint test https://github.com/aws/deep-learning-containers/blob/main/test/vllm/sagemaker/test_sm_endpoint.py,  that does the deploy/invoke/cleanup loop. Can we extend that file with YAML-driven cases instead of adding a parallel one? 
 Concretely: move the existing chat test into a YAML entry, generalize the fixture to handle both model_id (HF) and s3_model (this PR), and parametrize over (model, test_case) with a module-scoped endpoint. Keeps a single source of truth for endpoint tests.

docker/vllm/Dockerfile.amzn2023
Jyothirmaikottu: Not particular, will move it to final stage.

docker/vllm/Dockerfile.amzn2023
```diff
@@ -391,23 +413,23 @@ RUN dnf upgrade -y --security --releasever latest --setopt=install_weak_deps=Fal
   && ln -sf /opt/venv/bin/python3 /usr/bin/python3
 
 COPY ./scripts/vllm/sagemaker_entrypoint.sh /usr/local/bin/sagemaker_entrypoint.sh
+COPY ./scripts/vllm/sagemaker_serve.py /usr/local/bin/sagemaker_serve.py
 RUN chmod +x /usr/local/bin/sagemaker_entrypoint.sh
+ENV PYTHONPATH="/usr/local/bin:${PYTHONPATH}"
```
Jyothirmaikottu, line 419: vLLM loads middleware by module name, so the directory with sagemaker_serve.py needs to be on PYTHONPATH for module discovery. The venv symlink handles the interpreter but not module paths for files outside the venv.

test/vllm/sagemaker/test_sm_model_serving.py
Jyothirmaikottu: I made the test config driven so that we can do model testing even on sagemaker. I will see if we can modify the existing test to be both config driven and sm test.

test/vllm/sagemaker/test_sm_model_serving.py
```diff
@@ -0,0 +1,278 @@
...
+        else:
+            body, ct = _build_json_body(request_cfg)
+
+        invoke_kwargs = {
+            "EndpointName": endpoint_name,
+            "Body": body,
+            "ContentType": ct,
+        }
+        if "route" in tc:
+            invoke_kwargs["CustomAttributes"] = f"x-amzn-sagemaker-route={tc['route']}"
```
Yadan-Wei: **`CustomAttributes` value mismatches between docstring and test.**

  - Docstring example: `CustomAttributes="route=/v1/audio/transcriptions"`
  - Test sends: `f"x-amzn-sagemaker-route={tc['route']}"`

  Both happen to work because the regex `r"route=(/[^\s,]+)"` matches `route=/...` as a substring inside `x-amzn-sagemaker-route=/...`. That's accidental. Pick one form —
  the bare `route=...` is shorter and matches the existing docstring; I'd remove the `x-amzn-sagemaker-` prefix from the test.

test/vllm/sagemaker/test_sm_model_serving.py
```diff
@@ -0,0 +1,278 @@
...
+            ),
+        ],
+    )
+
+    LOGGER.info(f"Deploying endpoint: {endpoint_name}")
+    endpoint = Endpoint.create(
+        endpoint_name=endpoint_name,
+        endpoint_config_name=endpoint_name,
+    )
+    endpoint.wait_for_status("InService")
```
Yadan-Wei: Would it be better to have a timeout here to avoid deployment hung?

test/vllm/sagemaker/test_sm_model_serving.py
```diff
@@ -0,0 +1,278 @@
...
+def _download_s3(s3_client, s3_uri):
+    bucket = s3_uri.split("/")[2]
+    key = "/".join(s3_uri.split("/")[3:])
+    buf = io.BytesIO()
+    s3_client.download_fileobj(bucket, key, buf)
+    buf.seek(0)
+    return buf.read()
+
+
+def _build_multipart_body(request_cfg, s3_client):
```
Yadan-Wei, line 65: From this test, it seems that SageMaker endpoint can accept multipart form data directly, so probably we can remove the conversion in middleware, let user build request body instead of passing Json and let middleware handle?

scripts/vllm/sagemaker_serve.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+def _get_content_type(headers: list[tuple[bytes, bytes]]) -> str:
+    for key, value in headers:
+        if key.lower() == b"content-type":
+            return value.decode().lower()
+    return ""
+
+
+def _build_multipart_body(data: dict, boundary: str) -> bytes:
```
Yadan-Wei: From the test you run by building multipart form data body directly and pass to endpoint, looks like Content-ype can be other than JSON, if that is the case, we do not need this part. I also need to check omni's setup.
Refer to https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_runtime_InvokeEndpoint.html

docker/vllm/Dockerfile.amzn2023
Yadan-Wei: A followup question is if the model works without this lib, do we really need it in the Dockerfile

Yadan-Wei (commented):
We have a major version, minor version argument in https://github.com/aws/deep-learning-containers/blob/main/docker/vllm/versions.env#L44, each time we add new compatible change, we need to bump minor version, for this case we should bump it too.

test/vllm/sagemaker/test_sm_model_serving.py
Jyothirmaikottu: Fixed — switched to bare `route=/v1/audio/transcriptions` matching the docstring.

test/vllm/sagemaker/test_sm_model_serving.py
Jyothirmaikottu: Added 30min timeout to wait_for_status.

test/vllm/sagemaker/test_sm_model_serving.py
Jyothirmaikottu: Agreed, removed the JSON-to-multipart conversion entirely middleware is now routing-only since SageMaker passes ContentType/Body.

scripts/vllm/sagemaker_serve.py
Jyothirmaikottu: Confirmed — stripped the middleware to pure routing, no content-type handling needed.

test/vllm/sagemaker/test_sm_model_serving.py
Jyothirmaikottu: Keeping separate for now — this test handles multipart file uploads, S3 fixtures, and custom route dispatch which would over-complicate the existing chat-focused test

docker/vllm/Dockerfile.amzn2023
Jyothirmaikottu: We need it for network isolation cases.

Yadan-Wei (approved):

PR #6107, opened 2026-05-18
[TensorFlow][Training][Sagemaker] TensorFlow 2.21.0 Currency Release

bhanutejagk:
TensorFlow 2.21 training DLC for SageMaker on Amazon Linux 2023, CUDA 12.9.1, Python 3.12, x86. Targets the post-refactor `main` branch architecture (per-framework dot-namespaced workflows, per-image config YAMLs under `.github/config/image/tensorflow/`, `scripts/docker/` + `scripts/ci/` split, `docker/<framework>/<version>/` layout). Replaces the legacy 2.19 master line.

## Files

**Image build**:
- `docker/tensorflow/2.21/Dockerfile.{cuda,cpu}` — multi-stage builder-base → builder-oss → runtime-base → sagemaker. CUDA image pre-bakes `all_reduce_perf` for EFA validation.
- `docker/tensorflow/2.21/{cuda,cpu}/pyproject.toml` + `uv.lock` — full dependency graph resolved by uv. Security pins for aiohttp 3.14.1, PyJWT 2.13, cryptography 48, starlette 1.3, tornado 6.5.

**CI**:
- `.github/config/image/tensorflow/2.21-sagemaker-{cpu,cuda}.yml` — image configs (metadata + build args + release flags).
- `.github/workflows/tensorflow.pipeline.yml` — reusable orchestrator: build → sanity → security → telemetry → unit → single-gpu → sagemaker → efa.
- `.github/workflows/tensorflow.pr-2.21-{cpu,cuda}.yml` — PR triggers.
- `.github/workflows/tensorflow.autorelease-2.21-sagemaker.yml` — autorelease (cron: Tue/Thu 17:30 UTC, registered in `release-schedule.yml`).
- `.github/workflows/tensorflow.tests-{unit,single-gpu,sagemaker}.yml` — framework-specific test runners.

**Tests**:
- `test/tensorflow/unit/` — TF-specific unit tests (framework-agnostic checks migrated out to `test/sanity/scripts/test_sanity_training.py` per review feedback).
- `test/tensorflow/single_gpu/` — training smoke, data pipeline, mixed precision, GPU memory. Pool-reuse memory test handles TF's BFC allocator correctly (no test-time flake).
- `test/tensorflow/integration/sagemaker/` — SDK v3 `ModelTrainer` MNIST tests: single-host CPU, multi-host CPU, MWMS CPU, single-GPU, multi-host GPU, MirroredStrategy multi-GPU, MWMS multi-host GPU, plus Experiments and HPO. Uses a custom training loop with `strategy.run` + per-replica gradient aggregation (Keras 3 `fit()` silently skips MWMS). Conftest inherits `image_uri` + `region` from `test/conftest.py` — no duplication.
- `test/sanity/scripts/test_sanity_training.py` — shared training-image sanity suite (env, PATH, binaries, SSH config, `/opt/ml` paths, OpenMPI, EFA + NCCL, CUDA runtime, cuDNN, license artifacts, venv, entrypoint). Framework-guarded via `@tensorflow_only` and `@training_cluster_only` decorators so PT and xgboost images run only the applicable subset.

**Security**:
- `test/security/data/ecr_scan_allowlist/tensorflow/framework_allowlist.json` — 4 entries (2 CUDA toolkit CVEs, RUSTSEC-2026-0185 quinn-proto, RUSTSEC-2026-0195 quick-xml — both bundled in the `uv` binary shipped for customer venv management). All have per-CVE rationale + review-by dates. Source-fix preferred over allowlist wherever possible.

## Notable scope decisions

- `sagemaker>=3.4.0` (master used `<3`).
- CPU image uses the `pytorch-cpu` index to keep NVIDIA libs out.
- `tf-models-official` and `tensorflow-text` dropped — no 2.21-compatible PyPI release from Google.
- EFA 1.49.0 with `--disable-ngc` flag to prevent the installer from misdetecting `nvidia/cuda:*-base-amzn2023` as an NGC container and skipping the OFI-NCCL RPM.
- `numpy>=2.1,<2.5` (was `==1.26.4`) — numba 0.65.1 caps `<2.5`.
- `cython<3` constraint removed (no longer needed for TF 2.21).
- Boost built `--without-python` to skip pyste/libboost_python dep chain.

## Out of scope (follow-up PRs)

- Release notes data files (`docs/src/data/tensorflow-training/2.21-*.yml`) and `docs/tensorflow/changelog/index.md`.
- TF 2.21 inference image (separate currency PR; mirrors 2.19 inference release sequence).
- SageMaker SDK `image_uri_config` update (`aws/sagemaker-python-sdk` cross-team handoff after GA).

## Test Plan

- PR CI: full sanity + security + telemetry + unit + single-gpu + sagemaker + efa suite across CPU + CUDA configs.
- SageMaker remote training: plain Keras single-host, multi-host CPU, MirroredStrategy single-host multi-GPU, MultiWorkerMirroredStrategy multi-host multi-GPU with NCCL.
- EFA: 2x p4d.24xlarge NCCL `all_reduce_perf` with EFA provider verification.

## Test Result

- **All TF 2.21 CI checks green at HEAD `67c7dace`**: build, ci-config, sanity-test, sanity-sagemaker, sagemaker-test, unit-test, telemetry, security-test, single-gpu-test, and **efa-test** across both `2.21-sagemaker-cpu` and `2.21-sagemaker-cuda` configs.
- Remaining PR-wide reds are all cross-framework and unrelated to TF 2.21: framework-owned ecr-scan CVEs on openfold3/ray/sglang/vllm/vllm-omni images, plus intermittent p4d ODCR / SSH-propagation flakes on other frameworks' efa-tests.
- ECR Enhanced Scan for the TF 2.21 images: 0 non-allowlisted HIGH/CRITICAL findings.
- EFA test: 2-node NCCL `all_reduce_perf` passes with `aws-ofi-nccl` + EFA provider + GDRDMA on `p4d.24xlarge`.

<details>

## PR Checklist

</details>


scripts/docker/tensorflow/entrypoint.sh
sallyseok, line 1: Let's call this just entrypoint.sh? To keep consistency with pytorch (and the other frameworks that use `dockerd_entrypoint.sh` for ec2 images only)

bhanutejagk: I see that the other frameworks use dockered_entrypoint for ec2 images and sagemaker_entrypoint for sm images. Then should we use sagemaker_entrypoint here since it is for sm image? 

In this case we have to change pytorch convention too. So its a tradeoff between being consistent with all other frameworks or just being ok with tensorflow and pytorch being unique.

what do you think?

test/tensorflow/integration/sagemaker/resources/mnist/data/test/x_test.npy
sallyseok, line 1: I'm guessing there's no way to fetch these datasets using tensorflow in mnist.py?

test/tensorflow/integration/sagemaker/conftest.py
sallyseok, line 1: Is this file necessary?

test/tensorflow/integration/TODO.md
sallyseok, line 1: We can remove this file, it's basically the same as the PT version and I don't think anyone is even tracking that one

scripts/docker/tensorflow/entrypoint.sh
bhanutejagk: Oh i see pytorch uses same entrypoint.sh script for both ec2 and sm images.

sallyseok: Yeah, looks like PT uses a shared one. Since we don't have ec2 for TF just entrypoint.sh makes sense

bhanutejagk: sounds good

test/tensorflow/integration/sagemaker/resources/mnist/data/test/x_test.npy
bhanutejagk: I checked the pytorch way of implementation. PT pre-stages in 
`s3://dlc-cicd-models/pytorch/sagemaker-test-data/` and should we mirror that for tensorflow?

test/tensorflow/integration/sagemaker/conftest.py
bhanutejagk: I think its a no op. we can clean up this too.

test/tensorflow/integration/TODO.md
bhanutejagk: sure.

test/tensorflow/integration/sagemaker/resources/mnist/data/test/x_test.npy
bhanutejagk: Uploaded the data to s3 using aws cli mirroring Pytorch and edited the files to make use of the data from s3 in the commit - [bf041a5](https://github.com/aws/deep-learning-containers/pull/6107/commits/bf041a50153de60e39e635efd7709765d766c37e)

sallyseok (dismissed):

.github/config/image/tensorflow/2.21-sagemaker-cuda.yml
```diff
@@ -0,0 +1,35 @@
...
+build:
+  dockerfile: "docker/tensorflow/2.21/Dockerfile.cuda"
+  target: "sagemaker"
+  python_version: "3.12"
+  cuda_version: "12.9.1"
+  tf_version: "2.21.0"
+  nccl_version: "2.29.7-1"
+  nccl_ofi_version: "1.19.0"
+  open_mpi_version: "4.1.8"
+  efa_version: "1.48.0"
```
sirutBuasai: I think latest EFA version now is 1.49.

.github/config/image/tensorflow/2.21-sagemaker-cuda.yml
```diff
@@ -0,0 +1,35 @@
...
+  prod_image: "tensorflow:2.21-cu129-amzn2023-sagemaker"
+
+build:
+  dockerfile: "docker/tensorflow/2.21/Dockerfile.cuda"
+  target: "sagemaker"
+  python_version: "3.12"
+  cuda_version: "12.9.1"
+  tf_version: "2.21.0"
+  nccl_version: "2.29.7-1"
+  nccl_ofi_version: "1.19.0"
```
sirutBuasai: OFI version is bundled with EFA so this version here isn't really used in dockerfile so there's no need to declare here or in dockerfile at all.

.github/workflows/tensorflow.pipeline.yml
```diff
@@ -0,0 +1,216 @@
...
+          ENVIRONMENT=$(yq '.release.environment' "${{ inputs.config-file }}")
+          echo "should-release=${SHOULD_RELEASE}" >> $GITHUB_OUTPUT
+          echo "environment=${ENVIRONMENT}" >> $GITHUB_OUTPUT
+      - id: spec
+        if: steps.check.outputs.should-release == 'true'
+        uses: ./.github/actions/generate-release-spec
+        with:
+          config-file: ${{ inputs.config-file }}
+
+  release:
```
sirutBuasai, line 207: You'll need to update this condition here based on Sally's bug fix https://github.com/aws/deep-learning-containers/commit/66f97123b6e38b61ad828a97fee61ec59e7050da

.github/workflows/tensorflow.pr-2.21-cpu.yml
```diff
@@ -0,0 +1,101 @@
...
+      matrix:
+        include: ${{ fromJson(needs.discover.outputs.configs) }}
+    uses: ./.github/workflows/tensorflow.pipeline.yml
+    with:
+      config-file: ${{ matrix.config_file }}
+      tag-suffix: pr-${{ github.event.pull_request.number }}
+      build: ${{ needs.check-changes.outputs.build-change == 'true' }}
+      run-sanity-test: ${{ needs.check-changes.outputs.build-change == 'true' || needs.check-changes.outputs.sanity-test-change == 'true' }}
+      run-security-test: ${{ needs.check-changes.outputs.build-change == 'true' }}
+      run-single-gpu-test: false
```
sirutBuasai: nit: there's no need to actually declare this since the pipeline already have the condition where device_type needs to be gpu anyway. This is really nitpicky though so it's up to you if you wanna remove this.

.github/workflows/tensorflow.tests-unit.yml
```diff
@@ -0,0 +1,78 @@
...
+          docker exec ${CONTAINER_ID} pip install -r /workdir/test/requirements.txt -q
+          docker exec \
+            -e PYTHONPATH=/workdir/test \
+            -e DLC_WORKDIR=/workdir \
+            -e DLC_TENSORFLOW_VERSION=${{ steps.fw.outputs.short-version }} \
+            ${CONTAINER_ID} \
+            pytest /workdir/test/tensorflow/unit/ -v
+          docker kill ${CONTAINER_ID}
+
+      - name: Skip notice
```
sirutBuasai: Can you delete all skip notices? We actually dont need this and it's just a waste of compute.

docker/tensorflow/2.21/cuda/pyproject.toml
```diff
@@ -0,0 +1,76 @@
...
+    # cryptography>=48.0.1 — GHSA-537c-gmf6-5ccf (HIGH)
+    # starlette>=1.3.1 — GHSA-82w8-qh3p-5jfq + GHSA-wqp7-x3pw-xc5r (HIGH)
+    # tornado>=6.5.6   — GHSA-3x9g-8vmp-wqvf + GHSA-mgf9-4vpg-hj56 (HIGH)
+    "PyJWT>=2.13.0",
+    "cryptography>=48.0.1",
+    "starlette>=1.3.1",
+    "tornado>=6.5.6",
+    # TF compat pins
+    "cython<3",
+    "numpy==1.26.4",
```
sirutBuasai: Is current tensorflow version not working with numpy v2?

numpy v1 is really old now so ideally we'd like to upgrade.

docker/tensorflow/2.21/Dockerfile.cuda
```diff
@@ -0,0 +1,285 @@
...
+# ============================================================================
+
+ARG DLC_MAJOR_VERSION=1
+ARG DLC_MINOR_VERSION=0
+ARG CUDA_VERSION=12.9.1
+ARG PYTHON_VERSION=3.12
+ARG TF_VERSION=2.21.0
+ARG OPEN_MPI_VERSION=4.1.8
+ARG NCCL_VERSION=2.29.7-1
+ARG NCCL_OFI_VERSION=1.19.0
```
sirutBuasai: You can remove this, it's not being used

.github/config/image/tensorflow/2.21-sagemaker-cuda.yml
```diff
@@ -0,0 +1,35 @@
...
+  job_type: "training"
+  prod_image: "tensorflow:2.21-cu129-amzn2023-sagemaker"
+
+build:
+  dockerfile: "docker/tensorflow/2.21/Dockerfile.cuda"
+  target: "sagemaker"
+  python_version: "3.12"
+  cuda_version: "12.9.1"
+  tf_version: "2.21.0"
+  nccl_version: "2.29.7-1"
```
sirutBuasai: EFA 1.49, uses nccl 2.30.4-1

docker/tensorflow/2.21/Dockerfile.cuda
```diff
@@ -0,0 +1,285 @@
...
+RUN NCCL_HOME=$(/opt/venv/bin/python -c "import nvidia.nccl; print(nvidia.nccl.__path__[0])") \
+  && if [ ! -e "${NCCL_HOME}/lib/libnccl.so" ]; then \
+       ln -s "$(basename $(ls ${NCCL_HOME}/lib/libnccl.so.* | head -1))" "${NCCL_HOME}/lib/libnccl.so"; \
+     fi \
+  && cd /tmp \
+  && curl -fsSL https://github.com/NVIDIA/nccl-tests/archive/refs/heads/master.tar.gz | tar xz \
+  && cd nccl-tests-master \
+  && make MPI=1 MPI_HOME=/opt/amazon/openmpi NCCL_HOME=${NCCL_HOME} CUDA_HOME=/usr/local/cuda \
+  && cp build/all_reduce_perf /usr/local/bin/all_reduce_perf \
+  && cd / && rm -rf /tmp/nccl-tests-master
```
sirutBuasai: Why do we need nccl-tests built into the container? Historically, we've never built nccl-tests into the container and simply build it at test time. This is done so that we're not bloating the image size. Is there customer signal that they dont mind bigger image size in favor of having a built in nccl-tests? Moreover, this nccl-tests is using master and not a pinned ref, in which case it will easily break CI tests over time.

docker/tensorflow/2.21/Dockerfile.cuda
```diff
@@ -0,0 +1,285 @@
...
+RUN --mount=type=cache,target=/root/.cache/uv cd /tmp/build && uv sync --frozen --no-dev --extra sagemaker --no-install-project --inexact \
+  && rm -rf /tmp/build /tmp/uv-*
+
+# Boost 1.82 for SM Studio extensions (compiled with _GLIBCXX_USE_CXX11_ABI=1
+# to match TF's ABI). Some SM Studio extensions link against boost.
+RUN curl -fsSLO "https://sourceforge.net/projects/boost/files/boost/1.82.0/boost_1_82_0.tar.gz/download" \
+  && mv download boost_1_82_0.tar.gz \
+  && tar xzf boost_1_82_0.tar.gz && cd boost_1_82_0 \
+  && ./bootstrap.sh \
+  && ./b2 define=_GLIBCXX_USE_CXX11_ABI=1 threading=multi --prefix=/usr -j$(nproc) cxxflags=-fPIC cflags=-fPIC install || true \
```
sirutBuasai: any reason why we're doing `|| true`? This wont catch any build time failures.

test/tensorflow/integration/sagemaker/test_mnist_cuda.py
```diff
@@ -0,0 +1,176 @@
...
+
+def _assert_s3_file_exists(region, s3_url):
+    """Verify that the given s3:// URL points to an existing object.
+
+    head-object via boto3 raises if the key is missing, which surfaces as a
+    clear test failure when SageMaker didn't upload the model artifact."""
+    parsed_url = urlparse(s3_url)
+    s3 = boto3.resource("s3", region_name=region)
+    s3.Object(parsed_url.netloc, parsed_url.path.lstrip("/")).load()
+
```
sirutBuasai: These two functions are duplicated between `mnist_cpu` and `mnist_cuda`, we can combine these into conftest.py and use it from the fixture or helper there. Look into best practices for running fixtures in pytest. Or at the very least DRY (don't repeat yourself) and abstract common code into a reusable funciton.

test/tensorflow/integration/sagemaker/test_mnist_cuda.py
```diff
@@ -0,0 +1,176 @@
...
+RESOURCE_DIR = os.path.join(os.path.dirname(__file__), "resources")
+SOURCE_DIR = os.path.join(RESOURCE_DIR, "scripts")
+SINGLE_GPU_INSTANCE = "ml.g4dn.xlarge"
+# DLC test account has ml.g4dn.xlarge quota=1; multi-host needs 2 instances.
+# ml.g4dn.2xlarge has quota=8, same T4 GPU.
+MULTI_HOST_GPU_INSTANCE = "ml.g4dn.2xlarge"
+MULTI_GPU_INSTANCE = "ml.g4dn.12xlarge"
+IMAGE_URI = os.environ["TEST_IMAGE_URI"]
+DEFAULT_REGION = "us-west-2"
+MNIST_S3_URI = "s3://dlc-cicd-models/tensorflow/sagemaker-test-data/MNIST/"
```
sirutBuasai: In these variables, if they're commonly reused across multiple files, we should declare it in conftest.py instead. That way, if we need to change a variable, we don't need to change them in multiple places across files.

sirutBuasai: More over, some of these should also be a fixture or be a pytest parametrize arg, eg: instnace type. Look at existing pytest python files like sglang or vllm sagemaker test.

test/tensorflow/single_gpu/test_data_pipeline.py
```diff
@@ -0,0 +1,43 @@
+"""tf.data pipeline smoke tests — runs on CPU and GPU containers."""
```
sirutBuasai: This says run on cpu and gpu but why is it in single_gpu directory where it gets skipped on cpu?

test/tensorflow/pytest.ini
```diff
@@ -0,0 +1 @@
+[pytest]
```
sirutBuasai: If there are no useful markers to be declared, we don't need this file

test/tensorflow/integration/sagemaker/test_experiments_cpu.py
```diff
@@ -0,0 +1,128 @@
...
+    After `train(wait=True)` returns, we read the training job ARN from
+    `model_trainer._latest_training_job` (avoiding ListTrainingJobs
+    eventual-consistency races) and list TrialComponents whose `source_arn`
+    matches the job ARN to verify auto-creation."""
+    boto_session = boto3.session.Session(region_name=DEFAULT_REGION)
+    sagemaker_session = Session(boto_session)
+
+    # Random unique-id to keep concurrent CI runs from colliding.
+    random.seed(f"{datetime.datetime.now().strftime('%Y%m%d%H%M%S%f')}")
+    unique_id = random.randint(1, 6000)
```
sirutBuasai: why not just use `random_suffix_name` function

test/tensorflow/unit/test_environment.py
sirutBuasai, line 1: These are standard sanity tests that applies to training images. It should be moved to sanity test instead of a tensorflow unit test.

There are probably caveats that are tensorflow specific tests too but we can discuss about that.

You should research the current dockerfiles, compile results on what tests are generic and should be moved to sanity. Then we can discuss on tests that are specific to TF.

test/tensorflow/unit/test_filesystem.py
sirutBuasai, line 1: Same here, research to see which tests are generic and should be in sanity

test/tensorflow/unit/test_ssh_config.py
sirutBuasai, line 1: This whole test is a generic sanity test rather than TF specific.

test/tensorflow/unit/test_smoke.py
```diff
@@ -0,0 +1,77 @@
...
+def test_tensorflow_built_with_cuda():
+    """TF wheel must be a CUDA-capable build. This is a build-time fact;
+    works even when the test runner has no physical GPU."""
+    import tensorflow as tf
+
+    assert tf.test.is_built_with_cuda(), "TF wheel was not compiled with CUDA support"
+
+
+@cuda_only
+def test_cudart_loadable():
```
sirutBuasai: these are generic cuda sanity tests that should apply to all frameworks.

sirutBuasai: See if such test don't already exist on sanity directory, if not, we should add them to sanity.

.github/config/image/tensorflow/2.21-sagemaker-cuda.yml
bhanutejagk: Sure will update

.github/config/image/tensorflow/2.21-sagemaker-cuda.yml
bhanutejagk: Sounds good.

.github/workflows/tensorflow.pipeline.yml
bhanutejagk: oh yeah missed it. Thanks for pointing. Will push the fix.

test/tensorflow/pytest.ini
bhanutejagk: sure

docker/tensorflow/2.21/Dockerfile.cuda
bhanutejagk: I actually copied the bake block from [docker/pytorch/2.11/Dockerfile.cuda](https://github.com/aws/deep-learning-containers/blob/main/docker/pytorch/2.11/Dockerfile.cuda#L176-L187) — PT does this too. I thought it is due to the reason that binary is useful to customers running multi-node training who want to verify EFA/NCCL connectivity before launching real workloads. Also the size of the binary which bloats the image is comparably small to the whole image size. 

The alternative (which SGLang/vLLM use) is to just enable the `cuda-rhel9.repo` in the Dockerfile so `setup_nccl_tests.sh`'s test-time `dnf install libnccl-devel` fallback works — that was the [actual failure](https://github.com/aws/deep-learning-containers/pull/6107/changes/5e6f85e0537aa5ef267a062289f40b5c6cd592ba) that pushed me to bake in the first place. 

Which way do you think is better 
(a) keep the bake but pin the nccl-tests ref
(b) revert the bake and add the `cuda-rhel9.repo` enable (SGLang pattern). 

If (b) then we should consider changing in pytorch too i guess.

docker/tensorflow/2.21/Dockerfile.cuda
bhanutejagk: It's copied from [TF 2.19 in master](https://github.com/aws/deep-learning-containers/blob/master/tensorflow/training/docker/2.19/py3/cu125/Dockerfile.gpu#L384) — the || true came along with it. But you're right that it's hiding real failures. Checked the build log: 67 compile errors, all in Boost.Python (`fatal error: pyconfig.h: No such file or directory` — uv venv doesn't ship pyconfig.h). The other 38 Boost sub-libraries install cleanly, which is what SM Studio actually links against. Cleaner fix is using `--without-python` on b2 so it skips Boost.Python explicitly and returns 0 without needing the mask. I will work on updating it since the legacy code is too old unless you suggest any further improvements.

test/tensorflow/integration/sagemaker/test_mnist_cuda.py
bhanutejagk: Yeah i will work on creating conftest.py file

test/tensorflow/integration/sagemaker/test_mnist_cuda.py
bhanutejagk: sure working on it.

test/tensorflow/unit/test_environment.py
bhanutejagk: addressed in last comment for discussion.

test/tensorflow/unit/test_filesystem.py
bhanutejagk: addressed in the last comment for discussion.

test/tensorflow/unit/test_ssh_config.py
bhanutejagk: addressed in last comment for discussion.

test/tensorflow/unit/test_smoke.py
bhanutejagk: Compiled the full classification across all 4 files in my PR (test_environment, test_filesystem, test_ssh_config, test_smoke). 34 of 48 tests are framework-agnostic and belong in sanity. 6 are TF-specific and stay in unit. test_ssh_config.py is 100% generic — whole file moves. Full classification:

Keep in test/tensorflow/unit/ (TF-specific):
 - test_environment.py: test_kmp_affinity, test_kmp_blocktime (Intel MKL env for TF's oneDNN/Eigen backend)
 - test_filesystem.py: test_venv_has_tensorflow
 - test_smoke.py: test_tensorflow_matmul_runs, test_cpu_devices_detected, test_tensorflow_built_with_cuda

Move to test/sanity/ (34 tests across categories):
 - Container env (DLC_CONTAINER_TYPE, PYTHON* env, LANG)
 - PATH / LD_LIBRARY_PATH contents (venv, openmpi, efa, cuda paths)
 - Binaries (python, mpirun, sshd, fi_info, nvcc)
 - SSH config (whole test_ssh_config.py file)
 - SageMaker paths (/opt/ml/*)
 - OpenMPI wrapper correctness (double-wrap guard)
 - EFA + NCCL config (ofi-nccl plugin, /etc/nccl.conf)
 - CUDA runtime + cuDNN loadability
 - CPU image cuda-absence guard
 - OSS license files
 - Entrypoint executable

Proposed approach: 
New `test/sanity/scripts/test_sanity_training.py` mirroring `test_sanity_vllm_sglang.py` pattern, gated on framework in [pytorch_runtime, tensorflow, xgboost, ...] and per-device/customer flags. Wired into `_reusable.sanity-tests.yml` as a new step. 


Since this touches cross-framework surface (new sanity file + workflow step + PT/XGBoost autoreleases pick it up), so I'd prefer to make it as a follow-up PR. Need your thoughts on this.

docker/tensorflow/2.21/Dockerfile.cuda
sirutBuasai: I'd rather keep the image size small. Lets make a note so that after this one merge, you can make another PR to remove it from PyTorch too.

test/tensorflow/single_gpu/test_data_pipeline.py
bhanutejagk: Yeah honestly this test is a historic residue. This was meant to be a copy of PT's test_dataloader_multiworker, but what PT's version actually validates is — `pin_memory=True + assert x.is_pinned()` catches `/dev/shm-too-small` regressions in containers. When it is ported it to TF it dropped both those bits: `pin_memory` because tf.data has no equivalent concept (TF handles GPU transfers internally), and `is_pinned()` for the same reason (nothing to assert against). Going to just delete the file.

docker/tensorflow/2.21/cuda/pyproject.toml
bhanutejagk: sure will upgarde it.

test/tensorflow/integration/sagemaker/test_experiments_cpu.py
bhanutejagk: yeah we can use it too. Updating the logic

test/tensorflow/unit/test_smoke.py
sirutBuasai: the workflow triggers are designed to be specific, so it wont trigger a full workflow run from other frameworks but only sanity test.

but if you think you would want to make another PR. Then comment on this PR the list of things that you are going to do for a follow up PR.

bhanutejagk: Ok then it don't make sense for having an another PR. Will fallback in this same PR.

docker/tensorflow/2.21/Dockerfile.cuda
bhanutejagk: Sure then will proceed with doin this changes.

scripts/docker/common/install_efa_amzn2023.sh
```diff
@@ -33,19 +33,23 @@ function install_efa {
...
-            ldconfig
-        fi
+
+    # EFA 1.48+ auto-detects NGC containers via /opt/nvidia/nvidia_entrypoint.sh
+    # (present in nvidia/cuda:*-amzn2023 bases) and then skips the AL2023
+    # libnccl-ofi RPM. Force the standard install path so the OFI-NCCL plugin
+    # is installed and libnccl-net-ofi.so ends up at /opt/amazon/ofi-nccl/lib64/.
+    # --disable-ngc was added in EFA 1.48; guard so 1.47 callers still work.
+    # sort -V (GNU coreutils) provides deterministic semver comparison —
+    # avoids bash string-order pitfalls like "1.9" > "1.10" being true.
```
sirutBuasai: Lets shrink the comment to 
```
# EFA 1.48+ auto-detects NGC containers via /opt/nvidia/nvidia_entrypoint.sh
# (present in nvidia/cuda:*-amzn2023 bases) and then skips the AL2023
# libnccl-ofi RPM.
```

We're wasting code space + context window when using agent if we overly comment

test/tensorflow/integration/sagemaker/conftest.py
```diff
@@ -0,0 +1,106 @@
...
+def source_dir(resource_dir):
+    return os.path.join(resource_dir, "scripts")
+
+
+# ── Environment / config fixtures ──────────────────────────────────────────
+
+
+@pytest.fixture(scope="session")
+def image_uri():
+    return os.environ["TEST_IMAGE_URI"]
```
sirutBuasai: We already have `image-uri` as input in the `test/conftest.py` already, we dont need this. Use the input arg instead for test script

test/tensorflow/integration/sagemaker/conftest.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+@pytest.fixture(scope="session")
+def default_region():
+    return "us-west-2"
```
sirutBuasai: same here. Already declared in `test/conftest.py`

bhanutejagk: okay

test/tensorflow/integration/sagemaker/conftest.py
bhanutejagk: sounds good.

scripts/docker/common/install_efa_amzn2023.sh
bhanutejagk: sure.

sirutBuasai (approved):

PR #6113, opened 2026-05-19
test(vllm): add EFA test for vLLM Ubuntu (NCCL + NIXL)

Yadan-Wei:
## Summary
- Wires the existing 2x p4d EFA fixture into `pr-vllm-ec2.yml` so vLLM Ubuntu images get NCCL-over-EFA validation on every build.
- Adds two NIXL EFA checks gated on `RUN_NIXL_TESTS=1` (vLLM only): a libfabric plugin packaging smoke, and a multi-node disaggregated prefill/decode test that **strictly verifies KV cache transfer over libfabric/EFA** via NixlConnector Prometheus metrics.

## Why
PyTorch DLCs already have `test/efa/test_efa.py` running against every image. vLLM ships the same EFA stack (installer, OpenMPI, aws-ofi-nccl) but never had its own EFA test. NIXL with the libfabric backend is the path vLLM uses for disaggregated PD over EFA — it deserves coverage too.

## What changed

**Reuses the existing PyTorch EFA fixture, doesn't duplicate it.**

| File | Change |
|---|---|
| `test/efa/test_efa.py` | Adds optional NIXL smoke + disaggregated PD steps after the existing NCCL test, gated on `RUN_NIXL_TESTS=1`. Wraps each step in a verbose `_step` helper that prints stdout/stderr/exit, and dumps prefill/proxy/decode logs even on failure (try/finally). |
| `test/efa/scripts/nccl_allreduce.sh` | Defaults `CUDA_HOME=/usr/local/cuda` instead of erroring out (PyTorch sets it; vLLM doesn't). |
| `test/efa/scripts/setup_nccl_tests.sh` (new) | No-op on PyTorch DLCs. On vLLM Ubuntu, `apt install libnccl-dev` if missing and compiles `all_reduce_perf` against `NCCL_HOME=/usr`. |
| `test/efa/scripts/nixl_libfabric_smoke.py` (new) | Imports `nixl`, instantiates an agent, calls `create_backend("LIBFABRIC", {"provider": "efa"})`. Catches packaging regressions on the `nixl-cu*` wheel. |
| `test/efa/scripts/nixl_disagg_pd.sh` (new) | Master orchestrator. Launches prefill `vllm serve` with `NixlConnector + LIBFABRIC + kv_load_failure_policy:fail`, polls worker decode server, launches proxy, sends a `/v1/completions` request, then **scrapes decode's `/metrics` and asserts `vllm:nixl_xfer_time_seconds_count >= 1` and `vllm:nixl_num_failed_transfers == 0`** — directly proves at least one KV cache block crossed prefill→decode over libfabric/EFA. |
| `test/efa/scripts/nixl_disagg_pd_decode.sh` (new) | Daemonized worker decode endpoint. Uses `kv_role:"kv_both"` (NixlConnector ignores `kv_role`; per-request `kv_transfer_params` does the routing) with matching `--block-size 128` and `VLLM_KV_CACHE_LAYOUT=HND` env var. |
| `test/efa/scripts/toy_proxy_server.py` (new) | Minimal FastAPI proxy doing the upstream-style 2-step handshake: inject `kv_transfer_params={do_remote_decode:True,...}` into prefill body with `max_tokens=1`, extract populated `remote_engine_id/block_ids/host/port/request_id` from prefill response, forward those into decode's request body. ~110 lines, modeled after upstream's `tests/v1/kv_connector/nixl_integration/toy_proxy_server.py`. |
| `.github/workflows/reusable-efa-tests.yml` | New `run-nixl-tests` boolean input; passes through to pytest as `RUN_NIXL_TESTS`. Per-PR concurrency group keyed on workflow + PR number. |
| `.github/workflows/pr-vllm-ec2.yml` | Adds `efa-test` job (gated on `build-image + sanity-test + security-test`) calling the existing reusable workflow with `run-nixl-tests: true`. |

## Verified on real p4d (CI run [26269852212](https://github.com/aws/deep-learning-containers/actions/runs/26269852212/job/77320828507))
- NCCL all_reduce_perf across 2x p4d.24xlarge: **23 GB/s busbw at 1 GiB**, EFA provider selected (4 NICs, GDRDMA on all 4 channels).
- NIXL libfabric smoke: `Backend LIBFABRIC was instantiated`.
- NIXL disagg PD: prefill→decode completion returns coherent text, NixlConnector emits `KV Transfer metrics: Num successful transfers=1, Avg MB per transfer=4.5, Throughput (MB/s)=1003`, and `vllm:nixl_xfer_time_seconds_count` confirms at least one transfer arrived at decode.

## Test plan
- [x] `pr-vllm-ec2` `efa-test` job runs against the new vLLM image and is green.
- [x] PyTorch CI path (`pr-pytorch-ec2-cuda`) still passes — `setup_nccl_tests.sh` short-circuits on the preinstalled binary; `RUN_NIXL_TESTS` defaults to `0`.
- [x] Both EFA tests share a per-PR concurrency group so PyTorch and vLLM EFA runs don't fight for p4d capacity within the same PR.
- [x] Failure modes are loud: `kv_load_failure_policy:fail` makes NIXL transport errors return 5xx (instead of silently re-prefilling); the `vllm:nixl_*` metric assertions catch a missing transfer even when the completion text is coherent.

## Notes
- NixlConnector's `kv_role` is advisory at the engine level in vLLM 0.21.0 — what actually triggers the remote-pull path is the per-request `kv_transfer_params` dict the proxy injects. Both servers run with `kv_both` matching upstream's `tests/v1/kv_connector/nixl_integration/run_accuracy_test.sh`.
- The proof metric is **not** `vllm:prompt_tokens_total` (that always counts the full prompt, regardless of whether KV came from cache or recompute) — it's `vllm:nixl_xfer_time_seconds_count`, the histogram counter NixlConnector increments per successful transfer.


.github/scripts/efa/ec2_helpers.py
```diff
@@ -275,16 +275,34 @@ def launch_efa_instances(aws_session, ami_id, instance_type, key_name, sg_id, co
...
+      forward-compat then `exec "$@"`): keep entrypoint, pass `bash` so the
+      compat env var lands in the environment that NCCL+EFA later inherit.
+      Overriding it breaks NCCL on hosts with older CUDA drivers.
+    """
     devices = get_efa_devices(conn)
     device_args = " ".join(f"--device {d}" for d in devices)
 
+    if "vllm" in image_uri.lower():
+        entrypoint_arg = "--entrypoint /bin/bash"
+        cmd = "-c 'sleep infinity'"
```
sirutBuasai: why do need need this `-c sleep infinity`?

Also, ideally we don't want to have the `if <framework>` statements because its an antipattern. We've moved to v2 and revamped the release logic for this very reason. We've also already introduced this anti pattern in the new release logic code so we're still in the process of cleaning up this antipattern too.

Yadan-Wei: Good catch on the antipattern — fixed in 85550e9f. setup_container and efa_instances now take explicit entrypoint / cmd kwargs (defaults work for PyTorch, no URI sniffing), and the vLLM caller in pr-vllm-ec2.yml passes the overrides through new container-entrypoint / container-cmd inputs on the reusable workflow.

On -c 'sleep infinity': the container is launched with -id and needs to stay alive long enough for the test to docker exec into it. We override the entrypoint to /bin/bash for vLLM because its baked-in dockerd_entrypoint.sh execs vllm serve "$@" and would parse a generic CMD (e.g., bash) as a model tag. But once --entrypoint /bin/bash is set, the CMD becomes argv to bash — passing nothing would inherit the image's CMD, and passing bash would make bash try to execute a script named bash. -c 'sleep infinity' is the explicit "do nothing forever" payload. PyTorch doesn't need any of this because its entrypoint.sh ends in exec "$@", so passing cmd="bash" runs bash directly with stdin held open by -id.

sirutBuasai (approved):

PR #6151, opened 2026-05-27
test(vllm): add Qwen3-Reranker-4B smoke and benchmark tests

Yadan-Wei:
## Summary
- Add `vllm_reranker_smoke_test.sh` and `vllm_reranker_benchmark_test.sh` for Qwen3-Reranker-4B on the AL2023 vLLM image; both use Path C — load the model with `--runner pooling` and `--hf_overrides` (`Qwen3ForSequenceClassification` + `classifier_from_token: ["no","yes"]`) so the standard Cohere-compatible `/v1/rerank` endpoint registers, then exercise it end-to-end.
- Wire the model into `.github/config/model-tests/vllm-model-tests.yml` for both smoke-test and benchmark on `x86-g6xl-runner` (1× L4, ~8 GB BF16 weights fit comfortably).
- Model artifact uploaded to `s3://dlc-cicd-models/llm-models/qwen3-reranker-4b.tar.gz` (6.0 GiB, flat layout).

The `--hf_overrides` JSON and jinja chat template are baked into the scripts (not the YAML `extra_args`) to avoid YAML quoting fragility.

## Test plan
- [ ] Smoke job `test-model (qwen3-reranker-4b)` passes on PR run — asserts top result is the relevant doc, descending score order, relevant-vs-irrelevant margin > 0.1, single-doc request works.
- [ ] Benchmark job `benchmark (qwen3-reranker-4b)` passes baseline profile (`min_rps: 1`) on the dispatch workflow.
- [ ] Confirm baseline rps/latency from first benchmark run is reasonable, then tune `min_rps` if needed.

### DIFF
--- a/.github/config/model-tests/vllm-model-tests.yml
+++ b/.github/config/model-tests/vllm-model-tests.yml
@@ -29,6 +29,12 @@ smoke-test:
       extra_args: "--runner pooling --dtype bfloat16 --max-model-len 8192 --trust-remote-code"
       test_script: "vllm_embedding_smoke_test.sh"
 
+    - name: "qwen3-reranker-4b"
+      s3_model: "qwen3-reranker-4b.tar.gz"
+      fleet: "x86-g6xl-runner"
+      extra_args: "--dtype bfloat16 --max-model-len 10000 --gpu-memory-utilization 0.85"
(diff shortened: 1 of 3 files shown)


Yadan-Wei:
Re-opening from non-fork branch — CI requires same-repo PRs.

PR #6152, opened 2026-05-27
test(vllm): add Qwen3-Reranker-4B smoke and benchmark tests

Yadan-Wei:
## Summary

Add Qwen3-Reranker-4B end-to-end test coverage on the AL2023 vLLM image (EC2 smoke + benchmark + SageMaker endpoint), plus a handful of unrelated CI fixes that surfaced while iterating.

### Reranker test coverage
- New EC2 smoke test (`vllm_reranker_smoke_test.sh`) and benchmark (`vllm_reranker_benchmark_test.sh`) under `test/vllm/scripts/`. Both load the model via Path C — `--runner pooling` + `--hf_overrides {"architectures":["Qwen3ForSequenceClassification"],"classifier_from_token":["no","yes"],"is_original_qwen3_reranker":true}` — so the standard Cohere-compatible `/v1/rerank` endpoint registers, then exercise it end-to-end.
- New SageMaker endpoint test entry in `vllm-model-tests.yml` `sagemaker:` section, gated by `required_image_pattern: amzn2023`. Runs on `ml.g6.xlarge` (24 GB L4; same VRAM tier as g5.xlarge with better us-west-2 capacity).
- Wire-up in `.github/config/model-tests/vllm-model-tests.yml` for smoke + benchmark on `x86-g6xl-runner` (1× L4, ~8 GB BF16 weights fit comfortably).
- Model artifact at `s3://dlc-cicd-models/llm-models/qwen3-reranker-4b.tar.gz` (6.0 GiB).

### Shared chat template
- New `test/vllm/scripts/amzn2023/qwen3_reranker_chat_template.jinja` is the single source of truth.
- EC2: copied into the container alongside other test scripts via the existing `docker cp` step in `reusable-vllm-model-tests.yml` and `dispatch-vllm-benchmark.yml`. Scripts read it from `/models/qwen3_reranker_chat_template.jinja`.
- SageMaker: read at test runtime by `test_sm_model_serving.py`, flattened (newlines → `{{ "\n" }}`), shell-quoted, and injected as `SM_VLLM_CHAT_TEMPLATE`.
- The SM test runner is moved to `test/vllm/sagemaker/amzn2023/test_sm_model_serving.py` and the Ubuntu PR workflow path filters are scoped to top-level files only, so amzn2023-only YAML changes no longer trigger Ubuntu builds.

### SM-specific quoting workarounds
The vLLM AMZN2023 SageMaker container runs vLLM under `standard-supervisor` (from `model-hosting-container-standards`). It joins argv with single spaces into one string, then supervisord re-parses via `shlex.split`, which strips unprotected double quotes and breaks tokens on whitespace. Two SM env values needed shell-quote-protective wrapping:
- `SM_VLLM_HF_OVERRIDES` — JSON literal wrapped in single quotes in the YAML value.
- `SM_VLLM_CHAT_TEMPLATE` — `_flatten_jinja()` switched to double-quoted Jinja `\n` expressions and outer single-quote-wraps the whole template at injection time.
Upstream fix would be `shlex.join()` (or `shlex.quote()` in the join) at `model_hosting_container_standards/supervisor/scripts/standard_supervisor.py:198`.

### Adjacent fixes (rolled in to keep CI green)
- **`transformers<5.10` pin** in `docker/vllm/Dockerfile.amzn2023`. transformers 5.10.0 added an unguarded `prepare_inputs_layout → fetch_audio` call that breaks voxtral with the `MistralCommonFeatureExtractor` shipped in mistral-common 1.11.2. Voxtral was passing on `e7e19d66` (last build before transformers 5.10.x got resolved into the image) and failing after. Upstream issue: vllm-project/vllm#44554 — drop the pin once that is fixed.
- **CVE allowlist updates** for the existing mooncake `libetcd_wrapper.so` static-link issue:
  - `CVE-2026-39821` (golang.org/x/net) added to `vllm/` and `vllm_server/`.
  - `CVE-2026-42504` (go/stdlib) added to `vllm/` and `vllm_server/`.
  - Removed a stale duplicate `CVE-2026-39821` v0.38.0 entry in `vllm/`.
- **`/tmp` cleanup** in `reusable-vllm-upstream-tests.yml`. The `x86-g6xl-runner` CodeBuild fleet has a 7.6 GiB tmpfs that isn't always purged between jobs; earlier ray-ec2 jobs leak multi-GB `/tmp/ray-ec2-*` dirs, which crashed our cuda-test at "Checkout vLLM tests" with ENOSPC. Added a per-job step that removes known leakage patterns before the heavy steps.
- **Ubuntu PR workflow path filters** — replaced `test/vllm/scripts/**` + `!`-negations (which dorny/paths-filter handles unreliably) with explicit enumeration of top-level scripts only. amzn2023-only changes (under `test/vllm/scripts/amzn2023/`, `test/vllm/scripts/benchmark/`, `test/vllm/sagemaker/amzn2023/`) no longer trigger Ubuntu vllm builds.

## Test plan
- [x] EC2 smoke `test-model (qwen3-reranker-4b)` on `PR - vLLM EC2 AMZN2023` — asserts top result is the relevant doc, descending score order, relevant-vs-irrelevant margin > 0.1, single-doc request works.
- [ ] EC2 benchmark `benchmark (qwen3-reranker-4b)` on the dispatch workflow with `min_rps: 20`.
- [ ] SM endpoint `test-model (qwen3-reranker-4b)` on `PR - vLLM SageMaker AMZN2023` — deploys to `ml.g6.xlarge`, hits `/v1/rerank`, validates `results[0].relevance_score`.
- [x] `voxtral-mini-4b` smoke test passes on `PR - vLLM EC2 AMZN2023` (transformers pin verified).
- [x] `security-test / ecr-vulnerability-scan` passes on `PR - vLLM SageMaker AMZN2023` (CVE-2026-42504 allowlisted).
- [x] Ubuntu workflows (`PR - vLLM EC2`, `PR - vLLM SageMaker`) skip when only amzn2023-only paths change.


docker/vllm/Dockerfile.amzn2023
```diff
@@ -193,6 +193,12 @@ RUN --mount=type=cache,target=/root/.cache/uv ls /tmp/vllm-dist/*.whl \
   && uv pip install "${VLLM_WHL}" \
     --extra-index-url ${PYTORCH_CUDA_INDEX_BASE_URL}/cu$(echo $CUDA_VERSION | cut -d. -f1,2 | tr -d '.')
 
+# Pin transformers to <5.10 to avoid the prepare_inputs_layout->fetch_audio
+# AttributeError on voxtral with vllm 0.22.1rc0 + mistral-common 1.11.2.
+# See https://github.com/huggingface/transformers/blob/v5.10.1/src/transformers/processing_utils.py#L726
+# (call lacks hasattr guard; MistralCommonFeatureExtractor doesn't implement fetch_audio).
+RUN --mount=type=cache,target=/root/.cache/uv uv pip install "transformers<5.10"
```
Jyothirmaikottu, line 200: Is there a reason why we pinned here in the dockerfile? Any future model that requires transformers 5.10+ features will be blocked until this pin is removed. Since voxtral is the only model affected, and its S3 tarball (voxtral-mini-4b.tar.gz) already has a requirements.txt at the model root, you can scope the pin there instead.

Yadan-Wei: But our target any release shouldn't break current model for customers, any change on their side means a break?

Yadan-Wei: Also, I have a upstream PR to fix this, when vllm has a new release, we can revert this.

Jyothirmaikottu (approved):

PR #6182, opened 2026-06-04
docs: add Text Embeddings Inference (TEI) documentation

dwarez:
## Purpose
Add documentation for the Hugging Face Text Embeddings Inference (TEI) DLCs so users can find which image to use to serve embedding, reranker, and sequence-classification models on Amazon SageMaker AI:

## PR Checklist

### DIFF
--- a/docs/src/global.yml
+++ b/docs/src/global.yml
@@ -88,7 +88,7 @@ display_names:
     huggingface-pytorch-inference: "HuggingFace PyTorch Inference"
     huggingface-vllm: "HuggingFace vLLM Inference"
     huggingface-sglang: "HuggingFace SGLang Inference"
-    huggingface-tei: "HuggingFace TEI Inference"
+    huggingface-tei: "HuggingFace Text Embeddings Inference"
     huggingface-vllm-inference-neuronx: "HuggingFace vLLM Inference NeuronX"
     huggingface-pytorch-training-neuronx: "HuggingFace PyTorch Training NeuronX"
     huggingface-pytorch-inference-neuronx: "HuggingFace PyTorch Inference NeuronX"
--- a/docs/src/global.yml
+++ b/docs/src/global.yml
@@ -88,6 +88,7 @@ display_names:
     huggingface-pytorch-inference: "HuggingFace PyTorch Inference"
     huggingface-vllm: "HuggingFace vLLM Inference"
     huggingface-sglang: "HuggingFace SGLang Inference"
+    huggingface-tei: "HuggingFace TEI Inference"
     huggingface-vllm-inference-neuronx: "HuggingFace vLLM Inference NeuronX"
     huggingface-pytorch-training-neuronx: "HuggingFace PyTorch Training NeuronX"
     huggingface-pytorch-inference-neuronx: "HuggingFace PyTorch Inference NeuronX"
@@ -200,6 +201,7 @@ table_order:
   - huggingface-pytorch-inference
   - huggingface-vllm
   - huggingface-sglang
+  - huggingface-tei
   - huggingface-vllm-inference-neuronx
   - huggingface-pytorch-training-neuronx
   - huggingface-pytorch-inference-neuronx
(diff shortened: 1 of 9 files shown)


docs/src/global.yml
```diff
@@ -88,6 +88,7 @@ display_names:
     huggingface-pytorch-inference: "HuggingFace PyTorch Inference"
     huggingface-vllm: "HuggingFace vLLM Inference"
     huggingface-sglang: "HuggingFace SGLang Inference"
+    huggingface-tei: "HuggingFace TEI Inference"
```
alvarobartt:
```suggestion
    huggingface-tei: "HuggingFace Text Embeddings Inference"
```

docs/src/tables/huggingface-tei.yml
```diff
@@ -0,0 +1,14 @@
+# Table Configuration - HuggingFace TEI
+columns:
+  - field: framework_version
+    header: "Framework"
+  - field: python
+    header: "Python"
```
alvarobartt: FYI though the image is labelled with Python it really doesn't come with Python installed, so this might be a nice opportunity to remove those (though it might impact tagging, but we should soon resolve those as it's misleading cc @Jyothirmaikottu)

docs/tei/deployment/sagemaker.md
```diff
@@ -0,0 +1,187 @@
+# Amazon SageMaker AI Deployment
+
+Deploy TEI on Amazon SageMaker AI endpoints. The container accepts model configuration via environment variables and serves on port 8080.
```
alvarobartt:
```suggestion
Deploy Text Embeddings Inference (TEI) on Amazon SageMaker AI endpoints. The container accepts model configuration via environment variables and serves on port 8080.
```

docs/tei/deployment/sagemaker.md
```diff
@@ -0,0 +1,187 @@
...
+
+## SageMaker Python SDK v2
+
+```python
+from sagemaker.huggingface import HuggingFaceModel, get_huggingface_llm_image_uri
+
+model = HuggingFaceModel(
+    # Resolves the TEI GPU image for your session's region.
+    # For CPU instances, use the "huggingface-tei-cpu" backend instead.
+    image_uri=get_huggingface_llm_image_uri("huggingface-tei", version="1.8.2"),
```
alvarobartt, line 22: FYI we're soon releasing 1.9.3, see https://github.com/awslabs/llm-hosting-container/pull/192

alvarobartt (commented):
Thanks @DWarez! I'd say that we name it "Text Embeddings Inference" instead of "TEI" is the first occurrences on each document, as users won't be as familiar with TEI acronym as they are with SGLang or vLLM, so using the complete name provides the information that it's an engine for embeddings IMO 🤗

docs/tei/deployment/sagemaker.md
dwarez: should I specify this version such that the example will already be up-to-date when the release happens? Even tho it could break the example if the docs go live before the new release

docs/src/tables/huggingface-tei.yml
dwarez: I removed the field but left it in the tag/examples

docs/tei/deployment/sagemaker.md
alvarobartt: It depends on when this PR is merged, but IMO if they merge as-is it's fine, we can update in a follow-up PR, maybe @Jyothirmaikottu knows about the current issues with testing on https://github.com/awslabs/llm-hosting-container/pull/189 and maybe you can share an ETA on when that will be released? Thanks 🙏🏻

Jyothirmaikottu: @alvarobartt Let's not include the 1.9.3 version in the documentation as its still in the build/test process. It's a good idea to keep that as a separate PR. I still need to debug why the SM endpoint is not spinning up as expected.

jinyan-li1 (dismissed):

dwarez:
closing this one in favor of #6189

PR #6196, opened 2026-06-05
fix(vllm-omni): switch SageMaker endpoint tests to g6 to avoid ICE

Yadan-Wei:
## Summary

- vllm-omni SageMaker smoke tests have been failing with `InsufficientInstanceCapacity` on `ml.g5.xlarge` / `ml.g5.2xlarge` in us-west-2.
- Switch sync TTS and async TTS to `ml.g6.xlarge`; switch video async to `ml.g6.2xlarge`. Same 24 GB VRAM / 32 GB host RAM class on L4 (sm_89), which uses the same PyTorch SDPA fallback path as A10G for these models.
- Add `timeout=1800` to `endpoint.wait_for_status("InService")` so a stuck-in-`Creating` endpoint surfaces as `TimeoutExceededError` instead of waiting indefinitely.

The async-fixture's existing `_CAPACITY_TOKENS` skip remains in place as a safety net if g6 also runs out.

## Test plan

- [ ] `test_vllm_omni_tts_endpoint` deploys on `ml.g6.xlarge` and TTS round-trips
- [ ] `test_vllm_omni_tts_async_endpoint` deploys on `ml.g6.xlarge` and async TTS round-trips
- [ ] `test_vllm_omni_video_async_endpoint` deploys on `ml.g6.2xlarge` and VACE video round-trips
- [ ] No new ICE failures in CI matrix

### DIFF
--- a/test/vllm-omni/sagemaker/test_sm_omni_endpoint.py
+++ b/test/vllm-omni/sagemaker/test_sm_omni_endpoint.py
@@ -87,14 +87,14 @@ def model_endpoint(aws_session, image_uri, model_id, instance_type):
             endpoint_name=endpoint_name,
             endpoint_config_name=endpoint_name,
         )
-        endpoint.wait_for_status("InService")
+        endpoint.wait_for_status("InService", timeout=1800)
 
         yield endpoint
     finally:
         _cleanup([endpoint, endpoint_config, model])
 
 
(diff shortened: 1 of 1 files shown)


Yadan-Wei:
Recreating from origin branch instead of fork.

PR #6215, opened 2026-06-09
Base: CUDA-version tagging + matrix-driven workflows

Eren-Jeager123:
## Summary

Two coupled changes to the **base** DLC image line:

1. **New tag schema** — switch from generation-number tags (`runtime-cuda-v1` / `-v2`) to **CUDA-version-encoded** tags.
2. **Workflow refactor** — replace the per-generation `pr-base-v1/v2` and the hardcoded autorelease blocks with **per-target** workflows that follow the PyTorch/sglang patterns.

> The `base_dlc` tag template lives in the separately-published release-logic component (consumed by the release job). That change is deployed, and a gamma release was run from this branch to validate the end-to-end path (`runtime-cu130-amzn2023` / `devel-cu130-amzn2023` published to gamma successfully).

## 1. Tag schema

| Variant | Mutable tag | Immutable tag |
|---|---|---|
| runtime, CUDA 12.9 | `runtime-cu129-amzn2023` | `runtime-cu129-amzn2023-<date>` |
| runtime, CUDA 13.0 | `runtime-cu130-amzn2023` | `runtime-cu130-amzn2023-<date>` |
| devel, CUDA 12.9 | `devel-cu129-amzn2023` | `devel-cu129-amzn2023-<date>` |
| devel, CUDA 13.0 | `devel-cu130-amzn2023` | `devel-cu130-amzn2023-<date>` |

**Why:** `runtime-cuda-v1` and `runtime-cuda-v2` were both produced by the `-vMAJOR` version variant from two *different* CUDA images (12.9 vs 13.0), so the shared `runtime-cuda` family could resolve to different digests across releases. Encoding the CUDA version directly into the tag gives each image a distinct, stable mutable tag.

Mechanically: `container_type` changes `runtime-cuda`/`devel-cuda` → `runtime`/`devel`, and the release tag template becomes `{container_type}-{cuda_version}-{os_version}` with a dated immutable variant.

## 2. Rename v1/v2 → cu129/cu130

The `v1`/`v2` naming encoded the CUDA generation; with CUDA-versioned tags it's redundant. Renamed across:
- `docker/base/v1` → `docker/base/cu129`, `v2` → `cu130` (+ Dockerfile COPY paths, `pyproject.toml` / `uv.lock` project names)
- 4 image configs `base-v{1,2}-{runtime,devel}.yml` → `base-cu{129,130}-{runtime,devel}.yml`
- docs (`docs/base/index.md`, changelog, `docs/src/data/base/*`)

## 3. Dockerfile + config build pins

- Each Dockerfile takes `ARG CUDA_VERSION` (full, e.g. `13.0.2`, for `FROM nvidia/cuda:${CUDA_VERSION}-*`), `ARG CUDA_VERSION_SHORT` (e.g. `cu130`, for the deps COPY path), and `ARG UV_VERSION` (via a `FROM ghcr.io/astral-sh/uv:${UV_VERSION} AS uv` stage).
- Build pins live in a config `build:` block (`python_version`, `uv_version`), following the sglang amzn2023 pattern — single source of truth, no env file under `docker/`. `pyproject.toml` / `uv.lock` remain as files (hash-locked lockfile, not scalar pins).

## 4. Workflows — per-target, matrix + reusable building blocks

PyTorch splits workflows along its image axis (ec2 / sagemaker / cpu / cuda); base's equivalent axis is the build target, so workflows are split **per target** — runtime and devel build, test, and release **independently**:

| | runtime | devel |
|---|---|---|
| Auto release | `autorelease-base-runtime.yml` | `autorelease-base-devel.yml` |
| PR | `pr-base-runtime.yml` | `pr-base-devel.yml` |

**Auto release:** one cron per CUDA version, each run releasing a single image. Staggered so no two collide — runtime cu129 12:40 / cu130 1:40; devel cu129 12:50 / cu130 1:50 (PST Mon/Wed). `load-config` resolves the CUDA version (cron → version `case`, or `workflow_dispatch` input) and validates the config exists; pipeline is `load-config → build-image → {sanity, base-sanity, cuda} → generate-release-spec → release-image`. `generate-release-spec` gates on `should-release` (the config's `release.release`).

**PR:** `check-changes` (with `fetch-depth: 0`) detects which CUDA versions the PR touches and matrixes over CUDA version only (target fixed per workflow); falls back to the latest version discovered on disk for shared test-only changes.

**Build:** all four workflows build via the shared `./.github/actions/build-image` action (as sglang/vllm do) instead of an inline `docker buildx build` — removing duplicated build logic. Config build-args (`CUDA_VERSION` full, `CUDA_VERSION_SHORT`, plus the `build:` keys) are passed through `EXTRA_BUILD_ARGS` so values are forwarded space-safely.

Adding a future CUDA version: add `docker/base/cu131/` + the two config files + one cron line per autorelease workflow.

## Bug found & fixed during CI

`check-changes` ran `git diff origin/main...HEAD` on a shallow checkout, where `origin/main` is absent — the diff errored, detection fell back to the latest version on every run. Fixed with `fetch-depth: 0`. (`pr-pytorch-*` has the same latent issue; tracked separately.)

## Follow-ups (not in this PR)

- Extract the config→matrix parsing into a tested Python script + composite action (the inline yq/jq is duplicated; a Python script is more testable/debuggable).
- Switch `pr-pytorch-*`/`autorelease-pytorch-*` to the shared `build-image` action — they build inline and share the same word-splitting/duplication issues.
- Decide whether base should run a telemetry test, and what `DLC_CONTAINER_TYPE` should be (`container_type` is currently the release-tag variable, consumed by release logic, not a telemetry/sagemaker signal).
- A global schedule registry / lint to make cron collisions visible across all framework workflows.

## Testing

- `pre-commit`: `check-yaml`, `check-github-workflows` (Actions schema), `typos`, `flowmark`, filesystem/security hooks pass; `actionlint` clean on all four workflows.
- Simulated cron→version resolution, the config→build-args flow (correct full values, space-safe), and matrix building against the real config files.

> The release path was validated end-to-end in a gamma run from this branch (cu130 runtime + devel published to gamma). The temporary PR-trigger scaffolding used for that test has been removed.


.github/workflows/autorelease-base.yml
```diff
@@ -199,177 +229,97 @@ jobs:
...
   # ============================================================
-  # Release — v1 runtime
+  # Generate release specs — after all tests pass, before release
+  # (same placement as pr-pytorch). Runs the generate-release-spec
+  # action once per target via the load-config → generate-release-spec
+  # chain, exposing a runtime spec and a devel spec.
   # ============================================================
-  generate-release-spec-v1-runtime:
-    needs: [load-config-v1-runtime, build-images, sanity-test, base-sanity-test, cuda-test]
+  generate-release-spec:
```
sirutBuasai: the release gate currently checks only the runtime config's `release.release` flag and applies it to both targets. This works today since all configs have `release: true`, but it encodes an implicit assumption. 
Two cases to consider:


1) runtime and devel always release together. If this is intentional, then there are couple ways we can improve this.
    1a) Collapse runtime and devel image config file into single config file. But this will require a change to how the load-config action will work and will require a refactor of said job and potentially all other image config files.
      ```
        common:
          framework: "base_dlc"
          framework_version: "12.9.1"
          cuda_version: "cu129"
          python_version: "py313"
          os_version: "amzn2023"
          # ... shared fields
        targets:
          runtime:
            build_target: "runtime"
            ci_tag: "cuda129-runtime-amzn2023"
            container_type: "runtime"
            prod_image: "base:runtime-cu129-amzn2023"
          devel:
            build_target: "devel"
            ci_tag: "cuda129-devel-amzn2023"
            container_type: "devel"
            prod_image: "base:devel-cu129-amzn2023"
        release:
          release: true          # one flag, unambiguously controls both
          environment: production
      ```
     1b) Add a validation logic step as a pre-release requirement to check that the runtime and devel image release fields are the same. This is a faster and smaller hotfix.
2. runtime and devel images can be released independently. This means we'll need to adapt the release job to the matrix method
    2a) we can create step that does matrix parsing and and the release job will take a matrix input. We already have config in a matrix so this is pretty viable.
    2b) If we expect only base to be just runtime and devel in the future, we can just create two different jobs specifically for runtime and devel. This is somewhat hardcoding and not scalable but if we have the background assumption set, thats not an issue.

.github/workflows/autorelease-base.yml
```diff
@@ -1,99 +1,135 @@
...
-    runs-on: ubuntu-latest
-    outputs:
-      config: ${{ steps.load.outputs.config }}
-    steps:
-      - uses: actions/checkout@v5
-      - id: load
-        uses: ./.github/actions/load-config
-        with:
-          config-file: .github/config/image/base-v2-devel.yml
+      - name: Build configs matrix (runtime + devel for this CUDA version)
```
sirutBuasai: nit: you dont have to fix this right now but honestly if matrix is the future way to do things with other images that will support multiple version, we might as well write a script that does this. A python script is always going to be way more readable than bash. Especially inline workflow bash is going to be really hard to debug if it fails.

docs/base/index.md
```diff
@@ -86,5 +86,5 @@ These images are curated builds:
 
 - **Built from upstream `nvidia/cuda:*-amzn2023` images** — we add Python compiled from source, OSS license metadata, and security patches.
 - **Continuously patched** — security updates from {{ aws }} and NVIDIA are applied on every build.
-- **Versioned** — major version (`v1`, `v2`) tracks the CUDA major. New CUDA majors get a new tag suffix; existing tags stay on their pinned CUDA
-  release across minor patches.
+- **Versioned by CUDA release** — each tag encodes its CUDA version (`cu129`, `cu130`). New CUDA releases get a new tag; existing tags stay on their
```
sirutBuasai, line 89: lets bring this up to with the team. Previously for amzn2023 images, we want to use v1/v2 convention. but this PR changes that to cu129 and cu130. Honestly I do like this tagging more because it shows directly what cuda it's supposed to be. but there is an argument to be made for making the semver for amzn2023 consistent across our DLC.

.github/workflows/autorelease-base.yml
sirutBuasai: If you think converting to a script will be useful, I'd take it one step further and think about what other run jobs that are often reused that can be abstracted into its own actions/scripts. Mainly the super complicated ones should be a script, and reusable ones should be its own actions. If both, then we use both lol.

.github/workflows/autorelease-base.yml
```diff
@@ -1,99 +1,135 @@
...
-    runs-on: ubuntu-latest
-    outputs:
-      config: ${{ steps.load.outputs.config }}
-    steps:
-      - uses: actions/checkout@v5
-      - id: load
-        uses: ./.github/actions/load-config
-        with:
-          config-file: .github/config/image/base-v2-runtime.yml
+      - name: Resolve CUDA version
```
sirutBuasai: if configs are removed or renamed without updating the case, the job fails deep in the matrix-building step with a confusing yq error rather than a clear message at the point of resolution.

you could replace it with something like this
```
  - name: Resolve CUDA version
    id: resolve
    run: |
      if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
        CUDA="${{ inputs.cuda-version }}"
      else
        # Discover available CUDA versions from config files (sorted lexicographically).
        # Works because cu1XX < cu1XY for all realistic CUDA versions (NVIDIA is at 13.x).
        VERSIONS=($(ls .github/config/image/base-cu*-runtime.yml \
          | grep -oE 'cu[0-9]+' | sort -u))

        # Cron-to-slot mapping. Each slot index corresponds to the Nth discovered
        # CUDA version. When adding a new CUDA version: add configs + uncomment/add
        # a cron entry in on.schedule + add a slot entry here.
        # When removing a CUDA version: delete its configs. The remaining versions
        # shift down — update or remove the now-excess cron/slot entries.
        declare -A SLOT_INDEX=(
          ["40 20 * * 1,3"]=0
          ["40 21 * * 1,3"]=1
        )

        IDX="${SLOT_INDEX[${{ github.event.schedule }}]:-}"
        if [ -z "$IDX" ]; then
          echo "::error::Unknown cron '${{ github.event.schedule }}'. Add a SLOT_INDEX entry or remove the cron from on.schedule."
          exit 1
        fi

        if [ "$IDX" -ge "${#VERSIONS[@]}" ]; then
          echo "::error::Slot $IDX fired but only ${#VERSIONS[@]} CUDA version(s) exist (${VERSIONS[*]}). Remove the excess cron entry from on.schedule."
          exit 1
        fi

        CUDA="${VERSIONS[$IDX]}"
      fi

      # Validate resolved version has config files for both targets
      for TARGET in runtime devel; do
        CONFIG=".github/config/image/base-${CUDA}-${TARGET}.yml"
        if [ ! -f "$CONFIG" ]; then
          echo "::error::Resolved CUDA=${CUDA} but missing config: ${CONFIG}. Add the config file or remove this version's slot."
          exit 1
        fi
      done

      echo "Resolved CUDA version: ${CUDA}"
      echo "cuda=${CUDA}" >> "$GITHUB_OUTPUT"
```

Making this a script might be more readble though

.github/config/image/base-cu129-devel.yml
sirutBuasai, line 1: Take a look at current sglang amzn2023 config file. I've moved the env file that are put into docker for versioning into the config file themselves. This is better because we're just maintaining one file instead of multiple files in different places.

.github/workflows/autorelease-base.yml
Eren-Jeager123: I actually plan to pick up 2b, as I think it's similar with pytorch ec2/sagemaker to have 2 separate workflow.

.github/workflows/autorelease-base.yml
Eren-Jeager123: I will add the validation, agree that the yq error will be confused. However I think the sort infex is overcomplex in this case so I did not implement them, lmk if you want to discuss more about this.

.github/workflows/autorelease-base.yml
Eren-Jeager123: Will record as action item

.github/config/image/base-cu129-devel.yml
Eren-Jeager123: Done - add pytorch/uv version for now

.github/workflows/autorelease-base-devel.yml
```diff
@@ -0,0 +1,278 @@
+name: Auto Release - Base Devel
+
+on:
+  # schedule run must be on the default branch.
+  # One cron per CUDA version — each run releases a single devel image.
+  schedule:
+    - cron: '40 20 * * 1,3'   # cu129 (CUDA 12.9) — 12:40 PM PST Mon/Wed
```
sirutBuasai: the schedule between runtime and devel is running at the same time, we should stagger them. Make sure none of our autorelease workflows for all images have a collision.

.github/config/image/base-cu129-devel.yml
```diff
@@ -16,12 +16,17 @@ common:
...
+  container_type: "devel"
   build_target: "devel"
   ci_tag: "cuda129-devel-amzn2023"
   device_type: "gpu"
   contributor: "None"
-  prod_image: "base:devel-cuda-v1"
+  prod_image: "base:devel-cu129-amzn2023"
+
+# Build pins passed to the Dockerfile as --build-args (UPPER_CASE keys)
+build:
```
sirutBuasai, line 27: The way sglang build args work is that it's fully dynamic, if we add additional environments into this field, that env is added to the build args automatically. We want to do the same here and then remove the version.env file from the docker/ path overall

.github/workflows/pr-base-devel.yml
```diff
@@ -0,0 +1,328 @@
...
+      - "!docs/**"
+
+permissions:
+  contents: read
+  pull-requests: read
+
+env:
+  FORCE_COLOR: "1"
+  TARGET: devel
+  LATEST_CUDA_VERSION: "cu130"
```
sirutBuasai: In the check-changes job, discover the latest version from the filesystem instead of hardcoding:
```
  LATEST_CUDA_VERSION=$(ls .github/config/image/base-cu*-${TARGET}.yml \
    | grep -oP 'cu\d+' | sort -V | tail -1)
```
it's a oneline and we wont have to maintain this when adding new CUDA versions.

The LATEST makes sense in autorelease because it can be manually dispatched with an input being part of env. But for PR, this is unnecessary and adds overhead maintenance.

.github/config/image/base-cu129-devel.yml
```diff
@@ -16,12 +16,17 @@ common:
   os_version: "amzn2023"
   customer_type: "ec2"
   arch_type: "x86"
-  container_type: "devel-cuda"
+  container_type: "devel"
```
sirutBuasai, line 19: what is container_type actually used for?

sirutBuasai: mainly it's used for telemetry and sagemaker labels. So what's your plan for this container_type? since base doesnt have sagemaker, only telemetry is applicable. Have you run this through a telemetry test?

.github/workflows/pr-base-devel.yml
```diff
@@ -0,0 +1,328 @@
...
+
+      - name: Build ${{ matrix.ci_tag }}
+        run: |
+          CI_IMAGE_URI="${{ vars.CI_AWS_ACCOUNT_ID }}.dkr.ecr.${{ vars.AWS_REGION }}.amazonaws.com/ci:base-${{ matrix.ci_tag }}-pr-${{ github.event.pull_request.number }}"
+          echo "Building ${CI_IMAGE_URI}"
+          # Expand the config build: pins (KEY=value pairs) into --build-arg flags
+          BUILD_ARG_FLAGS=""
+          for pair in ${{ matrix.build_args }}; do
+            BUILD_ARG_FLAGS="${BUILD_ARG_FLAGS} --build-arg ${pair}"
+          done
```
sirutBuasai: the build-args output is space-separated KEY=value pairs like PYTHON_VERSION=3.13.12 UV_VERSION=0.11.17. If a value ever contained a space (e.g., SOME_ARG=hello world), word splitting would break this

SGlang already handles this in the build-image action. You should reuse the build-image action instead of creating your own build step.

The whole reason we're doing workflows matrix and resuable workflow is because we want things to be repeatable without duplicating code.

This comment applies to autorelease and pr workflows

.github/workflows/autorelease-base-devel.yml
Eren-Jeager123: Thanks for pointing this, let me change it. We should have a global schedule config in the future, otherwise is hard to maintain

.github/config/image/base-cu129-devel.yml
Eren-Jeager123: Can you clarify this? We don't have version.env file for base image dockerfile.

.github/config/image/base-cu129-devel.yml
Eren-Jeager123: This is only being used for our release logic. Previous the tag is `devel-cuda-v1`, now the tag is `devel-cu130-amzn2023`, so the container_type change. Also base image doesn't have telemetry test.

.github/config/image/base-cu129-devel.yml
sirutBuasai: My mistake then, all good

sirutBuasai (dismissed):

sirutBuasai (approved):

PR #6243, opened 2026-06-15
feat(tensorflow): add TF Serving 2.20 inference DLC on AL2023

bhanutejagk:
## Summary

The first TensorFlow inference DLC on the `main` (v2) branch. TensorFlow Serving 2.20.0 serving TF 2.21-trained SavedModels on AL2023 + CUDA 12.9.1 + Python 3.12, SageMaker-only, x86-only, with single-model and Multi-Model Endpoint (MME) support.

- New `docker/tensorflow/inference/2.20/` tree — `Dockerfile.cuda` and `Dockerfile.cpu` on AL2023, uv-driven dependency management, cuDNN pinned via `nvidia-cudnn-cu12==9.24.0.43`.
- `nginx-mod-njs` installed from AL2023 core (`--releasever latest`) — no source-build stage needed.
- SageMaker handler scripts ported from master TF 2.19's `build_artifacts/sagemaker/` with 3 inherited-master bug fixes surfaced by review: `GET /models/{name}` returning 500, missing traversal guard, `ProcessLookupError` leak on DELETE.
- New GitHub Actions workflows (CPU + CUDA PR triggers, shared pipeline, SageMaker integration tests, workflow-dispatch-only autorelease with security-test gating).
- 12 SageMaker integration tests + 1 sanity test file (30 test functions) covering single-model, MME dynamic load, MME traversal, MME concurrency, batch transform, content-types (JSON + CSV), custom `inference.py`, error paths, model versioning, nginx/gunicorn env tuning, TFS batching, and GPU-only Conv2D exercising cuDNN kernel dispatch.

## Why this PR

This is the first TensorFlow inference image on `main` (v2 architecture). The master branch's TF 2.19 inference image is on Ubuntu 22.04 + Python 3.10 + CUDA 12.2. Porting to `main`'s AL2023 + Python 3.12 + CUDA 12.9.1 substrate is the migration this PR delivers, aligned with the TF 2.21 training PR (#6107) already merged on `main`.

## Test plan

- [x] CI builds image successfully (both CPU and CUDA variants)
- [x] Sanity tests pass (imports, `tensorflow_model_server --version`, cuDNN sub-library presence via `ldconfig -p`, handler artifacts, entrypoint executable, `exec`-hands-off-PID1 assertion for SIGTERM-critical scripts)
- [x] Security test / ECR vulnerability scan passes
- [x] Telemetry test passes (environment + instance)
- [x] SageMaker integration tests pass on both CPU and CUDA (~41m CPU / ~53m CUDA)
- [x] Pre-commit passes (ruff, dockerfmt, mdformat, gitleaks)
- [x] OSS compliance check passes

## After merge

Follow-up PR + tasks required before customer-facing release:

1. **Flip to production** — set `environment : production` in `.github/config/image/tensorflow/2.20-inference-sagemaker-{cpu,cuda}.yml`
2. **Add cron schedule to autorelease** — add `schedule:` trigger (Tue/Thu 17:30 UTC, matching training) to `autorelease-tensorflow-inference-sagemaker-{cpu,cuda}.yml`
3. **Publish release logic CR** — [CR-293646336](https://code.amazon.com/reviews/CR-293646336) (tag template + inference parity tests in DLContainersReleaseLogicPython)
4. **Post-release tasks:**
   - Update available images doc (`docs/src/data/tensorflow-inference/*.yml`)
   - Add images to Security Scanning Dashboard
   - Update DLC Support Policy (ga date, eop = ga + 12 months)
   - Create release notes (`docs/tensorflow/` tree)
   - Update new image to ADC regions (ConsumableImages CR + ADC CDK + Special Partitions CDK)
   - Public ECR gallery content update

## Pre-review notes

- **Known limitation — LSTM/GRU inference not supported on TFS 2.20**: TFS 2.20 cannot compile TF 2.21 Keras 3 LSTM `while_loop` graphs under its XLA/JIT autotuner. Same forward-compat gap class as DecodeJxl. Not a deferral — LSTM/GRU inference is simply out-of-scope for this image; customers should use TFS 2.21+ or CPU inference for those workloads.

## Related

- #6107 — TF 2.21 training PR. This inference PR runs on the same v2/AL2023 substrate; workflow shapes mirror it, and the shared `test/security/data/ecr_scan_allowlist/tensorflow/framework_allowlist.json` file ships from there.
- [Tracking Doc](https://chorus.aws.dev/doc/0sSZwA6Xzu7p/TensorFlow-Serving-220-Inference-Tracking-Doc) — TFS 2.20 inference tracking doc (Chorus)


.github/workflows/tensorflow-inference.autorelease-2.20-sagemaker.yml
sirutBuasai, line 1: The naming convention should be <framework>.<workflow_type>. So in this case, it should be `tensorflow-inference.<pipeline/pr/autorelease>`. This change should propagate to the training workflow into `tensorflow-training` as well. Take a look at other existing workflow.

test/tensorflow/integration/inference/conftest.py
```diff
@@ -0,0 +1,199 @@
...
+@pytest.fixture(scope="session")
+def sagemaker_session(boto_session):
+    """SageMaker SDK v3 session (default_bucket / upload_data)."""
+    from sagemaker.core.helper.session_helper import Session
+
+    return Session(boto_session=boto_session)
+
+
+@pytest.fixture
+def unique_name():
```
sirutBuasai: There is already an existing function that does this in `test_utils.random_suffix_name()`

docker/tensorflow/inference/2.20/Dockerfile.cpu
```diff
@@ -0,0 +1,232 @@
...
+# Hard-set (do not expand ${LD_LIBRARY_PATH}) — it is unset here and would
+# collapse to a trailing colon, which ld.so treats as "." (CWD injection).
+ENV LD_LIBRARY_PATH="/usr/local/lib"
+
+# Models dir
+RUN mkdir -p ${MODEL_BASE_PATH}
+
+# License file (S3 bucket pre-provisioned for tensorflow-2.20)
+RUN curl -fsSLo /license.txt "https://aws-dlc-licenses.s3.amazonaws.com/tensorflow-${TFS_SHORT_VERSION}/license.txt" \
+ || echo "tensorflow-${TFS_SHORT_VERSION}/license.txt not yet provisioned in aws-dlc-licenses bucket; placeholder emitted" >/license.txt
```
sirutBuasai: Remove this placeholder, we need the actual license in the docker contiainer

docker/tensorflow/inference/2.20/Dockerfile.cuda
```diff
@@ -0,0 +1,262 @@
...
+# Hard-set (do not expand ${LD_LIBRARY_PATH}) — it is unset here and would
+# collapse to a trailing colon, which ld.so treats as "." (CWD injection).
+ENV LD_LIBRARY_PATH="/usr/local/cuda/lib64:/usr/local/lib"
+
+# Models dir
+RUN mkdir -p ${MODEL_BASE_PATH}
+
+# License file (S3 bucket pre-provisioned for tensorflow-2.20)
+RUN curl -fsSLo /license.txt "https://aws-dlc-licenses.s3.amazonaws.com/tensorflow-${TFS_SHORT_VERSION}/license.txt" \
+ || echo "tensorflow-${TFS_SHORT_VERSION}/license.txt not yet provisioned in aws-dlc-licenses bucket; placeholder emitted" >/license.txt
```
sirutBuasai: Remove placeholder

scripts/docker/tensorflow/inference/sagemaker/python_service.py
```diff
@@ -0,0 +1,761 @@
...
+            self._sync_local_mme_instance_status()
+            if model_name is None:
+                models_info = {}
+                uri = "http://localhost:{}/v1/models/{}"
+                for model, tfs_instance_status in self._mme_tfs_instances_status.items():
+                    try:
+                        info = json.loads(
+                            requests.get(
+                                uri.format(tfs_instance_status[0].rest_port, model)
+                            ).content
```
sirutBuasai: if a TFS instance hangs, /models and GET /models/{name} block forever while holding the file lock

scripts/docker/tensorflow/inference/sagemaker/python_service.py
```diff
@@ -0,0 +1,761 @@
...
+                    res.body = json.dumps(
+                        {"error": "Model {} is not loaded yet.".format(model_name)}
+                    ).encode("utf-8")
+                else:
+                    # Value is a list of TfsInstanceStatus (see _load_model);
+                    # pick the first instance's rest_port.
+                    port = self._mme_tfs_instances_status[model_name][0].rest_port
+                    uri = "http://localhost:{}/v1/models/{}".format(port, model_name)
+                    try:
+                        info = json.loads(requests.get(uri).content)
```
sirutBuasai: if a TFS instance hangs, /models and GET /models/{name} block forever while holding the file lock

scripts/docker/tensorflow/inference/sagemaker/python_service.py
```diff
@@ -0,0 +1,761 @@
...
+                                uri.format(tfs_instance_status[0].rest_port, model)
+                            ).content
+                        )
+                        models_info[model] = info
+                    except ValueError as e:
+                        log.exception("exception handling request: {}".format(e))
+                        res.status = falcon.HTTP_500
+                        res.body = json.dumps({"error": str(e)}).encode("utf-8")
+                res.status = falcon.HTTP_200
+                res.body = json.dumps(models_info)
```
sirutBuasai: the list-all branch sets status=500 on a per-model error but doesn't return/break so  `res.status = falcon.HTTP_200` cast it as 200 anyway. We need a corrupt model to return 500 correctly

test/tensorflow/integration/inference/conftest.py
```diff
@@ -0,0 +1,199 @@
...
+    return os.environ.get("AWS_REGION", "us-west-2")
+
+
+@pytest.fixture(scope="session")
+def sagemaker_role_arn() -> str:
+    """SageMaker execution role ARN. Skips the test if not set."""
+    arn = os.environ.get("SM_ROLE_ARN")
+    if not arn:
+        pytest.skip("SM_ROLE_ARN not set")
+    return arn
```
sirutBuasai: these already exists in `test/conftest.py`

also the SM_ROLE_ARN should use the `AWSSessionManager.resolve_role_arn()` from `test_utils/aws.py`.


Abstract things that are generic to the test suite to the parent module for reusability rather than hard coding it into the framework test suite.

test/tensorflow/integration/inference/conftest.py
```diff
@@ -0,0 +1,199 @@
...
+    return _SM_INSTANCE_TYPE_BY_DEVICE[device]
+
+
+@pytest.fixture(scope="session")
+def boto_session(aws_region: str):
+    """boto3 session bound to the configured region."""
+    import boto3
+
+    return boto3.Session(region_name=aws_region)
+
```
sirutBuasai: these already exists in `test/conftest.py`.

You should evaluate if certain fixtures can be generalized. If so, it should live in the parent folder conftest.

test/tensorflow/integration/inference/conftest.py
```diff
@@ -0,0 +1,199 @@
...
+        # Endpoint config name == endpoint name in our deploy flow.
+        for resource_cls, get_kwargs in (
+            (Endpoint, {"endpoint_name": endpoint_name}),
+            (EndpointConfig, {"endpoint_config_name": endpoint_name}),
+        ):
+            try:
+                resource_cls.get(session=boto_session, **get_kwargs).delete()
+            except Exception:
+                # Best-effort teardown: swallow NotFound / already-deleted.
+                pass
```
sirutBuasai: Instead of passing on exception. Take a look at the patterns of `_cleanup` in sglang and vllm. It is essentually using `finally` block within test setup fixtures rather than encaptulating cleanup as its own fixtures. It also uses logging on failures.

test/tensorflow/integration/inference/test_batch_transform.py
```diff
@@ -0,0 +1,160 @@
...
+                    break
+                if time.time() > deadline:
+                    raise TimeoutError(
+                        f"TransformJob {job_name} still in status {status!r} after 45 min"
+                    )
+                time.sleep(30)
+            assert status == "Completed", (
+                f"TransformJob failed: status={status!r} "
+                f"failure_reason={getattr(job, 'failure_reason', None)!r}"
+            )
```
sirutBuasai: Use `test_utils.wait_for_status()`. See pattern in existing test suite like sglang or vllm endpoint tests.

test/tensorflow/integration/inference/test_single_model_endpoint.py
```diff
@@ -0,0 +1,107 @@
...
+        # 3. Endpoint; wait for InService.
+        endpoint = Endpoint.create(
+            endpoint_name=endpoint_name,
+            endpoint_config_name=endpoint_name,
+            session=boto_session,
+        )
+        endpoint.wait_for_status("InService")
+
+        # 4. Invoke; body is a streaming bytes-like object.
+        payload = json.dumps({"instances": [[1.0, 2.0, 3.0]]})
```
sirutBuasai: Why not use deploy_endpoint fixture?

test/tensorflow/integration/inference/test_multi_model_endpoint.py
```diff
@@ -0,0 +1,126 @@
...
+        )
+
+        endpoint = Endpoint.create(
+            endpoint_name=endpoint_name,
+            endpoint_config_name=endpoint_name,
+            session=boto_session,
+        )
+        endpoint.wait_for_status("InService")
+
+        payload = json.dumps({"instances": [[1.0, 2.0, 3.0]]})
```
sirutBuasai: Why not use deploy_endpoint fixture?

test/tensorflow/integration/inference/test_tfs_batching.py
```diff
@@ -0,0 +1,57 @@
...
+            container_env={
+                "SAGEMAKER_TFS_ENABLE_BATCHING": "true",
+                # Small knobs so batching engages promptly.
+                "SAGEMAKER_TFS_MAX_BATCH_SIZE": "8",
+                "SAGEMAKER_TFS_BATCH_TIMEOUT_MICROS": "1000",
+                "SAGEMAKER_TFS_NUM_BATCH_THREADS": "1",
+                "SAGEMAKER_TFS_MAX_ENQUEUED_BATCHES": "10000",
+            },
+        )
+        cleanup_endpoint(endpoint_name, model_name=model_name)
```
sirutBuasai: If deploy_endpoint already consumes cleanup_endpoint internally at the end already, why do we need to recall this again? This applies to all the tests files that are doing this redundant pattern.

test/tensorflow/integration/inference/test_error_paths.py
```diff
@@ -0,0 +1,108 @@
...
+import pytest
+from botocore.exceptions import ClientError
+
+from .resources.build_sample_model import build_sample_model
+from .resources.helpers import upload_tarball
+
+# Scenarios: (scenario_id, body, content_type, extra_body_check).
+# ``extra_body_check`` runs on the container's response body string; None
+# means "just require non-empty status in the 4xx/5xx range".
+ERROR_SCENARIOS = [
```
sirutBuasai: Make this a pytest parametrize for the test to consume. This way the pytest report is clean and easily detected and debugged.

test/tensorflow/integration/inference/test_conv_gpu.py
```diff
@@ -0,0 +1,83 @@
...
+from .resources.helpers import read_predictions, upload_tarball
+
+# Explicit skipif so the skip decision is visible at collection time.
+_device = os.environ.get("SM_DEVICE_TYPE", "").lower()
+assert _device in {"cpu", "gpu"}, f"SM_DEVICE_TYPE must be 'cpu' or 'gpu'; got {_device!r}"
+pytestmark = pytest.mark.skipif(
+    _device != "gpu",
+    reason="cuDNN Conv2D smoke requires SM_DEVICE_TYPE=gpu (workflow sets this for GPU device-type only)",
+)
+
```
sirutBuasai, line 25: why not just use @pytest.mark.gpu

.github/workflows/tensorflow-inference.autorelease-2.20-sagemaker.yml
bhanutejagk: Sounds good. Modified

docker/tensorflow/inference/2.20/Dockerfile.cpu
bhanutejagk: done

docker/tensorflow/inference/2.20/Dockerfile.cuda
bhanutejagk: done

test/tensorflow/integration/inference/conftest.py
bhanutejagk: ok using it.

scripts/docker/tensorflow/inference/sagemaker/python_service.py
bhanutejagk: Added timeout=5 to both requests.get() calls. Also expanded the except clause to catch requests.exceptions.RequestException (which includes Timeout, ConnectionError, etc.)

test/tensorflow/integration/inference/conftest.py
bhanutejagk: Also verified for any other fixtures that can be used from parent and removed boto_session by replacing it with aws_session.session. matches sglang pattern

scripts/docker/tensorflow/inference/sagemaker/python_service.py
bhanutejagk: Added timeout=5 to both requests.get() calls. Also expanded the except clause to catch requests.exceptions.RequestException (which includes Timeout, ConnectionError, etc.)

scripts/docker/tensorflow/inference/sagemaker/python_service.py
bhanutejagk: Yeah, nice catch. Added `return` after the `except` block.

test/tensorflow/integration/inference/conftest.py
bhanutejagk: sure.

test/tensorflow/integration/inference/conftest.py
bhanutejagk: sure

test/tensorflow/integration/inference/conftest.py
bhanutejagk: Ok made changes to sync with other framework patterns

test/tensorflow/integration/inference/test_batch_transform.py
bhanutejagk: done. using `wait_for_status()` now.

test/tensorflow/integration/inference/test_multi_model_endpoint.py
bhanutejagk: Its just a gap. Good call. Using `deploy_endpoint` now.

test/tensorflow/integration/inference/test_single_model_endpoint.py
bhanutejagk: Its just a gap. Good call. Using `deploy_endpoint` now.

test/tensorflow/integration/inference/test_tfs_batching.py
bhanutejagk: This is redundant. Removed all cleanup_endpoint calls from every test file. and removed seperate `cleanup_endpoint` fixture entirely since `deploy_endpoint`  fixture's `finally` block now handles all cleanup internally.

test/tensorflow/integration/inference/test_error_paths.py
bhanutejagk: done changed it. BTW we are not using deploy_endpoint in this file since we created error_endpoint as module scoped and the endpoint is shared between all the 4 tests.

test/tensorflow/integration/inference/test_conv_gpu.py
bhanutejagk: done.

scripts/docker/tensorflow/inference/sagemaker/python_service.py
```diff
@@ -0,0 +1,793 @@
...
+                    ).encode("utf-8")
+                else:
+                    # Value is a list of TfsInstanceStatus (see _load_model);
+                    # pick the first instance's rest_port.
+                    port = self._mme_tfs_instances_status[model_name][0].rest_port
+                    uri = "http://localhost:{}/v1/models/{}".format(port, model_name)
+                    try:
+                        r = requests.get(uri, timeout=5)
+                        if r.status_code != 200:
+                            res.status = str(r.status_code)
```
sirutBuasai: `res.status = str(r.status_code)` writes `"404"` as the WSGI status line — no reason phrase, so gunicorn rejects it and the caller gets a 500 instead of the intended 4xx. Use `falcon.code_to_http_status(r.status_code)` like the other paths in this file.

test/tensorflow/integration/inference/test_error_paths.py
```diff
@@ -0,0 +1,141 @@
...
+            )
+            endpoint = Endpoint.create(
+                endpoint_name=endpoint_name,
+                endpoint_config_name=endpoint_name,
+                session=session,
+            )
+            endpoint.wait_for_status("InService")
+            yield endpoint
+        finally:
+            _cleanup([endpoint, endpoint_config, model_obj])
```
sirutBuasai: `error_endpoint` re-implements `deploy_endpoint` almost line-for-line (create → `wait_for_status` → GPU conditional → try/finally `_cleanup`), only because it needs `scope="module"`. Extract a shared `_deploy_endpoint(...)` helper both consume rather than keeping a second copy.

test/tensorflow/integration/inference/test_batch_transform.py
```diff
@@ -0,0 +1,164 @@
...
+            key_prefix=f"tf220-inference-tests/batch/{run_id}/model",
+        )
+
+        # 2. Write 3 input files — one JSON body per file (SplitType=None).
+        input_dir = workdir_path / "input"
+        input_dir.mkdir()
+        for i, row in enumerate([[1.0, 2.0, 3.0], [4.0, 5.0, 6.0], [7.0, 8.0, 9.0]]):
+            (input_dir / f"row_{i}.json").write_text(json.dumps({"instances": [row]}))
+
+        s3 = aws_session.session.client("s3")
```
sirutBuasai: `aws_session.session.client("s3")` builds a fresh client — `AWSSessionManager` already exposes `aws_session.s3`. Same at lines 71, 134, 149.

test/tensorflow/integration/inference/test_conv_gpu.py
```diff
@@ -0,0 +1,73 @@
...
+import tempfile
+
+import pytest
+from test_utils import random_suffix_name
+
+from .resources.build_sample_model import build_conv_sample_model
+from .resources.helpers import read_predictions, upload_tarball
+
+
+@pytest.mark.gpu
```
sirutBuasai: `@pytest.mark.gpu` isn't registered anywhere and nothing consumes it, so on a CPU run this Conv2D test is still collected and runs on CPU — defeating the cuDNN purpose (and it warns under `--strict-markers`). Either register + gate the marker, or `@pytest.mark.skipif(sm_device_type != "gpu")`. The sanity file gates correctly via `unittest.skipIf`.

scripts/docker/tensorflow/inference/sagemaker/multi_model_utils.py
```diff
@@ -0,0 +1,104 @@
...
+def timeout(seconds=60):
+    def _raise_timeout_error(signum, frame):
+        raise Exception(408, "Timed out after {} seconds".format(seconds))
+
+    try:
+        signal.signal(signal.SIGALRM, _raise_timeout_error)
+        signal.alarm(seconds)
+        yield
+    finally:
+        signal.alarm(0)
```
sirutBuasai: The `timeout` contextmanager is dead code (only `MultiModelException, lock` are imported), uses `signal.SIGALRM` which doesn't work reliably under `gevent.monkey.patch_all()`, and raises a bare `Exception(408, ...)` missing the `.code`/`.pid` the loader expects. Carried from master — delete it.

test/tensorflow/integration/inference/conftest.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+@pytest.fixture(scope="session")
+def sm_instance_type(sm_device_type) -> str:
+    """SM endpoint instance type derived from device type."""
+    return _SM_INSTANCE_TYPE_BY_DEVICE[sm_device_type]
+
+
+@pytest.fixture(scope="session")
+def sagemaker_session(aws_session):
```
sirutBuasai: `sagemaker_session` is generic — nothing TF-specific. Consider promoting it to `test/conftest.py` so other framework suites reuse it instead of reinventing, same as the other shared fixtures.

scripts/docker/tensorflow/inference/sagemaker/tensorflowServing.js
```diff
@@ -0,0 +1,273 @@
...
+                }
+            }
+        }
+    }
+
+    // for MME invocations, tfs-model-name is in the uri, or use default_tfs_model
+    if (!attributes['tfs-model-name']) {
+        var uri_pattern = /\/models\/[^,]+\/invoke/g
+        var model_name = r.uri.match(uri_pattern)
+        if (model_name[0]) {
```
sirutBuasai: `r.uri.match(uri_pattern)` returns `null` for `/ping` and `/invocations`, so `model_name[0]` throws a TypeError. On a single-model deployment (`_use_gunicorn == False`) these routes go through this njs handler, so ping + invocations 500 and the endpoint never reaches InService. Guard it: `if (model_name && model_name[0])`.

scripts/docker/tensorflow/inference/sagemaker/python_service.py
```diff
@@ -0,0 +1,793 @@
...
+                            ).content
+                        )
+                        models_info[model] = info
+                    except (ValueError, requests.exceptions.RequestException) as e:
+                        log.exception("exception handling request: {}".format(e))
+                        res.status = falcon.HTTP_500
+                        res.body = json.dumps({"error": str(e)}).encode("utf-8")
+                        return
+                res.status = falcon.HTTP_200
+                res.body = json.dumps(models_info)
```
sirutBuasai: Same lock-across-network-I/O concern as the GET-single-model path: `on_get` holds `lock()` while looping `requests.get` over every model (up to N×5s), serializing all model management across workers. Snapshot the model→port map under the lock, release, then fan out the HTTP calls unlocked. (Load path at ~298-357 has the same shape — lock spans `Popen` + `wait_for_model`.)

test/tensorflow/integration/inference/test_batch_transform.py
```diff
@@ -0,0 +1,164 @@
...
+                first = predictions[0]
+                values = first["output"] if isinstance(first, dict) and "output" in first else first
+                filename = key.rsplit("/", 1)[-1]
+                assert values == pytest.approx(expected_by_input[filename]), (
+                    f"{filename}: got {values!r}, expected {expected_by_input[filename]!r}"
+                )
+        finally:
+            # Best-effort model cleanup; TransformJob is a completed record.
+            try:
+                Model.get(model_name=model_name, session=aws_session.session).delete()
```
sirutBuasai: The `finally` re-`get()`s the model and does its own try/except/warning — that's a duplicate of `_cleanup`. It only needs the re-`get()` because the `Model.create(...)` at line 78 is discarded. Capture it and call `_cleanup([model])` (like `test_error_paths.py` does).

scripts/docker/tensorflow/inference/sagemaker/python_service.py
```diff
@@ -0,0 +1,793 @@
...
+        self.pid = pid
+
+    def __repr__(self):
+        return f"TFS Instance Status (rest_port : {self.rest_port}, grpc_port: {self.grpc_port}, pid: {self.pid}))"
+
+
+class PythonServiceResource:
+    def __init__(self):
+        if SAGEMAKER_MULTI_MODEL_ENABLED:
+            self._mme_tfs_instances_status: dict[str, [TfsInstanceStatus]] = {}
```
sirutBuasai: `dict[str, [TfsInstanceStatus]]` — `[TfsInstanceStatus]` is a list literal, not a type. Should be `dict[str, list[TfsInstanceStatus]]`.

scripts/docker/tensorflow/inference/sagemaker/serve.py
```diff
@@ -0,0 +1,535 @@
...
+        bucket = os.environ.get("SAGEMAKER_MULTI_MODEL_UNIVERSAL_BUCKET", None)
+        prefix = os.environ.get("SAGEMAKER_MULTI_MODEL_UNIVERSAL_PREFIX", None)
+
+        if not os.path.exists(CODE_DIR) and bucket and prefix:
+            self._download_scripts(bucket, prefix)
+
+        if self._enable_python_service:
+            lib_path_exists = os.path.exists(PYTHON_LIB_PATH)
+            requirements_exists = os.path.exists(REQUIREMENTS_PATH)
+            python_path_content = ["/opt/ml/model/code"]
```
sirutBuasai: `python_path_content = ["/opt/ml/model/code"]` is hardcoded even in MME/universal mode where `CODE_DIR == "/opt/ml/code"`, so PYTHONPATH points at the wrong dir. Use `CODE_DIR`. (Low impact — universal `inference.py` is loaded by absolute path.)

.github/workflows/tensorflow-inference.autorelease-2.20-sagemaker.yml
```diff
@@ -0,0 +1,37 @@
+name: "Auto Release - TensorFlow Inference 2.20 SageMaker"
+
+on:
+  workflow_dispatch:
```
sirutBuasai, line 9: Named "Auto Release" but `workflow_dispatch`-only, while training's autorelease has a `schedule:` cron. Matches the after-merge checklist (add cron on prod flip), so likely intentional while gamma — just confirming the name/behavior mismatch is deliberate.

scripts/docker/tensorflow/inference/sagemaker/python_service.py
bhanutejagk: okay.

test/tensorflow/integration/inference/test_error_paths.py
bhanutejagk: extracted `_provision_endpoint()` into conftest.py:53; both `deploy_endpoint` (conftest.py:156) and `error_endpoint` (test_error_paths.py:54) now call it, so endpoint setup lives in exactly one place and the scope difference is just the fixture decorator. Thanks for pushing on it. .

test/tensorflow/integration/inference/test_batch_transform.py
bhanutejagk: done.

scripts/docker/tensorflow/inference/sagemaker/multi_model_utils.py
bhanutejagk: deleted

test/tensorflow/integration/inference/conftest.py
bhanutejagk: Promoted to `test/conftest.py`

scripts/docker/tensorflow/inference/sagemaker/tensorflowServing.js
bhanutejagk: Guarded — tensorflowServing.js is now `if (model_name && model_name[0])`.

scripts/docker/tensorflow/inference/sagemaker/python_service.py
bhanutejagk: Fixed for on_get — it now snapshots the model→port map under the lock (:560-572) and releases before doing any HTTP as you suggested. 

Also left the load path holding the lock deliberately: `_handle_load_model_post` allocates ports, writes model-config.cfg, and mutates the shared status dict, so two concurrent LoadModels would race on port assignment and config writes — the serialization there is the point.

test/tensorflow/integration/inference/test_batch_transform.py
bhanutejagk: Fixed

scripts/docker/tensorflow/inference/sagemaker/python_service.py
bhanutejagk: Fixed

scripts/docker/tensorflow/inference/sagemaker/serve.py
bhanutejagk: done

.github/workflows/tensorflow-inference.autorelease-2.20-sagemaker.yml
bhanutejagk: Yes this is intentional so that after gamma testing the cron can be added.

scripts/docker/tensorflow/inference/sagemaker/python_service.py
```diff
@@ -0,0 +1,827 @@
...
+            self._default_handlers_enabled = True
+
+        self._tfs_enable_batching = SAGEMAKER_BATCHING_ENABLED == "true"
+        self._tfs_default_model_name = os.environ.get("TFS_DEFAULT_MODEL_NAME", "None")
+        self._tfs_inter_op_parallelism = os.environ.get("SAGEMAKER_TFS_INTER_OP_PARALLELISM", 0)
+        self._tfs_intra_op_parallelism = os.environ.get("SAGEMAKER_TFS_INTRA_OP_PARALLELISM", 0)
+        self._tfs_instance_count = int(os.environ.get("SAGEMAKER_TFS_INSTANCE_COUNT", 1))
+        self._gunicorn_workers = int(os.environ.get("SAGEMAKER_GUNICORN_WORKERS", 1))
+        self._tfs_wait_time_seconds = int(
+            os.environ.get("SAGEMAKER_TFS_WAIT_TIME_SECONDS", 55 // self._tfs_instance_count)
```
sirutBuasai: I dont think we should default to a wait time inverse scaling with instance count. We can default to a simple flat number like 60, 120, 180

.github/workflows/tensorflow-inference.tests-sagemaker.yml
```diff
@@ -0,0 +1,90 @@
...
+      - name: Run SageMaker inference tests
+        env:
+          PYTHONPATH: ${{ github.workspace }}/test
+          SM_DEVICE_TYPE: ${{ inputs.device-type }}
+        run: |
+          source .venv/bin/activate
+          MARK_FILTER=()
+          if [ "${{ inputs.device-type }}" = "cpu" ]; then
+            MARK_FILTER=(-m "not gpu")
+          fi
```
sirutBuasai: No need for this now

sirutBuasai (approved):

PR #6283, opened 2026-06-24
ci(pytorch): re-enable PR workflow triggers post-refactor

Eren-Jeager123:
## Purpose

Re-enable the **PyTorch** PR workflows after the workflow refactor (#6250), which had temporarily disabled every framework's `pull_request:` trigger behind a `# TEMP: Enable after all frameworks are refactored` marker. First of the post-merge re-enable action items.

Validating with a real CI run surfaced several latent regressions/bugs from the refactor, fixed here.

## Changes

**Re-enable + triggering**
- Uncomment the `pull_request:` trigger and drop the TEMP marker in `pytorch.pr-2.11-{cuda,cpu}.yml`; remove `workflow_dispatch` (PR workflows don't need it).
- Add `docker/pytorch/2.11/{cuda,cpu}/**` to the trigger `paths` and `build-change` filter — the workflows watched the Dockerfile but not `pyproject.toml` / `uv.lock`, so a dependency-only change triggered no build or scan.

**SageMaker**
- Build sagemaker images from the **sagemaker stage**: all four configs had `build.target: runtime`, so the `*-sagemaker` images were built from the EC2/runtime stage — missing the SM toolkit, `/opt/ml` BYOC dirs, and `SAGEMAKER_TRAINING_MODULE` (i.e. identical to EC2). Set `target: sagemaker` in the two sagemaker configs.
- Wire up the SageMaker training tests: added reusable `pytorch.tests-sagemaker.yml` + a `sagemaker-test` job gated on `customer-type == 'sagemaker'`. The tests existed but the refactor dropped the job that ran them; selects cpu/cuda by device type and launches remote SM training jobs.

**EFA**
- Run EFA on the **EC2 leg only** (`customer-type == 'ec2'`). The reusable EFA workflow uses a single global concurrency group (`efa-test-global`) to serialize EFA repo-wide — only one 2× p4d.24xlarge job runs at a time (limited p4d/EIP capacity). The sagemaker image is the same build as EC2 (sagemaker stage only layers extra args/packages), so EFA behavior is identical and testing it once on EC2 fully covers it; running both legs would just queue/cancel each other under the global group.

**Test workflows — preflight pattern**
- Refactor all five reusable PyTorch test workflows (unit, single-gpu, multi-gpu, multi-node-gpu, sagemaker) to match vllm/sglang: a `preflight` job resolves the image URI + checks existence (as outputs), and the test job `needs: [preflight]` gated by `image-exists == 'true'`. Removes per-step existence gating and the Skip notice step.

**CVE fixes**
- `msgpack >= 1.2.1` (GHSA-6v7p-g79w-8964) — transitive via deepspeed.
- `tornado >= 6.5.6` (GHSA-3x9g-8vmp-wqvf, GHSA-mgf9-4vpg-hj56) and `cryptography >= 48.0.1` (GHSA-537c-gmf6-5ccf) — transitive via the sagemaker extra (surfaced once the sagemaker stage builds correctly). Pins added to both cuda/cpu `pyproject.toml`; lock files regenerated.
- `quinn-proto` (RUSTSEC-2026-0185) is bundled inside the `uv` binary. uv is intentionally shipped (customers use it for venv management), so it is **not** removed — instead allowlisted under `test/security/data/ecr_scan_allowlist/pytorch_runtime/` (can't be patched independently of an upstream uv release).

## Test Plan / Test Result

- All PR checks green (build + all tests, cuda and cpu).
- Verified: dependency-path change flips `build-change`; EFA runs once (ec2) with no cross-leg cancellation; sagemaker-test runs against the correctly-built sagemaker image; security scan passes (Python CVEs fixed, quinn-proto allowlisted).

## Notes
- Scope is **PyTorch only**; other frameworks remain disabled and are tracked separately.
- A couple of empty `ci: retry` commits remain in history from re-running transient infra failures (NVIDIA CUDA mirror timeout; one-off SSH-connect timeout to a fresh p4d); they carry no net diff and squash away on merge.
- Out of scope (worth follow-ups): no `dnf` retry/timeout hardening in the build (slow NVIDIA mirror fails hard); EFA SSH-connect step has no readiness retry (occasional `Errno 110` on a just-booted p4d).

### DIFF
--- a/.github/workflows/pytorch.tests-sagemaker.yml
+++ b/.github/workflows/pytorch.tests-sagemaker.yml
@@ -75,7 +75,3 @@ jobs:
           cd test/
           PYTHONPATH=$(pwd):$PYTHONPATH python3 -m pytest -v --tb=short -rA --log-cli-level=INFO \
             ${{ steps.select.outputs.test-file }}
-
-      - name: Skip notice
-        if: steps.check.outputs.exists != 'true'
-        run: echo "Image not found in ECR — skipping tests (first release scenario)"
--- a/.github/workflows/pytorch.tests-sagemaker.yml
+++ b/.github/workflows/pytorch.tests-sagemaker.yml
@@ -17,13 +17,15 @@ permissions:
   contents: read
 
 jobs:
-  sagemaker-test:
-    # Launches remote SageMaker training jobs — no GPU needed on the runner.
+  preflight:
     runs-on:
       - codebuild-runner-${{ github.run_id }}-${{ github.run_attempt }}
         fleet:default-runner
         buildspec-override:true
-    timeout-minutes: 60
+    outputs:
+      image-uri: ${{ steps.image.outputs.image-uri }}
+      aws-account-id: ${{ steps.image.outputs.aws-account-id }}
+      image-exists: ${{ steps.check.outputs.exists }}
     steps:
       - uses: actions/checkout@v6
 
@@ -43,16 +45,26 @@ jobs:
         with:
           image-uri: ${{ steps.image.outputs.image-uri }}
 
+  sagemaker-test:
+    # Launches remote SageMaker training jobs — no GPU needed on the runner.
+    if: ${{ needs.preflight.outputs.image-exists == 'true' }}
+    needs: [preflight]
+    runs-on:
+      - codebuild-runner-${{ github.run_id }}-${{ github.run_attempt }}
+        fleet:default-runner
+        buildspec-override:true
+    timeout-minutes: 60
+    steps:
+      - uses: actions/checkout@v6
+
       - name: ECR login
-        if: steps.check.outputs.exists == 'true'
         uses: ./.github/actions/ecr-authenticate
         with:
-          aws-account-id: ${{ steps.image.outputs.aws-account-id }}
+          aws-account-id: ${{ needs.preflight.outputs.aws-account-id }}
           aws-region: ${{ vars.AWS_REGION }}
-          image-uri: ${{ steps.image.outputs.image-uri }}
+          image-uri: ${{ needs.preflight.outputs.image-uri }}
 
       - name: Select test file by device type
-        if: steps.check.outputs.exists == 'true'
         id: select
         run: |
           DEVICE=$(yq '.metadata.device_type' '${{ inputs.config-file }}')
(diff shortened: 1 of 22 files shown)


.github/workflows/pytorch.tests-sagemaker.yml
```diff
@@ -0,0 +1,81 @@
...
+        run: |
+          uv venv --python 3.12
+          source .venv/bin/activate
+          uv pip install -r test/requirements.txt
+          uv pip install -r test/pytorch/integration/sagemaker/requirements.txt
+          cd test/
+          PYTHONPATH=$(pwd):$PYTHONPATH python3 -m pytest -v --tb=short -rA --log-cli-level=INFO \
+            ${{ steps.select.outputs.test-file }}
+
+      - name: Skip notice
```
sirutBuasai: Can you actually remove this skip notice? Remove it from all pytorch tests workflows. I thought it removed it from everything but I guess I missed it.

docker/pytorch/2.11/Dockerfile.cpu
```diff
@@ -164,7 +164,8 @@ COPY --from=ghcr.io/astral-sh/uv:latest /uv /usr/local/bin/uv
 ENV UV_PROJECT_ENVIRONMENT="/opt/venv"
 COPY docker/pytorch/${FRAMEWORK_SHORT_VERSION}/cpu/pyproject.toml docker/pytorch/${FRAMEWORK_SHORT_VERSION}/cpu/uv.lock /tmp/build/
 RUN --mount=type=cache,target=/root/.cache/uv cd /tmp/build && uv sync --frozen --no-dev --extra sagemaker --no-install-project --inexact \
-  && rm -rf /tmp/build /tmp/uv-*
+  && rm -rf /tmp/build /tmp/uv-* \
+  && rm -f /usr/local/bin/uv
```
sirutBuasai: we dont actually wanna remove uv. Cx might use uv for virtual env. The fix for the CVE is to allowlist it.

docker/pytorch/2.11/Dockerfile.cuda
```diff
@@ -249,7 +249,8 @@ COPY --from=ghcr.io/astral-sh/uv:latest /uv /usr/local/bin/uv
 ENV UV_PROJECT_ENVIRONMENT="/opt/venv"
 COPY docker/pytorch/${FRAMEWORK_SHORT_VERSION}/cuda/pyproject.toml docker/pytorch/${FRAMEWORK_SHORT_VERSION}/cuda/uv.lock /tmp/build/
 RUN --mount=type=cache,target=/root/.cache/uv cd /tmp/build && uv sync --frozen --no-dev --extra sagemaker --no-install-project --inexact \
-  && rm -rf /tmp/build /tmp/uv-*
+  && rm -rf /tmp/build /tmp/uv-* \
```
sirutBuasai: ditto

sirutBuasai (dismissed):

sirutBuasai (approved):

PR #6285, opened 2026-06-24
Add OpenFold3 SageMaker DLC (build + tests, incl. precomputed-MSA under network isolation)

Eren-Jeager123:
## Purpose

Onboard **OpenFold3** (protein-structure prediction) into the DLC repo as a GPU **SageMaker async-inference** image, for **private release** to the JumpStart team.

### Image (`docker/openfold3/`, `scripts/docker/openfold3/`)
- **Dockerfile** — 2-stage build on `devel-cu130-amzn2023`; conda-lock environment from OpenFold3's upstream `production-linux-64.lock`; source pinned to upstream tag `0.4.1`; checkpoint `of3-p2-155k.pt` fetched from the public bucket over anonymous HTTPS (`curl`, not awscli). Runtime stays on the `devel` base — Triton/CUTLASS kernels JIT-compile at warmup.
- **Serving** — Flask + gunicorn (`/ping` + `/invocations`), SageMaker handler with a per-GPU pool. Per-GPU warmup compiles CUDA kernels (~5 min) at container startup so `/ping` only returns 200 once warm — the customer's first request is never slow.
- **MSA / network isolation** — MSA is **off by default** and the endpoint runs under **network isolation** (no outbound internet). Customers who need MSA supply a **precomputed alignment inline** in the request (`precomputed_msa: [{name, content}]` per chain); the handler stages it to disk and points OpenFold3 at it, so no network is required. The live MSA server (`use_msa_server=true`) needs outbound internet, so it is **rejected up front** with an actionable error pointing to the precomputed-MSA format — without the guard it would hang until the predict timeout (confirmed by a manual repro). Override via `OPENFOLD3_ALLOW_MSA_SERVER=1` for a non-isolated deploy.

### Config & workflows (`.github/`)
- `config/image/openfold3/sagemaker.yml` — `release: true`, `public_registry: false` (private-only). Published tag `0.4-cu130-amzn2023` (pytorch-runtime style), via the `openfold3` framework added to `DLContainersReleaseLogicPython`.
- `openfold3.pipeline.yml` — build → sanity → security → sagemaker → release.
- `openfold3.pr-amzn2023.yml` (PR trigger) + `openfold3.autorelease-sagemaker-amzn2023.yml` (scheduled) + `openfold3.tests-sagemaker.yml` (preflight + endpoint-test, vllm/sglang pattern).
- `release-schedule.yml` — autorelease cron registered.

### Security
- ECR enhanced scan flagged 43 CRITICAL/HIGH CVEs. **14 fixed** by upgrading pure-Python utility libs (GitPython, urllib3, pillow, dulwich, msgpack) in both conda envs + purging the conda package cache. **29 allowlisted** (`test/security/data/ecr_scan_allowlist/openfold3/`) — openssl/curl (conda-vendored native, upstream-pinned), torch 2.5.1 (official OpenFold3 pin), quinn-proto (inside the `uv` binary) — each with a `review_by` date.

### Tests (`test/openfold3/sagemaker/`)
- SageMaker async-endpoint integration test (SDK v3): one **`ml.g6.12xlarge`** endpoint (4× NVIDIA L4, 96 GB GPU) deployed with **`enable_network_isolation=True`** serves all five cases (`MaxConcurrentInvocationsPerInstance=4`). The CI runner itself is a CPU `default-runner` — it only drives boto3; all GPU work runs on the SageMaker endpoint instance. The five cases — small protein single request (smoke), large protein single request, **large protein with a customer-supplied precomputed MSA**, **`use_msa_server=true` rejected fast** (negative test for the isolation guard), and large protein 4-concurrent (asserts the GPU pool keeps 4-concurrent wall-time under 2× the single-request time).
- Query inputs and async output both use the account default `sagemaker-<region>-<account>` bucket (the SageMaker execution role can only read/write `*sagemaker*` buckets — writing elsewhere fails silently). Queries are pre-uploaded via `Scripts/upload_openfold3_test_queries.sh`.
- Sanity tests (labels, filesystem, CUDA) run via the shared reusable workflow.

### Companion CRs (separate review)
- `DLContainersInfraCDK` — adds the `openfold3` ECR repo (standard `Principal:"*"` pull policy; "private" = not mirrored to the public Gallery).
- `DLContainersReleaseLogicPython` — adds the `openfold3` framework + `0.4-cu130-amzn2023` tag scheme.

## Test Plan

PR pipeline (`pull_request` / `workflow_dispatch`): build → sanity → security (ECR scan) → SageMaker async-endpoint test. Local pre-commit (ruff, dockerfmt, actionlint, gitleaks, yaml/json) passes on all files.

## Test Result

All PR pipeline jobs pass: build, sanity, sanity-sagemaker, ECR vulnerability scan, and the SageMaker async-endpoint test. `release-gate` / `release` are correctly skipped on PR.

Endpoint test (run under network isolation, **5/5 passed**; endpoint reaches InService in ~35 min incl. warmup, then per-case warm runtimes below):
- `test_small_single` — PASSED (~85s) — 76-res protein, single request.
- `test_large_single` — PASSED (~206s) — 622-res protein, single request.
- `test_large_precomputed_msa` — PASSED (~216s) — customer-supplied precomputed MSA accepted, valid structure, no outbound network.
- `test_msa_server_rejected_under_isolation` — PASSED (~5s) — `use_msa_server=true` rejected fast with a clear error (no hang).
- `test_large_concurrent_uses_gpu_pool` — PASSED — 1-request 206s vs 4-concurrent 272s (ratio 1.32), confirming the GPU pool parallelizes rather than serializing (~4x).

A **gamma pre-production release** was also validated: the image published to the gamma registry (`028651357192`, us-west-2) with all three tags (`0.4.1-cu130-amzn2023`, `0.4-cu130-amzn2023`, dated) plus SOCI variants.

> **Note:** the companion CDK (ECR repo) and ReleaseLogic (tag config) CRs are now **deployed**, so `release: true` is enabled. PR runs still never release (the release-gate requires main-ref + an autorelease/dispatch caller); the first publish happens on the scheduled autorelease (Tue 22:30 UTC) or a manual dispatch after this PR merges to main.


malav-shastri (dismissed):

.github/workflows/openfold3.pipeline.yml
```diff
@@ -0,0 +1,147 @@
...
+          echo "should-release=${SHOULD_RELEASE}" >> $GITHUB_OUTPUT
+          echo "environment=${ENVIRONMENT}" >> $GITHUB_OUTPUT
+      - id: spec
+        if: steps.check.outputs.should-release == 'true'
+        uses: ./.github/actions/generate-release-spec
+        with:
+          config-file: ${{ inputs.config-file }}
+
+  release:
+    if: ${{ needs.release-gate.outputs.should-release == 'true' }}
```
sirutBuasai: So you'll need this change added too https://github.com/aws/deep-learning-containers/pull/6307/changes

docker/openfold3/Dockerfile
```diff
@@ -0,0 +1,117 @@
...
+#   runtime  — DLC devel base (kept: Triton/CUTLASS JIT compiles at warmup)
+
+ARG BASE_IMAGE=public.ecr.aws/deep-learning-containers/base:devel-cu130-amzn2023
+
+# =============================================================================
+# builder
+# =============================================================================
+FROM ${BASE_IMAGE} AS builder
+
+ARG MINIFORGE_VERSION=25.11.0-1
```
sirutBuasai, line 14: put miniforge_version field into the build: key in config file too.

The main idea is that any variables declared as `ARG` inside the dockerfile should be declared in the config file. Then the config file value is the actual value passed in while the dockerfile ARG values are simply fallback if someone forgets to add it.

sirutBuasai (approved):

PR #6286, opened 2026-06-24
ci: enable pull_request trigger for vLLM AL2023 PR workflow

Jyothirmaikottu:
## What

Re-enables the `pull_request:` auto-trigger for the **vLLM AL2023** PR workflow (`vllm.pr-amzn2023.yml`), which was commented out during the workflow refactor (left `workflow_dispatch`-only), and lands the two fixes the AL2023 pipeline surfaced once re-enabled.

This PR is **scoped to vLLM AL2023 only**. The vLLM Ubuntu and XGBoost trigger re-enables that were originally bundled here have been split into dedicated PRs so each framework's CI is reviewed and validated independently:
- vLLM Ubuntu → #6296
- XGBoost 3.0.5 + 3.2 → #6297

## Changes

- **`vllm.pr-amzn2023.yml`** — re-enable the `pull_request:` trigger (path filters unchanged; `workflow_dispatch` retained).
- **`vllm_server/framework_allowlist.json`** — allowlist `CVE-2026-54232`. The scanner flags vllm `0.22.1rc0` as below the `0.22.1` fix tag (an rc sorts before the final). The CVE is a dependency-confusion issue in the **upstream** vLLM project Dockerfile's `flashinfer-jit-cache` install; our `Dockerfile.amzn2023` installs it pinned from the `flashinfer.ai` index. The real version fix arrives with the 0.23.0 image bump. `review_by: 2026-07-31`.
- **`test/efa/scripts/setup_nccl_tests.sh`** — force `CXXSTD=-std=c++17` for the nccl-tests build. nccl-tests auto-selects c++17 for CUDA ≥ 13 but detects the CUDA version via `which nvcc`; the AL2023 runtime image ships no `which`, so detection returns empty, it falls back to `-std=c++14`, and the build fails against the CUDA 13 libcu++ headers NCCL pulls in (`#error libcu++ requires at least C++ 17`). Reproduced in a `nvidia/cuda:13.0.2-runtime-amzn2023` container; pinning c++17 builds all 11 perf binaries. (Shared with the Ubuntu EFA path, which already passes — Ubuntu ships `which`.)

## Notes

- Only the AL2023 trigger auto-runs after merge; it does not retroactively trigger already-open PRs.


.github/workflows/vllm.pr-amzn2023.yml
```diff
@@ -1,27 +1,26 @@
 name: "PR - vLLM AL2023"
 
 on:
-  # TEMP: Enable after all frameworks are refactored
   workflow_dispatch:
```
sirutBuasai: remove this as well. We don't want to manually trigger PR

sirutBuasai (dismissed):

sirutBuasai:
Can you also delete this line? https://github.com/aws/deep-learning-containers/blob/1833626a3851550c0b4f12dce1520e23329d641b/.github/workflows/vllm.tests-upstream.yml#L173

Honestly, I forgot why I added that line and whether the test still skips.

.github/config/model-tests/vllm-model-tests.yml
```diff
@@ -180,10 +180,12 @@ sagemaker:
...
-      # Pre-wrapped in literal single quotes so the JSON survives standard-supervisor's
-      # `" ".join(argv)` + supervisord shlex.split round-trip, which would otherwise
-      # strip the inner double quotes from --hf-overrides and break json.loads.
-      SM_VLLM_HF_OVERRIDES: '''{"architectures":["Qwen3ForSequenceClassification"],"classifier_from_token":["no","yes"],"is_original_qwen3_reranker":true}'''
+      # Plain JSON, no literal-single-quote wrapping. standard-supervisor (>=0.1.15,
+      # pinned in docker/vllm/Dockerfile.amzn2023) shlex.quote()s each argv element
+      # before the supervisord round-trip, so the inner double quotes survive intact
+      # and json.loads succeeds. Wrapping in single quotes here would pass them
+      # through verbatim and break json.loads on the leading quote.
+      SM_VLLM_HF_OVERRIDES: '{"architectures":["Qwen3ForSequenceClassification"],"classifier_from_token":["no","yes"],"is_original_qwen3_reranker":true}'
```
Jyothirmaikottu, line 188: Dropped the single-quote wrapping here. We needed it before, but the newer standard-supervisor handles the quoting for us now, so plain JSON just works. Leaving the old wrapping in would actually break it.

test/efa/scripts/nixl_disagg_pd.sh
```diff
@@ -43,7 +43,12 @@ SIDE_CHANNEL_PORT=5559
 KV_CONFIG='{"kv_connector":"NixlConnector","kv_role":"kv_both","kv_connector_extra_config":{"backends":["LIBFABRIC"],"kv_load_failure_policy":"fail"}}'
 
 # Side-channel host: needs to be the IP that the worker can reach this box on.
-SIDE_CHANNEL_HOST=$(ip -4 -o addr show scope global | awk '{print $4}' | cut -d/ -f1 | head -1)
+# The AL2023 runtime image ships no iproute2 (`ip`), so resolve the primary
+# private IP via the kernel routing table using a UDP socket — connect() on a
+# datagram socket sends no packets, it just selects the source address for the
+# default route. Python is always present in the vLLM image.
+SIDE_CHANNEL_HOST=$(python3 -c 'import socket; s=socket.socket(socket.AF_INET, socket.SOCK_DGRAM); s.connect(("8.8.8.8", 80)); print(s.getsockname()[0]); s.close()')
```
Jyothirmaikottu, line 50: The old `ip` command isn't in the AL2023 image, so this was coming back empty and the test couldn't find its own IP. Switched to a tiny Python snippet to grab it instead, since Python is always there.

test/efa/scripts/nixl_disagg_pd_decode.sh
```diff
@@ -26,8 +26,12 @@ SIDE_CHANNEL_PORT=5659
 KV_CONFIG='{"kv_connector":"NixlConnector","kv_role":"kv_both","kv_connector_extra_config":{"backends":["LIBFABRIC"],"kv_load_failure_policy":"fail"}}'
 
 # Side-channel host must be reachable from the prefill node (cross-host pull).
-# Default to the first non-loopback IPv4 address on this box.
-SIDE_CHANNEL_HOST=$(ip -4 -o addr show scope global | awk '{print $4}' | cut -d/ -f1 | head -1)
+# The AL2023 runtime image ships no iproute2 (`ip`), so resolve the primary
+# private IP via the kernel routing table using a UDP socket — connect() on a
+# datagram socket sends no packets, it just selects the source address for the
+# default route. Python is always present in the vLLM image.
+SIDE_CHANNEL_HOST=$(python3 -c 'import socket; s=socket.socket(socket.AF_INET, socket.SOCK_DGRAM); s.connect(("8.8.8.8", 80)); print(s.getsockname()[0]); s.close()')
```
Jyothirmaikottu, line 33: Same fix as nixl_disagg_pd.sh — `ip` isn't in the AL2023 image so this came back empty. Grabbing the IP via Python instead.

test/efa/scripts/setup_nccl_tests.sh
```diff
@@ -41,7 +41,17 @@ if [ ! -f "${NCCL_HOME}/include/nccl.h" ]; then
         apt-get install -y --no-install-recommends libnccl-dev
     elif command -v dnf >/dev/null 2>&1; then
         echo "Installing libnccl-devel from dnf..."
-        dnf install -y --setopt=install_weak_deps=False libnccl-devel
+        # The image symlinks /usr/bin/python3 to the venv, so dnf's shebang
+        # interpreter lacks the dnf module. Run dnf with the system python.
+        SYS_PY=""
```
Jyothirmaikottu, line 46: `dnf` was blowing up because our python3 points at the venv, which doesn't know about dnf. So we just point it at the real system python to install the package.

test/efa/scripts/setup_nccl_tests.sh
```diff
@@ -87,12 +97,13 @@ SM=$(nvidia-smi --query-gpu=compute_cap --format=csv,noheader | head -1 | tr -d
...
 NVCC_GENCODE="-gencode=arch=compute_${SM},code=sm_${SM}"
 
+# Pin C++17: nccl-tests' CUDA>=13 auto-detect uses `which`, absent on AL2023, so it wrongly falls back to c++14 and fails on CUDA 13 libcu++ headers.
 cd /tmp
 rm -rf nccl-tests
 git clone --depth 1 https://github.com/NVIDIA/nccl-tests.git
 cd nccl-tests
 make -j1 MPI=1 MPI_HOME=/opt/amazon/openmpi NCCL_HOME="${NCCL_HOME}" \
-    CUDA_HOME=/usr/local/cuda NVCC_GENCODE="${NVCC_GENCODE}"
+    CUDA_HOME=/usr/local/cuda NVCC_GENCODE="${NVCC_GENCODE}" CXXSTD="-std=c++17"
```
Jyothirmaikottu, line 106: Forcing C++17 here. nccl-tests tries to figure this out on its own but can't on AL2023, so it guessed wrong and the CUDA 13 build failed. Pinning it fixes that.

test/security/data/ecr_scan_allowlist/vllm_server/framework_allowlist.json
```diff
@@ -78,5 +78,20 @@
         "vulnerability_id": "CVE-2026-42504",
         "reason": "go/stdlib 1.24.11 statically linked inside mooncake libetcd_wrapper.so, cannot be patched independently of mooncake-transfer-engine rebuild",
         "review_by": "2026-08-30"
+    },
+    {
+        "vulnerability_id": "RUSTSEC-2026-0185",
```
Jyothirmaikottu, line 83: Allowlisting these three for now — they're all baked into the upstream vLLM wheel, so we can't patch them on our side until upstream ships a fix. The version bump will clear them out.

sirutBuasai (approved):

PR #6298, opened 2026-06-26
feat(vllm-omni): accept application/json on video routes

Jyothirmaikottu:
Restores JSON->multipart conversion in the SageMaker routing middleware for the form-data video routes (`/v1/videos`, `/v1/videos/sync`), so callers can send `application/json` instead of multipart. The proven multipart path is unchanged (byte-for-byte passthrough).

Stays on vllm-omni 0.21.0rc1 — no framework bump. Bumps `dlc_minor_version` 3 -> 4.

### DIFF
--- a/.github/workflows/vllm-omni.tests-model.yml
+++ b/.github/workflows/vllm-omni.tests-model.yml
@@ -59,7 +59,7 @@ jobs:
           uv venv --python 3.12
           source .venv/bin/activate
           uv pip install pyyaml
-          python3 scripts/ci/parse_model_config.py --config .github/config/model-tests/vllm-omni-model-tests.yml --runner-type codebuild-fleet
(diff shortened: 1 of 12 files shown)


.github/workflows/vllm-omni.tests-model.yml
```diff
@@ -59,7 +59,7 @@ jobs:
           uv venv --python 3.12
           source .venv/bin/activate
           uv pip install pyyaml
-          python3 scripts/ci/parse_model_config.py --config .github/config/model-tests/vllm-omni-model-tests.yml --runner-type codebuild-fleet
```
Jyothirmaikottu, line 62: multipart test is not needed on ec2 so we can add this flag to control that

test/vllm/scripts/vllm_regression_test.sh
```diff
@@ -9,5 +9,6 @@ fi
 
 # Regression Test # 7min
 cd vllm_source/tests
-uv pip install $UV_FLAGS modelscope
```
Jyothirmaikottu, line 12: upstream vllm regression which also blocks our release.

Yadan-Wei (approved):

PR #6345, opened 2026-07-06
test(vllm): add MiniCPM5-1B model tests (smoke-test + sagemaker)

ytp327:
## What

Adds **MiniCPM5-1B** (dense 1B `LlamaForCausalLM`, bf16) to the vLLM model-test config in both suites:

- **smoke-test** (EC2): served on `x86-g6xl-runner`, `--max-model-len 8192 --dtype bfloat16`. Uses the default smoke test (`/v1/completions` non-empty check).
- **sagemaker**: `ml.g6.xlarge` endpoint, `amzn2023` image, `SM_VLLM_MAX_MODEL_LEN=8192`, with a `/v1/chat/completions` test case validated via `json_field:choices.0.message.content`.

## Why

MiniCPM5-1B is the first plain chat/text-generation model in the SageMaker suite (existing entries are ASR / rerank). Standard architecture, loads with plain `vllm serve` — no `--trust-remote-code`.

## Validation

Validated offline on a GPU devbox (L4) against both AL2023 DLC images:

- `server-cuda-v2` (EC2): `/v1/chat/completions` -> HTTP 200, correct output, `finish_reason: stop`
- `server-sagemaker-cuda-v2` (SageMaker): `/invocations` -> HTTP 200, correct output, `finish_reason: stop`

Both at `--max-model-len 8192` (model native max is 131072; capped to keep KV cache modest on the L4).

## Model artifact

Uploaded to `s3://dlc-cicd-models/llm-models/minicpm5-1b.tar.gz` (matches the `s3_model` reference).

## Note

Opened as a same-repo branch (not a fork) so the image-preflight step can resolve account/region and the model-test matrix actually runs (fork PRs skip these).

### DIFF
--- a/.github/config/model-tests/vllm-model-tests.yml
+++ b/.github/config/model-tests/vllm-model-tests.yml
@@ -20,6 +20,17 @@ smoke-test:
       fleet: "x86-g6xl-runner"
       extra_args: "--tensor-parallel-size 1 --max-model-len 4096 --dtype bfloat16"
 
+    # MiniCPM5-1B — dense 1B LlamaForCausalLM, bf16 (~2GB), no custom code.
+    # Standard architecture, loads with plain `vllm serve` (no --trust-remote-code).
+    # Hybrid reasoning emits <think> blocks; default smoke test only checks for a
+    # non-empty /v1/completions response, so no reasoning parser is required here.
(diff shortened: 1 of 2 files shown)


ytp327:
All tests passed:
image

Yadan-Wei (approved):

PR #6349, opened 2026-07-06
ci(vllm): add nemotron-nano-12b-v2 + split sagemaker endpoint config with workflow-level filter

Yadan-Wei:
## Summary
- Add **NVIDIA Nemotron-Nano-12B-v2** (Mamba2-Transformer hybrid, 12B bf16) to the vLLM smoke-test matrix on `x86-g6exl-runner` (L40S 48GB).
- Split the `sagemaker:` section of `vllm-model-tests.yml` into a dedicated `vllm-sagemaker-endpoint-tests.yml` so smoke-test and endpoint-test configs evolve independently.
- Gate SageMaker endpoint tests by image type at the **workflow preflight** step (mirrors sglang's model-test pattern), using `required_image_pattern` matched against the image config YAML — not the image URI.

## Nemotron verification
Verified end-to-end on a g6e.2xlarge (L40S 48GB) with `public.ecr.aws/deep-learning-containers/vllm:server-cuda-v2` (vllm 0.24.0+amzn2023 — well above the required `>=0.10.1`):
- Loads as `NemotronHForCausalLM`; Mamba2 SSD Triton kernels warm up cleanly.
- Steady-state GPU memory ~42 GiB / 46 GiB — flat under 8-way concurrent load (vLLM pre-allocates the KV pool).
- KV cache pool: 369K tokens → ~45x concurrency at 4096-token requests.
- Both `/think` (reasoning-on) and `/no_think` (reasoning-off) modes return correct output.
- Model card recommends `--mamba_ssm_cache_dtype float32` (skipping it degrades accuracy) and `--trust-remote-code` for the custom `configuration_nemotron_h.py` — both set in `extra_args`.

Note: startup logs a soft `Using default Mamba SSU config. Performance might be sub-optimal!` — the vLLM package doesn't ship an L40S-tuned kernel config file. Correctness is unaffected; possibly worth pursuing separately if perf becomes a concern.

## Model artifact
Uploaded to `s3://dlc-cicd-models/llm-models/nemotron-nano-12b-v2.tar.gz` (15.3 GiB) — matches the `s3_model` field in the new smoke-test entry.

## SageMaker endpoint tests: split + image-based gating
Two changes work together:

**File split.** Extracted the `sagemaker:` block from `vllm-model-tests.yml` into a new `vllm-sagemaker-endpoint-tests.yml`. The file was renamed to make its purpose explicit — the sibling smoke-test file also runs SageMaker-image tests, so plain "sagemaker-tests" was ambiguous. Both files own their own `s3_prefix` / `test_fixtures_prefix`.

**Workflow-level filter, sglang-style.** SageMaker entries carry `required_image_pattern: {os_version: amzn2023}`. Filtering happens in the `vllm.tests-sagemaker.yml` preflight step, not inside pytest:
1. Preflight runs `scripts/ci/parse_model_config.py --section sagemaker --image-config <image config>`.
2. The parser reads the **image config YAML** (not the image URI — `os_version` isn't reliably part of image URIs), flattens `metadata:` into `{os_version: "amzn2023", framework: "vllm_server", ...}`, and applies `_model_matches_image` to each entry.
3. Emits `matrix=[...]` — full list on amzn2023, `[]` on ubuntu.
4. The `endpoint-test` job is gated on `fromJson(matrix)[0] != null`, so on ubuntu the job is skipped entirely instead of pytest-collecting-and-skipping each entry.

**Parser update.** `parse_model_config.py` now handles a section that's a flat list (like `sagemaker:`) in addition to the existing dict-of-runner-buckets shape (like `smoke-test:`). Smoke-test output is byte-for-byte identical to before.

## Consumers updated
- `test_sm_model_serving.py`: `CONFIG_PATH` points at the new file; the test is now agnostic to image type (the workflow decides what runs).
- `vllm.pr-amzn2023.yml` / `vllm.pr-ubuntu.yml`: add the new file to top-level `paths:` and to the `sagemaker-test-change` filter, so edits to endpoint entries fire the sagemaker suite (not the smoke suite).
- `vllm.tests-sagemaker.yml`: adds the preflight parse step and the matrix gate.
- No changes needed to `benchmark_report.py` or the autorelease workflows.

## Commits
1. `ci(vllm): add nemotron-nano-12b-v2 to smoke tests`
2. `ci(vllm): split sagemaker endpoint tests into vllm-sagemaker-tests.yml`
3. `ci(vllm): rename vllm-sagemaker-tests to vllm-sagemaker-endpoint-tests`
4. `ci(vllm): drop required_image_pattern from sagemaker endpoint tests`
5. `ci(vllm): filter sagemaker tests by image via required_image_pattern`
6. `ci(vllm): gate sagemaker tests via workflow preflight, not in pytest`

## Test plan
- [ ] PR CI runs the `sagemaker-test-change` and `model-test-change` filters correctly (verify on this PR — it touches both files).
- [ ] Nemotron smoke test entry launches on `x86-g6exl-runner` and passes the smoke-test script.
- [ ] Existing SageMaker endpoint tests (voxtral / granite-speech / qwen3-reranker / minicpm5-1b) still discover their configs via the renamed path.
- [ ] On an amzn2023 sagemaker image config, the preflight parse step emits a 6-entry matrix and the endpoint-test job runs.
- [ ] On an ubuntu sagemaker image config, the preflight parse step emits `matrix=[]` and the endpoint-test job is skipped entirely (not just each test pytest-skipped).


Yadan-Wei:
Because we only have SageMaker middleware for al23 image, so we cannot run model that needs multipart/form API on ubuntu image, so this PR still skip them for ubuntu image.

Jyothirmaikottu (approved):

PR #6365, opened 2026-07-08
feat(pytorch): add PyTorch 2.13 currency scaffold

bhanutejagk:
# PyTorch 2.13 currency

Ships the PyTorch 2.13.0 DLC image (EC2 and SageMaker variants, GPU + CPU) as part of the ongoing currency cadence.

## What's shipped

- **PyTorch 2.13.0** with torchvision 0.28.0, torchaudio 2.11.0
- **CUDA 13.3.0** base (bumped from initial 13.0.2 to get cuBLAS 13.3+, which TransformerEngine 2.17.0 requires for `cublasLtGroupedMatrixLayoutInit_internal`)
- Dep bumps: NCCL 2.30.7-1, DeepSpeed 0.19.2, TransformerEngine 2.17.0, EFA 1.49.0, gdrcopy 2.6
- flash-attn 2.8.3 held (no torch-2.13 prebuilt wheel upstream — served from S3 wheel-cache after initial source build)
- CVE fixes: soupsieve 2.8.4, click 8.3.3, pillow 12.3.0
- **Note on torch/CUDA versioning:** torch 2.13.0 wheels come from PyTorch upstream's `cu130` index — upstream doesn't publish `cu133` wheels. Running cu130-built torch on a cu133 base is officially supported per NVIDIA's minor-version compatibility guarantee within CUDA 13.x.

## Infrastructure

- Routes pytorch builds to a dedicated `x86-pytorch-build-runner` CodeBuild fleet (m6a.12xlarge, per an internal infra CR)
- Wheel-cache S3 key now includes torch major.minor to prevent ABI reuse across torch versions
- EFA installer passes `--disable-ngc` on 1.48+ (required to install libnccl-net.so on NGC base images)
- Moved GPU-import tests (`test_import_gpu[flash_attn|transformer_engine]`) from `unit/` to `single_gpu/` — TE 2.17.0 eagerly dlopens libcuda.so.1 at import, which is only mounted on GPU-fleet workers

## Known tech debt

`NVIDIA_DISABLE_REQUIRE=true` bypass on `pytorch.tests-single-gpu.yml`: works around a stale `nvidia-container-toolkit` (1.19.1) that rejects the `com.nvidia.cuda.requires=cuda>=13.3` OCI label. The runner's actual driver (580.159.04) exceeds CUDA 13.3's declared minimum (580.65), so the check is a false positive. Fix landed in nvidia-container-toolkit v1.20.0-rc.1 (2026-07-09); waiting on the upstream base AMI to consume the newer toolkit. Bypass will be reverted in a follow-up PR once the base image updates.

## Testing

CI on this PR exercises: unit-test, sanity-test, single-gpu-test (flash-attn func + backward, TE Linear forward/backward, SDPA backend), telemetry-test, sagemaker-test, security-test (ECR CVE scan), EFA-test. All 2.13 test suites pass on both `ec2-cuda` and `sagemaker-cuda` variants.


docker/pytorch/2.13/cuda/pyproject.toml
```diff
@@ -0,0 +1,59 @@
...
+    "s3fs>=2026.2.0",
+    "aiohttp>=3.14.0",
+    "tornado>=6.5.6",  # CVE GHSA-3x9g-8vmp-wqvf, GHSA-mgf9-4vpg-hj56 (transitive)
+    "cryptography>=48.0.1",  # CVE GHSA-537c-gmf6-5ccf (transitive)
+]
+
+[tool.uv]
+
+[[tool.uv.index]]
+name = "pytorch-cu130"
```
Eren-Jeager123, line 52: Please check this. We are using cu133 right

bhanutejagk: that seems like the version of torch wheels. i looked up and [cu133 torch wheels ](https://download.pytorch.org/whl/cu133/torch/) DO NOT exist upstream. 
We bumped the CUDA base image to 13.3.0, but the pyproject still fetches torch wheels from cu130. Its like we are using cu130 torch wheels on top of a cu133 base image since they are compatable.

The base image bump to cu133 was for cuBLAS — specifically, TransformerEngine 2.17.0's `cublasLtGroupedMatrixLayoutInit_internal` requires cuBLAS 13.3+, which the newer base image provides

scripts/ci/build/pytorch_runtime/lib/fetch_wheels.sh
```diff
@@ -3,31 +3,34 @@
...
 # Usage:
 #   bash fetch_wheels.sh --dest-dir <dir> --bucket <bucket> --cuda-version <ver> \
-#     --packages "flash-attn:2.8.3,transformer-engine-torch:2.12.0"
+#     --torch-version <ver> --packages "flash-attn:2.8.3,transformer-engine-torch:2.12.0"
 #
 # Exit code: 0 if all wheels found, 1 if any cache miss.
-# Cache key: s3://<bucket>/wheels/<pkg_underscore>/<cuda_short>/<wheel_filename>
-# Example: s3://dlc-cicd-wheels/wheels/flash_attn/cu130/flash_attn-2.8.3-cp312-cp312-linux_x86_64.whl
+# Cache key: s3://<bucket>/wheels/<pkg_underscore>/<cuda_short>/<torch_short>/<wheel_filename>
+# Example: s3://dlc-cicd-wheels/wheels/flash_attn/cu130/torch213/flash_attn-2.8.3-cp312-cp312-linux_x86_64.whl
```
Eren-Jeager123, line 10: can you explain why do we need to add `torch213/ `path? I saw they are all flash attention 2.8.3, are there any differences between e.g. torch212/ and torch213/?

docker/pytorch/Dockerfile.cuda
```diff
@@ -61,7 +61,7 @@ COPY docker/pytorch/${FRAMEWORK_SHORT_VERSION}/wheel[s]/ /tmp/wheels/
 RUN CUDNN_HOME=$(python -c "import nvidia.cudnn; print(nvidia.cudnn.__path__[0])") && \
   NCCL_HOME=$(python -c "import nvidia.nccl; print(nvidia.nccl.__path__[0])") && \
   cp ${CUDNN_HOME}/include/*.h /usr/local/cuda/include/ && \
-  cp ${NCCL_HOME}/include/*.h /usr/local/cuda/include/ && \
+  cp -a ${NCCL_HOME}/include/. /usr/local/cuda/include/ && \
```
Eren-Jeager123, line 64: could you explain?

.github/workflows/pytorch.tests-single-gpu.yml
```diff
@@ -64,7 +64,16 @@ jobs:
...
+          # TODO: Remove NVIDIA_DISABLE_REQUIRE=true once the runner AMI ships
+          # nvidia-container-toolkit >= 1.20. The current toolkit (1.19.1) has
+          # a stale semver parser and compat table that reject the
+          # `com.nvidia.cuda.requires=cuda>=13.3` label, even though the
+          # runner's actual driver (580.159.04) exceeds CUDA 13.3's declared
+          # minimum (580.65). Upstream fix landed in nvidia-container-toolkit
+          # v1.20.0-rc.1 (2026-07-09). To remove: delete this comment + the
+          # `-e NVIDIA_DISABLE_REQUIRE=true` flag, then verify CI passes on
+          # the updated runner AMI.
+          CONTAINER_ID=$(docker run -d --rm --gpus all -e NVIDIA_DISABLE_REQUIRE=true --shm-size=2g \
```
Eren-Jeager123, line 70: is `NVIDIA_DISABLE_REQUIRE=true` still required now?

docker/pytorch/Dockerfile.cuda
bhanutejagk: Before commit 51a936cc, this line was cp ${NCCL_HOME}/include/*.h ... — flat glob, top-level .h files only. torch 2.13's NCCL wheel added a nccl_device/ subdirectory containing coop.h that TE 2.17 needs at build time. The flat glob missed it → TE source build failed with `#include <nccl_device/coop.h>: file not found`
  
  Changed to cp -a ${NCCL_HOME}/include/. /usr/local/cuda/include/ — recursive copy preserving directory structure. Now nccl_device/coop.h and any future subdirs get through.

scripts/ci/build/pytorch_runtime/lib/fetch_wheels.sh
bhanutejagk: flash-attn's C++ code calls PyTorch internal functions. Between torch 2.12 and 2.13, PyTorch renamed one of those functions (materialize_cow_storage → materialize_cow).
  So:

  - Compiled with torch 2.12 → .so asks for `materialize_cow_storage`
  - Compiled with torch 2.13 → .so asks for `materialize_cow`

  Both wheels are named flash_attn-2.8.3-*.whl, but they're not interchangeable — install the torch-2.12-built wheel in a torch-2.13 image and import flash_attn fails with undefined symbol.

  Before torch213/ in the key, cu130's cache slot held ONE wheel — and it kept getting overwritten across torch bumps, breaking builds. Adding torch213/ gives each torch version its own S3 path so wheels can't collide

.github/workflows/pytorch.tests-single-gpu.yml
bhanutejagk: The flag is a temporary bypass for a stale-parser bug in `nvidia-container-toolkit 1.19.1` (which the current AWS DLAMI still ships). That toolkit version rejects the `com.nvidia.cuda.requires=cuda>=13.3` label with a "legacy mode" fallback error, even though our fleet's actual driver (580.159.04) far exceeds CUDA 13.3's stated minimum (580.65). The upstream fix landed in toolkit v1.20.0 but no GA, so no DLAMI has consumed it yet.

This is early-warning if driver becomes too old but we use the latest ones. no harm for us

bhanutejagk: But we have to remove the flag after there is a upstream release for [nvidia-container-toolkit  v1.20.0](https://github.com/NVIDIA/nvidia-container-toolkit/releases/tag/v1.20.0) since it is still in[ rc ](https://github.com/NVIDIA/nvidia-container-toolkit/releases) and our DLAMI consumes it

Eren-Jeager123 (approved):

PR #6392, opened 2026-07-13
add: huggingface-vllm v0.25

dwarez:
## Purpose
Upgrading huggingface-vllm image to vLLM v0.25

## PR Checklist

### DIFF
--- a/.github/config/image/huggingface-vllm/sagemaker.yml
+++ b/.github/config/image/huggingface-vllm/sagemaker.yml
@@ -12,7 +12,7 @@ metadata:
   device_type: "gpu"
   job_type: "general"
   contributor: "huggingface"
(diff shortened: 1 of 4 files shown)


.github/config/image/huggingface-vllm/sagemaker.yml
sallyseok: Kinda late but just noticed this, it should be changed to one of your (future) prod tags so like `huggingface-vllm:0.22-gpu-py312-cu130-ubuntu22.04` would be a good one because it's short. This field is used when a PR only makes changes to the test code and not the dockerfile source code -- the CI will test against the prod image instead of building one from scratch.

sallyseok (approved):

PR #6396, opened 2026-07-13
Add WhisperX ASR container (EC2 + SageMaker, AL2023 GPU)

Yadan-Wei:
## Purpose

Onboard a new **WhisperX ASR container** to AWS Deep Learning Containers (AL2023, GPU, CUDA 12.8, Python 3.12), for both **EC2** and **SageMaker** inference.

WhisperX is a batch CLI, not a server, so this PR adds the serving layer it lacks: a FastAPI/uvicorn app (`server.py`) that keeps the model resident across requests and exposes an OpenAI-compatible transcription API plus the SageMaker `/invocations` + `/ping` contract.

**Container**
- `docker/whisperx/Dockerfile.amzn2023` — multi-stage AL2023 GPU build (builder → runtime), two targets (`whisperx-ec2-amzn2023`, `whisperx-sagemaker-amzn2023`).
- Pipeline: pyannote VAD → faster-whisper transcription → wav2vec2 word alignment → pyannote diarization.
- Baked pyannote diarization weights (pulled from the CI models bucket in `pre_build.sh`); Whisper + per-language aligners cache lazily at runtime.
- CUDA forward-compat shim (`cuda_compat.sh`) for older-driver SageMaker hosts.

**Server (`scripts/docker/whisperx/server.py`)**
- Routes: `POST /v1/audio/transcriptions` (OpenAI-compatible), `POST /invocations` (SageMaker), `GET /ping` (readiness), `GET /v1/models`.
- One Whisper model per container, pinned at launch via `WHISPERX_DEFAULT_MODEL`; the request `model` field is accepted and ignored (OpenAI-compat no-op).
- **Bring-your-own-model:** the SageMaker entrypoint auto-detects a model staged at `/opt/ml/model` (from `ModelDataUrl`) and serves it with no env var required (mirrors the vLLM entrypoint's ladder: explicit `WHISPERX_DEFAULT_MODEL` override → `/opt/ml/model` → image default). faster-whisper loads a directory path directly, so a flat model dir works.
- Word-level alignment with a bounded LRU of per-language wav2vec2 aligners (evict-before-load keeps GPU memory bounded).
- Optional diarization (`diarize`, `min_speakers`, `max_speakers`) and subtitle output (`srt`/`vtt` via WhisperX's own writer, byte-identical to the CLI).
- All decoding/VAD knobs are launch-time env config (`WHISPERX_*`).

**Serving robustness**
- Inference is **serialized to 1** (a single WhisperX pipeline is not concurrency-safe), fronted by a hard admission cap (default 1 running + 2 queued) that sheds overload with 503 before reading the body; blocking work is offloaded off the event loop so `/ping` stays responsive.
- Passive readiness: a fatal CUDA fault flips `/ping` to 503 so orchestration can drain the host; transient OOM does not.
- Streamed, size-capped uploads (413 on oversize).
- `task=translate` disables word alignment/diarization (mirrors upstream — translated English text cannot be aligned to the source-language audio); a best-effort word request degrades to segment-level, an explicit diarize request returns 422.

**Dependency management**
- Dependencies pinned in `docker/whisperx/pyproject.toml` + `uv.lock`, installed via `uv sync --frozen` (Ray/PyTorch DLC pattern). torch/torchaudio resolve from the PyTorch cu128 index; all else from PyPI. The image config `build:` block carries only python/cuda/dlc versions.

**CI**
- PR + pipeline + EC2/SageMaker/unit-test workflows and image configs under `.github/`.
- ECR scan allowlist for contract CVEs.

## Test Plan

- **Unit (CPU, no GPU):** `whisperx.tests-unit.yml` runs `test_server_options.py`, `test_transcribe_behavior.py`, `test_subtitle_format.py`, `test_sagemaker_entrypoint_model_resolution.py` with torch/whisperx/fastapi stubbed — covers env→options parsing, admission control (503) + upload cap (413), serialized inference, passive readiness + fatal-vs-transient CUDA classification, alignment degrade paths (ValueError / NotImplementedError / KeyError), the translate guard, subtitle-knob alignment forcing, LRU evict-before-load, boundary validation (422 / unknown-format 400), and the entrypoint's `/opt/ml/model` auto-detect ladder.
- **EC2 functional (GPU):** container up → `/ping` → transcription, word timestamps, diarization, subtitle formats, model-config (served id via `/v1/models`, served-name alias), and offline bring-your-own-model load (`test_ec2_byo_model.py`: `HF_HUB_OFFLINE=1` + a mounted non-default model, with a no-mount negative control that proves no Hub fallback).
- **SageMaker functional (GPU):** real endpoint deploy → sync + async `/invocations`, diarization, env-var model-config plumbing through `ContainerDefinition`, and an offline `ModelDataUrl` bring-your-own-model endpoint (`test_sm_model_data.py`).

## Test Result

- Unit suite: **79 passed** locally.
- `uv sync --frozen` verified consistent against the lock (139 packages resolved).
- Bring-your-own-model verified end to end on a GPU instance: the entrypoint auto-detected a `large-v3` model staged at `/opt/ml/model`, loaded it offline (`HF_HUB_OFFLINE=1`), and transcribed; the no-model negative control failed startup as expected.
- EC2 + SageMaker functional suites run in CI on GPU runners.

<details>

## PR Checklist

</details>


.github/workflows/whisperx.pipeline.yml
```diff
@@ -0,0 +1,215 @@
...
+      run-unit-test:
+        description: "Run CPU-only unit tests"
+        required: false
+        type: boolean
+        default: true
+      run-security-test:
+        description: "Run security tests"
+        required: false
+        type: boolean
+        default: true
```
sirutBuasai, line 33: FYI, if this is default as true, we need to set this to false explicitly in the release pipeline whenever you create that one.

scripts/docker/whisperx/cuda_compat.sh
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env bash
```
sirutBuasai: why cant we use existing cuda_compat script?

scripts/docker/whisperx/server.py
```diff
@@ -0,0 +1,769 @@
...
+#     WHISPERX_DEFAULT_MODEL, WHISPERX_COMPUTE_TYPE, WHISPERX_BATCH_SIZE,
+#     WHISPERX_SERVED_MODEL_NAME, ...
+def _env_bool(name: str, default: bool = False) -> bool:
+    val = os.environ.get(name)
+    if val is None or not val.strip():
+        return default
+    return val.strip().lower() in {"1", "true", "yes", "on"}
+
+
+def _build_asr_options() -> dict[str, Any]:
```
sirutBuasai, line 80: So all settings are configured at container launch time via env variables right? If so, we'll need a pretty thorough documentation after release

.github/workflows/whisperx.pipeline.yml
Yadan-Wei: Thanks for the reminder. I will pay attention when creating the release pipeline.

scripts/docker/whisperx/cuda_compat.sh
Yadan-Wei: I saw our cuda_compat script is under pytorch directly so the agent create a new one, I can add our existing one to whipserX container to make them consistent.

scripts/docker/whisperx/server.py
Yadan-Wei: Yes, most of parameters used by whisperx command line become environment variable in container launch and I will create a user guide to list all of them.

sirutBuasai (dismissed):

jinyan-li1 (approved):

PR #6417, opened 2026-07-16
Bump nvidia-cudnn-cu12 to 9.24.0.43 for TF 2.21 GPU image

bhanutejagk:
## Problem
TF 2.21 GPU training jobs on any CUDA GPU fail with:

```
Autotuner could not find any supported configs for HLO:
  %cudnn-conv-bias-activation.7 = ...
  custom_call_target="__cudnn$convBiasActivationForward"
  backend_config={..., "device_type":"DEVICE_TYPE_INVALID"}
```

This affects any workload using `Conv2D` layers with `model.fit()` — a
common CNN training pattern. Reproduced on:
- Tesla T4 (compute capability 7.5)
- NVIDIA A10G (compute capability 8.6)
- NVIDIA L4 (compute capability 8.9)

## Root cause
XLA in the tensorflow 2.21.0 wheel emits an HLO with
`backend_config.device_type: DEVICE_TYPE_INVALID` for the fused
`cudnn-conv-bias-activation` custom-call. cuDNN 9.5.1 cannot service
this call pattern and returns no supported configs, causing the
autotuner to fail before training starts.

## Fix
Bump the pinned cuDNN wheel from 9.5.1.17 to 9.24.0.43 in the TF 2.21
CUDA pyproject.toml, and regenerate uv.lock so `uv sync --frozen`
picks up the new wheel.

## Verification
End-to-end verification on an NVIDIA A10G devbox:
- Built the TF 2.21 CUDA image with the pin change.
- Confirmed `nvidia-cudnn-cu12 9.24.0.43` present in the built image.
- Ran the MNIST Conv2D `model.fit()` smoke test that fails on the
  released image.
- Result: MNIST training completes cleanly, test accuracy 0.857,
  no `DEVICE_TYPE_INVALID` or `CUDNN_STATUS_NOT_SUPPORTED` errors.

## Test plan
- [ ] CI single-gpu-test passes
- [ ] CI sagemaker-test passes
- [ ] Post-release: deep canary for tensorflow-training on x86 clears
      once the rebuilt image ships

### DIFF
--- a/docker/tensorflow/2.21/cuda/pyproject.toml
+++ b/docker/tensorflow/2.21/cuda/pyproject.toml
@@ -9,7 +9,7 @@ dependencies = [
     # ships forward-ABI-compatible SONAMEs so TF 2.21's 9.3-linked binary loads
     # against 9.5 without rebuild. Dockerfile.cuda copies the .so files out into
     # /usr/local/cuda/lib64.
-    "nvidia-cudnn-cu12==9.5.1.17",
+    "nvidia-cudnn-cu12==9.24.0.43",
     "nvidia-nccl-cu12==2.30.7",
     "mpi4py",
     "boto3",
--- a/docker/tensorflow/2.21/cuda/uv.lock
+++ b/docker/tensorflow/2.21/cuda/uv.lock
@@ -2057,13 +2057,13 @@ wheels = [
 
 [[package]]
 name = "nvidia-cudnn-cu12"
-version = "9.5.1.17"
+version = "9.24.0.43"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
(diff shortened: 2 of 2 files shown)


bhanutejagk:
@Yadan-Wei — this is the TF 2.21 GPU regression follow-up. Verified end-to-end on A10G devbox; would appreciate a review when you have a moment.

bhanutejagk:
Closing in favor of #6418 which is opened from the same repository so
the CI vars/secrets propagate correctly (the ECR-login step here failed
because vars.CI_AWS_ACCOUNT_ID and vars.AWS_REGION are withheld from
external-fork PRs). Same two-file diff, same verification.

PR #6430, opened 2026-07-20
[tei] initial migration: 1.9.3 CPU + GPU images and pipeline setup

jinyan-li1:
## Summary
Port of TEI 1.9.3 (CPU + GPU) into `aws/deep-learning-containers`, with DLC pipeline setup.
Dockerfiles + SageMaker test = close port of upstream `awslabs/llm-hosting-container` (`huggingface/pytorch/tei/docker/1.9.3/{cpu,gpu}/Dockerfile` + `tests/huggingface/sagemaker_dlc_test.py`).
`release: false` — build + test only.

## Divergences from upstream

**Dockerfiles**
- COPY paths → `tei-src/` staging dir (populated by `pre_build.sh` before docker build)
- Dropped `--mount=type=secret` (needs newer BuildKit) and `sccache` (no persistent backend in our CI; buildx layer cache is enough)
- Added `LABEL dlc_major_version="1"` (required by DLC sanity gate)
- Added `ARG FRAMEWORK / FRAMEWORK_VERSION / CACHE_REFRESH` stubs
- Added trailing `apt-get upgrade` in the final stage for CVE freshness

**SageMaker test**
- Dropped TGI branch (deprecated, out of scope)
- GPU instance: `ml.g6.4xlarge` (upstream: `ml.g6.12xlarge` — same L4 GPU / compute cap 8.9, single-GPU is plenty, cheaper, generally better capacity)
- CPU instance: `ml.m5.xlarge` (upstream had `ml.g6.12xlarge` — GPU instance for CPU test, looks like a copy-paste bug)
- Explicit `inference_ami_version=al2-ami-sagemaker-inference-gpu-3-1` (driver 550, matches CUDA 12.9 container)
- All models pre-staged in S3 at `s3://dlc-cicd-models/tei-models/` (avoids HF Hub cold-start)
- Multi-image gating preserved from upstream; TEI is currently the only image
- Dropped `SM_NUM_GPUS` env — TEI's router auto-detects via `nvidia-smi`

**Net-new DLC setup** (not in upstream)
- `.github/config/image/tei/sagemaker-1.9.3-{cpu,gpu}.yml`, `.github/workflows/tei.*.yml`, `scripts/ci/build/tei/pre_build.sh`, `test/security/data/ecr_scan_allowlist/tei/framework_allowlist.json`
- Sanity workflow opts TEI out of the OSS-compliance step

## Test plan
- [x] cpu build clean
- [x] gpu build clean
- [x] sagemaker E2E green on cpu
- [x] sagemaker E2E green on gpu
- [x] security scan clean or fully allowlisted
- [x] `release: false` verified in both configs
- [x] All above tests passed: https://github.com/aws/deep-learning-containers/pull/6430/commits/0285ea08dac0629e0936b30c68c3fc8e4c51d925


.github/config/image/tei/sagemaker-1.9.3-cpu.yml
```diff
@@ -0,0 +1,28 @@
...
+  framework: "tei"
+  framework_version: "1.9.3"
+  # CPU runtime base is debian:bookworm-slim; os_version preserves the customer-facing tag suffix.
+  os_version: "ubuntu24.04"
+  customer_type: "sagemaker"
+  platform: "sagemaker"
+  arch_type: "x86"
+  device_type: "cpu"
+  job_type: "inference"
+  prod_image: "tei-cpu:2.0.1-tei1.9.3-cpu-py310-ubuntu24.04"
```
alvarobartt, line 15: Might be a great time to drop the Python tags which are misleading since the container doesn't come with Python installed.

.github/config/image/tei/sagemaker-1.9.3-cpu.yml
```diff
@@ -0,0 +1,28 @@
...
+
+build:
+  dockerfile: "docker/tei/1.9.3/cpu/Dockerfile"
+  target: "sagemaker"
+  python_version: "3.10"
```
alvarobartt, line 20: Same here, let's make this optional and drop it.

.github/config/image/tei/sagemaker-1.9.3-gpu.yml
```diff
@@ -0,0 +1,28 @@
+image:
+  name: "sagemaker-tei-gpu"
+  description: "SageMaker TEI 1.9.3 GPU"
```
alvarobartt, line 3: Not sure you intend to add support for other accelerators, but might be great to name this with NVIDIA instead, see:


```suggestion
image:
  name: "sagemaker-tei-nvidia"
  description: "SageMaker TEI 1.9.3 NVIDIA"
```

.github/config/image/tei/sagemaker-1.9.3-gpu.yml
```diff
@@ -0,0 +1,28 @@
...
+  platform: "sagemaker"
+  arch_type: "x86"
+  device_type: "gpu"
+  job_type: "inference"
+  prod_image: "tei:2.0.1-tei1.9.3-gpu-py310-cu129-ubuntu24.04"
+
+build:
+  dockerfile: "docker/tei/1.9.3/gpu/Dockerfile"
+  target: "sagemaker"
+  python_version: "3.10"
```
alvarobartt, line 19: Same here, Python is misleading and IMO I'd rather name it `nvidia` instead of `gpu`

```suggestion
  prod_image: "tei:2.0.1-tei1.9.3-gpu-cu129-ubuntu24.04"

build:
  dockerfile: "docker/tei/1.9.3/gpu/Dockerfile"
  target: "sagemaker"
```

.github/config/image/tei/sagemaker-1.9.3-gpu.yml
```diff
@@ -0,0 +1,28 @@
...
+metadata:
+  framework: "tei"
+  framework_version: "1.9.3"
+  os_version: "ubuntu24.04"
+  customer_type: "sagemaker"
+  platform: "sagemaker"
+  arch_type: "x86"
+  device_type: "gpu"
+  job_type: "inference"
+  prod_image: "tei:2.0.1-tei1.9.3-gpu-py310-cu129-ubuntu24.04"
```
alvarobartt, line 14: Should we go for the full name instead i.e., `text-embeddings-inference` other than the shortened `tei` which might be less familiar and more confusing to users?

docker/tei/1.9.3/cpu/entrypoint.sh
```diff
@@ -0,0 +1,13 @@
...
+    echo "HF_MODEL_ID must be set"
+    exit 1
+fi
+export MODEL_ID="${HF_MODEL_ID}"
+
+if [[ -n "${HF_MODEL_REVISION}" ]]; then
+    export REVISION="${HF_MODEL_REVISION}"
+fi
+
+exec text-embeddings-router --port 8080 --json-output
```
alvarobartt, line 13: So you don't intent to or don't support argument forwarding, right?

jinyan-li1:
Thanks for the review! The current live tags (`2.0.1-tei1.9.3-{gpu,cpu}-py310-cu129-ubuntu24.04`) are in customer use today, so this migration preserves them byte-for-byte to avoid breaking `docker pull`. Fully agree `py310` and `gpu` are both misleading here — we can consider dropping them in the next version bump where a tag change is expected. Following up on the entrypoint question in the per-line thread.

.github/config/image/tei/sagemaker-1.9.3-cpu.yml
jinyan-li1: Keeping in sync with the live tag. We can consider dropping `py310` in the next version bump.

.github/config/image/tei/sagemaker-1.9.3-cpu.yml
jinyan-li1: Drives the `-py310-` segment in the tag template — we can consider dropping in the next version bump.

.github/config/image/tei/sagemaker-1.9.3-gpu.yml
jinyan-li1: Same as the `py310` / `gpu` discussion — we can consider in the next version bump. Keeping this PR to a minimal port.

.github/config/image/tei/sagemaker-1.9.3-gpu.yml
jinyan-li1: Same tag-preservation reasoning — live tag is `2.0.1-tei1.9.3-gpu-py310-cu129-ubuntu24.04`. We can consider both `py310` and `gpu`→`nvidia` in the next version bump.

.github/config/image/tei/sagemaker-1.9.3-gpu.yml
jinyan-li1: Prefer keeping `tei` since it matches the ECR repo names customers pull from (`tei` / `tei-cpu`). Aligning the framework key with the repo name keeps things easy to trace. Happy to reconsider if you feel strongly.

docker/tei/1.9.3/cpu/entrypoint.sh
jinyan-li1: This is a verbatim copy of the upstream [`entrypoint.sh`](https://github.com/awslabs/llm-hosting-container/blob/main/huggingface/pytorch/tei/docker/1.9.3/cpu/entrypoint.sh) (no arg forwarding today) — happy to defer to HF on this one.

Adding `"$@"` would be backward compatible: with no args passed, behavior is identical, so it could ship under the current tag safely. Let me know if you'd like it added.

jinyan-li1: looked into this more — sagemaker runs containers as `docker run <image> serve` ([docs](https://docs.aws.amazon.com/sagemaker/latest/dg/your-algorithms-inference-code.html)), so `$1` would be `serve`. `text-embeddings-router` is HF's rust binary (built via `cargo` in the Dockerfile) and it uses clap with no positional args, so forwarding via `"$@"` would crash it on the sagemaker path.

gonna keep it [verbatim to upstream](https://github.com/awslabs/llm-hosting-container/blob/main/huggingface/pytorch/tei/docker/1.9.3/cpu/entrypoint.sh) for now since we're preserving the current tag byte-for-byte. if HF wants arg forwarding to work here, would need an upstream change to the router (accept a `serve` positional, or ignore unknown ones) — happy to align once that lands or we can revisit for the next version upgrade.

jinyan-li1:
Update: synced with @alvarobartt offline and keeping this PR as-is & migration-only, the nits will be addressed in upcoming version upgrades.

Jyothirmaikottu (approved):

PR #6434, opened 2026-07-21
Add llama.cpp ARM64 (Graviton) CPU-inference container

Jyothirmaikottu:
## Description

Introduces a new **llama.cpp** framework: an ARM64 (aarch64) **CPU-only** inference container for AWS Graviton, serving quantized GGUF models via `llama-server`'s OpenAI-compatible HTTP API. This is the first ARM64 image in the current pipeline and the first CPU-only LLM-serving image.

llama.cpp is compiled from a pinned upstream ref for aarch64 with `-mcpu=neoverse-v1` (Graviton3 baseline). A single image is portable across Graviton3/4/5 because llama.cpp repacks quantized weights to the CPU-optimal int8 layout at load time, so a Graviton3-baseline build still receives the newer-generation int8/bf16 kernels at runtime.

## What's included

- `docker/llama_cpp/Dockerfile.amzn2023` — multi-stage AL2023 aarch64 build (builder compiles `llama-server`; slim CPU-only runtime), with telemetry + OSS-compliance wired in like the other frameworks.
- `scripts/docker/llama_cpp/dockerd_entrypoint.sh` — EC2 `serve` entrypoint.
- `.github/config/image/llama_cpp/ec2-amzn2023.yml` — image config (`arch_type: arm64`, `device_type: cpu`, `job_type: inference`; release gated off for now — see note).
- `.github/config/model-tests/llama-cpp-model-tests.yml` — CPU smoke test on an ARM64 fleet with a small `Q4_0` GGUF.
- `.github/workflows/llama-cpp.{pipeline,tests-model,pr-amzn2023,autorelease-ec2-amzn2023}.yml` — build/test/release pipeline scoped to the CPU/EC2 path.
- `test/llama_cpp/scripts/llama_cpp_model_smoke_test.sh` — serves a GGUF and asserts `/health` + a chat completion.
- `test/security/data/ecr_scan_allowlist/llama_cpp/framework_allowlist.json`.

## Testing

The Dockerfile builds and serves a model end-to-end on a Graviton4 instance:
- `docker build` succeeds (image ~305 MB).
- Container serves Qwen2.5-0.5B `Q4_0`; `/health` ready in ~5s; `/v1/chat/completions` returns a correct response.

## Note for reviewers

The CI build job targets an ARM64 CodeBuild fleet (`fleet:arm64-llama-cpp-build-runner`). Until that fleet is provisioned the build job cannot run, so `release` is set to `false` in the image config. This PR is intended for review of the framework structure; enabling the build fleet is a follow-up.


.github/config/model-tests/llama-cpp-model-tests.yml
```diff
@@ -0,0 +1,19 @@
...
+# codebuild-fleet runners (ARM64):
+#   arm64-c8g4xl-runner = c8g.4xlarge (Graviton4, 16 vCPU / 32 GiB)
+
+s3_prefix: "s3://dlc-cicd-models/llm-models"
+
+smoke-test:
+  codebuild-fleet:
+    - name: "qwen2.5-0.5b-q4_0"
+      s3_model: "qwen2.5-0.5b-instruct-q4_0.gguf.tar.gz"
+      fleet: "arm64-c8g4xl-runner"
```
Jyothirmaikottu, line 18: tbd - both model and fleet

PR #6461, opened 2026-07-29
Patch mesa and emacs CVEs in PyTorch training 2.8/2.9/2.10

Eren-Jeager123:
## Description

Patches three CVEs in PyTorch training **2.9 / 2.10** and removes the now-stale `CVE-2026-40393` mesa allowlist entries.

None of the three is fixable by upgrading. In every case the flagged package is dead weight, so the fix is to replace or remove it.

| CVE | Package | Severity | Fix |
|---|---|---|---|
| CVE-2026-40393 | mesa | CRITICAL | `libgl1-mesa-glx` → `libgl1` |
| CVE-2025-1244 | emacs | HIGH | drop `emacs` from apt install list |
| CVE-2024-53920 | emacs | HIGH | same removal |

**2.8 is intentionally out of scope** — it reaches end of patch next week, so its Dockerfiles and allowlists are unchanged from `master`.

### CVE-2026-40393 (mesa)

The scan flags `libgl1-mesa-glx 23.0.4-0ubuntu1~22.04.1`, which is an **empty transitional package** — it ships zero `.so` files. The patched mesa source (`23.2.1-1ubuntu3.1~22.04.4`, USN-8427-1) no longer builds that binary, so there is no upgrade target and `apt-get upgrade` cannot clear the finding:

```
$ apt-get install -y --only-upgrade 'mesa*'
0 upgraded, 0 newly installed, 0 to remove

$ apt-cache madison libgl1-mesa-glx
libgl1-mesa-glx | 23.0.4-0ubuntu1~22.04.1 | jammy-updates/universe
libgl1-mesa-glx | 22.0.1-1ubuntu2         | jammy/universe
```

`libGL.so.1` is owned by `libgl1`, not `libgl1-mesa-glx`. The real mesa libraries (`libgl1-mesa-dri`, `libglapi-mesa`, `libglx-mesa0`) are already at the patched `23.2.1-1ubuntu3.1~22.04.4`. All three share the source name `mesa`, which is why the scanner reports the stale stub.

### CVE-2025-1244 / CVE-2024-53920 (emacs)

Upstream fixes are emacs 29.4 and 30.1, but Ubuntu jammy ships the `1:27.1` line and will not rebase. Jammy's backport is `1:27.1+1-3ubuntu5.2+esm1` — delivered **only via Ubuntu Pro / ESM Apps** — while public repos cap at the unfixed `1:27.1+1-3ubuntu5.2`. Neither fix path is reachable in a DLC build, and emacs is not needed for training.

This matches the existing removal in `huggingface/pytorch/training/docker/2.9/py3/cu130/Dockerfile.gpu`.

## Verification

Confirmed in the PR build of the 2.10 images (CPU + GPU):

- `test_ecr_enhanced_scan` **passed** on both — **no mesa CVE in the scan results at all**, confirming the swap eliminates the finding rather than suppressing it
- `test_oss_compliance`, `test_pip_check`, `test_license_file`, `test_boottime_container_security`, `test_safety_package_not_installed` all passed
- image size 4241 MB against a 7200 MB baseline

Also verified in a standalone build of the edited install block: `libgl1-mesa-glx` absent and **not re-pulled transitively**, `emacs` absent, `libGL.so.1` present, `opencv-python==4.11.0.86` imports.

No test references `emacs`, `mesa`, or `libgl1`.

## Versioning

`dlc_major_version` bumped `1` → `2` in the four 2.9 / 2.10 Dockerfiles (both the ec2 and sagemaker stages).

Removing emacs is a backward-incompatible change, so the patched images publish under a new major version rather than mutating v1. Customers pinned to v1 keep the emacs-bearing image. `DLContainersReleaseLogic` reads this label off the image manifest to build the `-v{major}.{minor}` release tag, so the label is the only change required.

The mesa swap on its own is *not* backward-incompatible: `libgl1-mesa-glx` is an empty transitional package that depends on `libgl1`, so `libgl1` was already installed. The swap drops an empty dpkg alias and leaves the same `libGL.so.1` in place.

## Test changes

**`test_dlc_labels.py`** — `test_dlc_major_version_dockerfiles` asserts the DLC major versions across matching Dockerfiles form a contiguous sequence starting at 1. PyTorch training 2.9 / 2.10 supersede v1 in place rather than adding a v2 variant beside it, so the expectation becomes the current major version only. Scoped to framework `pytorch` + job_type `training`; `huggingface_pytorch` (still v1) and all other frameworks keep the `1..N` rule.

**`test_pre_release.py`** — `test_package_version_regression_in_image` now skips `fastcore` and `python-fasthtml`:

- `fastai` calls `L.starmap` through its **latest** release (2.8.7); fastcore removed it in 2.0, so the training images hold `fastcore<2` (added in #6419 after a real `AttributeError: 'list' object has no attribute 'starmap'` crash in `vision_learner` during SageMaker training).
- `python-fasthtml 0.14.6` requires `fastcore>=2.0.4`, so under that pin pip necessarily resolves back to `0.14.5`. One root cause, two symptoms.
- The already-released images predate the pin and carry a fastcore 2.x that **crashes fastai**, so those versions are not a baseline worth enforcing.

No version satisfies both the pin and the no-downgrade assertion, and there is no fastai release that drops the `L.starmap` call. Real downgrades in every other package are still caught.

## Deliberately NOT changed

- **black** (`CVE-2026-31900`, `CVE-2026-32274`) — already pinned `>=26.3.1`; installed version resolves to 26.5.1. The finding's `file_path` is `spacy/tests/package/requirements.txt`, an inert test fixture that still reads `black==22.3.0`. Not fixable by bumping.
- **pyasn1** (`CVE-2026-30922`, `GHSA-jr27-m4p2-rc6r`) — fixable (0.6.4 exists), but 2.10 has no pyasn1 pin to edit; needs one added. Separate PR.
- **flash_attn** (`CVE-2026-31253`) — only `2.8.3.post1` available and the fix is unconfirmed; 2.9/2.10 build it from source.
- **torch**, **onnx** — no upstream fix / version is the framework itself.

## Note before merge

`dlc_developer_config.toml` is scoped for CI (`build_frameworks = ["pytorch"]`, `build_inference = false`). **This must be reverted to defaults before merge** or `test_dlc_developer_config.py` quick-checks will fail.

## Tests run

- [x] Local container verification (build + runtime checks above)
- [x] PR CI — 2.10 CPU + GPU security suite passed apart from the fastcore regression addressed here
- [ ] PR CI re-run after narrowing scope to 2.9 / 2.10

### DIFF
--- a/dlc_developer_config.toml
+++ b/dlc_developer_config.toml
@@ -37,12 +37,12 @@ deep_canary_mode = false
 [build]
 # Add in frameworks you would like to build. By default, builds are disabled unless you specify building an image.
 # available frameworks - ["base", "vllm", "sglang", "autogluon", "huggingface_vllm", "huggingface_vllm_omni", "huggingface_sglang", "huggingface_llamacpp", "huggingface_tensorflow", "huggingface_pytorch", "huggingface_tensorflow_trcomp", "huggingface_pytorch_trcomp", "pytorch_trcomp", "tensorflow", "pytorch", "stabilityai_pytorch"]
-build_frameworks = []
+build_frameworks = ["pytorch"]
 
 
 # By default we build both training and inference containers. Set true/false values to determine which to build.
 build_training = true
-build_inference = true
+build_inference = false
 
 # Set do_build to "false" to skip builds and test the latest image built by this PR
 # Note: at least one build is required to set do_build to "false"
@@ -124,7 +124,7 @@ nightly_pr_test_mode = false
 dlc-pr-base = ""
 
 # Standard Framework Training
-dlc-pr-pytorch-training = ""
+dlc-pr-pytorch-training = "pytorch/training/buildspec-2-8-ec2.yml"
 dlc-pr-tensorflow-2-training = ""
 dlc-pr-autogluon-training = ""
 
--- a/dlc_developer_config.toml
+++ b/dlc_developer_config.toml
@@ -124,7 +124,7 @@ nightly_pr_test_mode = false
 dlc-pr-base = ""
 
 # Standard Framework Training
-dlc-pr-pytorch-training = "pytorch/training/buildspec-2-8-ec2.yml"
+dlc-pr-pytorch-training = ""
 dlc-pr-tensorflow-2-training = ""
 dlc-pr-autogluon-training = ""
 
--- a/pytorch/training/docker/2.10/py3/Dockerfile.cpu
+++ b/pytorch/training/docker/2.10/py3/Dockerfile.cpu
@@ -74,12 +74,11 @@ RUN apt-get update \
     ca-certificates \
     cmake \
     curl \
-    emacs \
     git \
     jq \
     libcurl4-openssl-dev \
     libglib2.0-0 \
-    libgl1-mesa-glx \
+    libgl1 \
     libsm6 \
     libssl-dev \
     libxext6 \
--- a/pytorch/training/docker/2.10/py3/Dockerfile.cpu
+++ b/pytorch/training/docker/2.10/py3/Dockerfile.cpu
@@ -206,8 +206,6 @@ RUN pip install --no-cache-dir -U torch==${PYTORCH_VERSION} \
     torchdata==${TORCHDATA_VERSION} \
     s3torchconnector \
     fastai \
-    # pin fastcore<2: fastcore 2.0 removed L.starmap, which pinned fastai's optimizer.py still calls
-    "fastcore<2" \
     accelerate \
     # pin numpy requirement for fastai dependency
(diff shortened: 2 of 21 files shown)


Eren-Jeager123:
Superseded by per-version PRs so each gets its own CI run (the PR build handles one framework version at a time):

- #6465 — PyTorch training 2.9
- #6466 — PyTorch training 2.10

2.8 is dropped from scope (end of patch next week). The Dockerfile and allowlist changes in both new PRs are byte-identical to the ones verified here — including the run where `test_ecr_enhanced_scan` passed with no mesa CVE in the results.

Closing in favor of those two.

PR #6467, opened 2026-07-30
vllm: bump AL2023 images to upstream commit d223c90

Jyothirmaikottu:
## Summary
Bumps the AL2023 vLLM images (EC2 + SageMaker) to build from upstream vLLM commit
[`d223c90`](https://github.com/vllm-project/vllm/commit/d223c900d85224c02f2162ee2c757a769e99f519)
(vllm-project/vllm#50060, "[Bugfix] Only pad transformers backend `value` when it is narrower").

### Config (both `ec2-amzn2023.yml` and `sagemaker-amzn2023.yml`)
```
vllm_ref:           7b3d595… -> d223c900d85224c02f2162ee2c757a769e99f519
flashinfer_version: 0.6.12   -> 0.6.15.post1
deepep_commit_hash: 73b6ea4  -> d4f41e4e93602a15e95f55f6ee8df8f1aaa0e4bb
dlc_minor_version:  1        -> 2
```
FlashInfer and DeepEP track what vLLM pins/expects at `d223c90`.

### Dockerfile (`docker/vllm/Dockerfile.amzn2023`) — NCCL override
DeepEP `d4f41e4e93` (required because `d223c90` patches DeepEP's `symmetric.hpp`) contains the EPv2/GIN
backend. Its `nccl.cu` uses `ncclDevCommRequirements::ginTrafficClass` / `ncclGinRequest_t`, which require
**NCCL >= 2.30.4**. `torch==2.13.0` transitively pins `nvidia-nccl-cu13==2.29.7`, so an initial build failed:

```
nccl.cu(99): error: class "ncclDevCommRequirements" has no member "ginTrafficClass"
```

Mirrors upstream vLLM's `docker/Dockerfile` fix: adds `ARG NCCL_VERSION=2.30.7` and sets `UV_OVERRIDE`
(`nvidia-nccl-cu13==2.30.7`) in both the build stage (before `install_python_libraries.sh`, which
reinstalls torch and thus resolves the overridden NCCL) and the deps stage (so the runtime venv matches the
NCCL the DeepEP wheel was compiled against).

A second, distinct failure surfaced once the version bump was in place — the DeepEP link step failed with:

```
/usr/bin/ld: cannot find -l:libnccl.so.2: No such file or directory
```

On CUDA 13 NCCL is a separate wheel and is no longer bundled in `torch/lib`, so DeepEP's setup passed
`nvidia/nccl/lib` only as `-Wl,-rpath` (runtime), never as `-L` (link-time). The build stage now exports
`LIBRARY_PATH` pointing at that dir so the linker can resolve `libnccl.so.2`. Verified end-to-end on a devbox:
`install_python_libraries.sh --mode wheel` builds the DeepEP wheel with exit 0 and no link error.

The earlier `CUDA_MAJOR >= 13` guards around the override were dropped — this Dockerfile's only CUDA 12
consumer (hyperpod) is filtered out of PR builds and not released, so the guard was dead scaffolding.

## Note
`d223c90` is ~1294 commits ahead of the previously pinned ref but is not a strict fast-forward — it does not
contain 7 commits that were on the prior ref (Minimax-M3, GLM5 Router GEMM, OpenMOSS support, DSv3.2 topK
perf, and two FlashInfer MoE kernels). Confirmed acceptable for the target day-0 model.

### DIFF
--- a/.github/config/image/vllm/ec2-amzn2023.yml
+++ b/.github/config/image/vllm/ec2-amzn2023.yml
@@ -19,7 +19,7 @@ build:
   cuda_version: "13.0.2"
   vllm_ref: "d223c900d85224c02f2162ee2c757a769e99f519"
   flashinfer_version: "0.6.15.post1"
-  deepep_commit_hash: "73b6ea4"
+  deepep_commit_hash: "d4f41e4e93602a15e95f55f6ee8df8f1aaa0e4bb"
   dlc_major_version: "2"
   dlc_minor_version: "2"
   use_sccache: "true"
--- a/.github/config/image/vllm/ec2-amzn2023.yml
+++ b/.github/config/image/vllm/ec2-amzn2023.yml
@@ -18,7 +18,7 @@ build:
   python_version: "3.12"
   cuda_version: "13.0.2"
   vllm_ref: "d223c900d85224c02f2162ee2c757a769e99f519"
-  flashinfer_version: "0.6.12"
+  flashinfer_version: "0.6.15.post1"
   deepep_commit_hash: "73b6ea4"
   dlc_major_version: "2"
   dlc_minor_version: "2"
--- a/.github/config/image/vllm/ec2-amzn2023.yml
+++ b/.github/config/image/vllm/ec2-amzn2023.yml
@@ -4,7 +4,7 @@ image:
 
 metadata:
   framework: "vllm_server"
-  framework_version: "0.24.0"
+  framework_version: "0.26.0"
   os_version: "amzn2023"
   customer_type: "ec2"
   arch_type: "x86"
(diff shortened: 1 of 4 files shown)


evakravi (dismissed):

Jyothirmaikottu:
NCCL 2.30.7 `UV_OVERRIDE` mirrors upstream vLLM [`docker/Dockerfile`](https://github.com/vllm-project/vllm/blob/d223c900d85224c02f2162ee2c757a769e99f519/docker/Dockerfile#L31) (`ARG NCCL_VERSION=2.30.7`), required for the DeepEPv2 GIN backend.

junpuf (approved):

PR #6501, opened 2026-08-06
add: llama.cpp inference containers — x86 CPU/GPU + ARM64 CPU (AL2023)

Jyothirmaikottu:
## Purpose

Add **llama.cpp** inference DLCs that serve quantized GGUF models via `llama-server` (OpenAI-compatible HTTP API), built from a pinned upstream ref on AL2023. This PR ships six images across three variants, each for EC2 and SageMaker:

| Variant | Arch | Device | Base image | ECR repo | Prod tag (EC2 / SageMaker) |
|---|---|---|---|---|---|
| CPU | x86_64 | cpu | `amazonlinux:2023` | `llama-cpp` | `server-cpu-v1` / `server-sagemaker-cpu-v1` |
| GPU (CUDA) | x86_64 | gpu | `nvidia/cuda:13.0.2-*-amzn2023` | `llama-cpp` | `server-cuda-v1` / `server-sagemaker-cuda-v1` |
| CPU (Graviton) | arm64 | cpu | `amazonlinux:2023` | `llama-cpp-arm64` | `server-cpu-v1` / `server-sagemaker-cpu-v1` |

Design notes:
- **x86 CPU** uses `GGML_CPU_ALL_VARIANTS` so one image carries every x86 microarch backend (SSE4.2/AVX/AVX2/AVX512/AMX) and dispatches at runtime — portable across modern EC2 x86.
- **x86 GPU** builds with `GGML_CUDA=ON` plus `GGML_CPU_ALL_VARIANTS`, so one image carries the CUDA backend and CPU fallbacks, dispatched at runtime. No GPU is needed to build (the CUDA toolkit comes from the devel base), so it builds on the existing x86 fleet.
- **ARM64** compiles with `-mcpu=neoverse-v1` (Graviton3 baseline) passed via `CMAKE_C_FLAGS`/`CMAKE_CXX_FLAGS`; a single image is portable across Graviton3/4/5 because llama.cpp repacks quantized weights to the CPU-optimal int8 layout at load time.

SageMaker variants front `llama-server` with an nginx reverse proxy mapping `GET /ping → /health` and `POST /invocations → /v1/chat/completions`.

Files follow the existing V2 layout (sglang as reference): per-image configs, a shared pipeline, one glob-based PR caller, a manual `dispatch-release` caller, model-test workflows, entrypoints, smoke test, and an ECR scan allowlist. Build fleets are arch-keyed in the pipeline: x86 (CPU + CUDA) on the dedicated `x86-llama-cpp-build-runner` fleet, arm64 on the shared `arm64-graviton4-build-runner` (Graviton4) fleet. CI model tests run the same Ornith-1.0-9B (Q4_K_M) GGUF — CPU-only on `x86-build-runner`, GPU-offloaded on the single-GPU `x86-g6xl-runner` (1× L4) — gated by `required_image_pattern` per arch/device. arm64 CPU model tests are deferred until an arm64 model-test fleet is onboarded.

### Release

All image configs ship with `release: false`. A manual **`llama-cpp.dispatch-release.yml`** (`workflow_dispatch`) is included so the first release of each image can be triggered by hand once it's validated (flip that image's config `release: true`, then run the workflow). Scheduled autorelease is intentionally **not** wired yet — it will be onboarded in a follow-up once manual releases are proven.

### Shared-file changes

A few shared, cross-framework files are touched to support the arm64/Graviton path and unblock the scan on the touched telemetry script:
- `.github/buildspec-cb-fleet.yml`: install arch-appropriate `yq`/`jq` (x86 unchanged; arm64 was pulling x86 binaries → `Exec format error`).
- `.github/workflows/_reusable.sanity-tests.yml`, `_reusable.telemetry-tests.yml`: route the image-exec test jobs to an arm64 fleet when the image is arm64 (x86 is a byte-for-byte no-op).
- `docker/sklearn/1.4-2-py312/requirements.txt`: bump `cryptography` 46.0.7 → 50.0.0 (HIGH CVE-2026-69247/69249). Required because editing the shared telemetry script re-triggers the sklearn image scan; matches the pin sklearn 1.9-0 already uses.
- `scripts/docker/telemetry/deep_learning_container.py`: add `llama_cpp` to the framework allowlist.

## Test Plan

- `docker build` for the x86 GPU target on a CPU-only host (validates the CUDA build compiles and links without a GPU present).
- `docker build` + local `llama-server` serve for the x86 CPU and arm64 CPU targets with a small GGUF.
- CI: sanity, security (ECR scan), telemetry, and model-test (CPU + GPU) across the discovered configs, per arch.

## Test Result

- Local x86 GPU build: PASS (all binaries linked). Local x86/arm64 CPU serve: `/health` → 200; `POST /v1/chat/completions` → correct OpenAI-shaped response.
- CI: x86 CPU/GPU builds, CPU + GPU model tests, and SageMaker endpoint tests passing; arm64 builds passing after the `-mcpu` fix; arm64 sanity/telemetry routed to the Graviton fleet. See the latest run on this PR for current status.


docker/llama_cpp/Dockerfile.gpu
```diff
@@ -0,0 +1,140 @@
...
+
+# Expose the CUDA compute + utility driver libraries to the container at runtime.
+ENV NVIDIA_DRIVER_CAPABILITIES=compute,utility
+
+RUN dnf install -y --allowerasing --setopt=install_weak_deps=False \
+  libgomp libcurl curl-minimal openblas shadow-utils \
+  python3 python3-pip \
+  # Upgrade cuda-compat (CUDA base ships an older one) to clear
+  # CVE-2026-24193 / CVE-2025-33219, which --security does not flag.
+  && dnf upgrade -y cuda-compat-13-0 \
```
sirutBuasai: If we're able to get `13-0` string from our config file with cuda `13.0.2` then that'd be better than hardcoding the dockerfile. We can use the pre_build.sh hooks for this parsing.

.github/config/image/llama_cpp/sagemaker-amzn2023.yml
```diff
@@ -0,0 +1,30 @@
...
+  platform: "sagemaker"
+  arch_type: "x86"
+  device_type: "cpu"
+  job_type: "general"
+  prod_image: "llama-cpp:server-sagemaker-cpu-v1"
+
+build:
+  dockerfile: "docker/llama_cpp/Dockerfile.cpu"
+  target: "llama-cpp-sagemaker"
+  python_version: "3.12"
```
sirutBuasai, line 19: This is not used anywhere in the dockerfile

sirutBuasai (approved):

PR #6508, opened 2026-08-07
ray-serve-llm: add ray-llm image variant for Ray Serve + vLLM inference

jinyan-li1:
## Summary

Adds a new `ray-llm` DLC variant that ships Ray Serve + vLLM on the `cu130-amzn2023` base image, for online LLM inference on EC2 and multi-node EKS via KubeRay.

**Runtime shape:**
- Pass-through entrypoint (same pattern as `ray-train`) so KubeRay's `rayStartParams` drives `ray start` unmodified
- Multi-node inference over EFA using vLLM's Ray executor + tensor parallelism
- Customer configures via `serveConfigV2` in the RayService (see `test/ray-llm/eks-multi-node/rayservice.yml` for a working example)

## Key components + versions
| Component | Version |
|---|---|
| ray[default,serve,llm] | 2.57.0 |
| vllm | 0.25.1 |
| torch (cu130 wheel) | 2.11.0 |
| transformers | 5.14.1 |
| Base image | `base:runtime-cu130-amzn2023` |
| EFA + aws-ofi-nccl | via `install_efa_amzn2023.sh` |
| libnccl (from NVIDIA cuda-rhel9 repo) | 2.x |

All third-party licenses generated at build time via `setup_oss_compliance.sh` (`THIRD_PARTY_SOURCE_CODE_URLS`, `PYTHON_PACKAGES_LICENSES`, `LINUX_PACKAGES_LICENSES`).

## Files added
- `docker/ray-llm/Dockerfile.cuda`, `pyproject.toml`, `uv.lock` — image definition + locked deps
- `.github/config/image/ray-llm/ec2-gpu.yml` — image config; framework `ray-llm`, publishes to shared `ray` ECR repo with `container_type: llm`
- `scripts/docker/ray-llm/entrypoint.sh` — pass-through with CUDA forward-compat check
- `test/ray-llm/unit/` — package-import + shell-tool sanity
- `test/ray-llm/ec2/` — single-GPU functional (docker run + `/v1/completions`/`/v1/chat/completions` against Ministral-8B)
- `test/ray-llm/eks-multi-node/` — KubeRay RayService with TP=2 across 2 g6.12xlarge nodes, sharding + EFA-linked NCCL verification
- `.github/workflows/ray-llm.*.yml` — pipeline, PR trigger, unit + ec2 + eks-multi-node test reusables
- `test/security/data/ecr_scan_allowlist/ray-llm/framework_allowlist.json` — CVE allowlist (bundled jars only)
- Small change to `scripts/docker/telemetry/deep_learning_container.py` — adds `"ray-llm"` to the framework allowlist so telemetry accepts our `metadata.framework: ray-llm`

## Test plan
- [x] CI pipeline green
- [x] EC2 test, EKS multi-node validation, EFA test all passed: https://github.com/aws/deep-learning-containers/actions/runs/32439346515/job/96646869221?pr=6508


.github/config/image/ray-llm/ec2-gpu.yml
```diff
@@ -0,0 +1,34 @@
...
+  python_short_version: "3.13"
+  cuda_version: "13.0.2"
+  vllm_version: "0.25.1"
+  dlc_major_version: "1"
+  dlc_minor_version: "0"
+
+release:
+  release: true
+  force_release: false
+  public_registry: false
```
sallyseok: intended?

scripts/docker/ray_llm/entrypoint.sh
sallyseok, line 1: Can we reuse `scripts/docker/common/start_cuda_compat.sh`?

scripts/docker/telemetry/deep_learning_container.py
```diff
@@ -241,6 +241,7 @@ def parse_args():
             "sglang_server",
             "lambda",
             "ray",
+            "ray-llm",
```
sallyseok: I see we don't have an entry for ray-train here; is ray-llm needed or can we group it in ray?

.github/config/image/ray-llm/ec2-gpu.yml
```diff
@@ -0,0 +1,34 @@
...
+
+metadata:
+  framework: "ray-llm"
+  framework_version: "2.57.0"
+  os_version: "amzn2023"
+  customer_type: "ec2"
+  arch_type: "x86"
+  device_type: "gpu"
+  job_type: "inference"
+  container_type: "llm"
```
sallyseok: What's the difference between `job_type` and `container_type`? Not necessarily related to this PR, but this already seems to be a vague distinction in this repo, and for ray llm it might become a real issue since we have ray, ray train, and ray llm now. Let's make the distinction clear and make sure it won't cause issues in the releas logic

.github/config/image/ray-llm/ec2-gpu.yml
sallyseok: intended?

.github/config/image/ray-llm/ec2-gpu.yml
sallyseok: Also this is kind of a nitpick but the framework name for vllm omni is using an underscore (`vllm_omni`) while this uses hyphen (line 6 above), while ray train just uses `ray` as the framework name. We need to clearly define the framework name for all three, container type vs job type, and make sure all of this matches the release logic

.github/config/image/ray-llm/ec2-gpu.yml
jinyan-li1: changed to false and gamma until i set up the release logic, thanks!

scripts/docker/telemetry/deep_learning_container.py
jinyan-li1: i think we would want more granular telemetry data, and ray train should do the same

scripts/docker/ray_llm/entrypoint.sh
jinyan-li1: yes, good catch, updated!

.github/config/image/ray-llm/ec2-gpu.yml
jinyan-li1: `job_type` = training/inference/general (docs, gates). `container_type` = release tag suffix. Ray family: ray → `serve-ml-cuda`, ray-train → `train-ml-cuda`, ray-llm → `serve-llm-cuda` (updated from `llm` this round).

Framework name kept as `ray_llm` (matching `vllm_omni`). I'll handle the release-logic entry so ray-llm ships in the same `ray` ECR repo; telemetry allowlist stays separate. Ray-train should follow the same pattern.

docker/ray_llm/Dockerfile.cuda
```diff
@@ -0,0 +1,154 @@
...
+
+# Shell tools vLLM needs at runtime:
+# - `which` for FlashInfer's nvcc probe (without it: FileNotFoundError)
+# - gcc/g++/make + CC for torch.compile's runtime kernel compilation
+# - wget for KubeRay's default health probes
+# - cuda-nvcc / cudart-devel for any user-side vLLM plugin builds
+# - libcurand-devel: FlashInfer JIT-compiles its sampling kernel on first use
+#   (top-k/top-p), which #includes curand.h. Missing header → RuntimeError from
+#   ninja, EngineCore fails to start, Ray Serve loops on replica init.
+RUN CUDA_MAJOR_MINOR=$(echo "13.0" | tr '.' '-') \
```
Yadan-Wei: 13.0 is hardcoded here, could you make it depends on cuda version, it looks like your config has cuda_version defined but not used anywhere in the Dockerfile.

jinyan-li1: thanks! did that in a new rev

.github/config/image/ray-llm/ec2-gpu.yml
sallyseok: Is the container_type actually consumed by the dockerfile or anything else? I know job_type is used by the telemetry

.github/config/image/ray-llm/ec2-gpu.yml
sallyseok: sorry should've been more clear, I was more asking about the `public_registry` part -- do we intend to release to public gallery?

jinyan-li1: oh yeah, we do want to release to public gallery - i made the change in https://github.com/aws/deep-learning-containers/pull/6508/changes/050f4e1a74c465c3c786992619869a95e763fcc6

.github/config/image/ray-llm/ec2-gpu.yml
jinyan-li1: I think container_type is consumed here and gets written into the release spec: https://github.com/aws/deep-learning-containers/blob/main/.github/actions/generate-release-spec/generate_release_spec.sh

sallyseok (approved):

PR #6514, opened 2026-08-10
sglang Kimi-k3 verification

ezhong0211:
## Purpose

Bump the AL2023 SGLang DLC from 0.5.14 to 0.5.17.

The AL2023 image is pinned to SGLang 0.5.14, which does not contain the
Kimi-K3 model implementation (`sglang/srt/models/kimi_k3.py`) or the
`kimi_k3` reasoning/tool-call parsers. As a result, the default EC2 SGLang
DLC cannot serve Kimi-K3 at all. The Ubuntu variant is already at 0.5.17
(auto-updated via `[Auto-Update] sglang` PRs) and serves K3 correctly. AL2023
lags because it builds SGLang from source rather than pulling an upstream
base image, so it is not covered by the currency automation.

This is a currency bump that delivers customer capability on B300-class
hardware. It does **not** add Kimi-K3 to the CI model-test matrix.

Pins updated, all sourced from upstream `v0.5.17`:

| Field | Old | New | Source |
|---|---|---|---|
| `framework_version` | `0.5.14+dlc1` | `0.5.17+dlc1` | matches Ubuntu config |
| `sglang_ref` | `bc8b3ab1...` | `b6a09f38fcc5e96574324b4acc19d421c539cfc6` | `refs/tags/v0.5.17` |
| `sglang_kernel_version` | `0.4.4` | `0.4.5` | upstream `pyproject.toml` + Dockerfile `SGL_KERNEL_VERSION` |
| `flashinfer_version` | `0.6.12` | `0.6.15.post1` | upstream `pyproject.toml` (`flashinfer_python[cu13]`) |
| `mooncake_version` | `0.3.11.post1` | `0.3.12.post1` | upstream Dockerfile `MOONCAKE_VERSION` |
| `dlc_minor_version` | `2` | `3` | rebuild |

Deliberately unchanged: `deepep_commit` (upstream v0.5.17 pins the same
`9af0e0d0...`), `torch_version` (`2.11.0`, matches upstream), and
`torch_cuda_arch_list` (`9.0;10.0;10.3`, already includes sm103/B300).

Note: the previous `sglang_ref` (`bc8b3ab1`) was a post-tag commit, not the
`v0.5.14` tag (`4289f36e`). This change moves to the clean `v0.5.17` tag SHA.

## Test Plan

**CI (automated):**
- Source build on `x86-sglang-build-runner` — primary risk is the
  `sglang-kernel` 0.4.4 -> 0.4.5 and FlashInfer 0.6.12 -> 0.6.15.post1
  source compilation.
- Sanity, telemetry, and security tests.
- `sglang.tests-upstream.yml` — upstream suite against 0.5.17.
- `sglang.tests-model.yml` — all existing entries in
  `sglang-model-tests.yml`, confirming no regression on currently
  shipped models.

**Manual (Kimi-K3, cannot run in CI):**
Kimi-K3 requires ~195 GB/GPU across 8 GPUs; the largest CI runner
(`gpu-h100-8gpu-runners`) provides 80 GB/GPU. K3 was therefore verified
manually against the CI-built AL2023 image on a `p6-b300.48xlarge`
(8x B300, 275 GB each):
- In-image check: `sglang.srt.models.kimi_k3` importable, `kimi_k3` present
  in `ReasoningParser.DetectorMap`.
- Serve: `--tp-size 8 --dcp-size 8 --trust-remote-code
  --mem-fraction-static 0.85 --reasoning-parser kimi_k3
  --tool-call-parser kimi_k3`
- `/health`, `/v1/completions`, `/v1/chat/completions`, `/get_model_info`.

## Test Result

*(fill in after the CI-built image is verified on the B300 droplet)*

- [ ] CI build green
- [ ] Upstream tests green
- [ ] Existing model tests green (no regression)
- [ ] Manual K3 verification on `p6-b300.48xlarge` against CI-built AL2023 image

**Out of scope:** no entry is added to `sglang-model-tests.yml`. Kimi-K3's
1.56 TB of weights (2.4x the largest runner's total VRAM, before KV cache)
cannot be scheduled on any available fleet, so a config entry would never run.

<details>

## PR Checklist

- [ ] I ran `pre-commit run --all-files` locally before creating this PR. (Read [DEVELOPMENT.md](https://github.com/aws/deep-learning-containers/blob/main/DEVELOPMENT.md) for details).

</details>

### DIFF
--- a/.github/config/image/sglang/ec2-amzn2023.yml
+++ b/.github/config/image/sglang/ec2-amzn2023.yml
@@ -30,7 +30,7 @@ build:
   gdrcopy_version: "2.6"
   torch_cuda_arch_list: "9.0;10.0;10.3"
   dlc_major_version: "1"
-  dlc_minor_version: "4"
+  dlc_minor_version: "3"
 
 release:
   release: true
--- a/.github/config/image/sglang/ec2-amzn2023.yml
+++ b/.github/config/image/sglang/ec2-amzn2023.yml
@@ -17,7 +17,7 @@ build:
   target: "sglang-ec2"
   python_version: "3.12"
   cuda_version: "13.0.3"
-  sglang_ref: "b6a09f38fcc5e96574324b4acc19d421c539cfc6"
+  sglang_ref: "29481685462732237d80d86076d6563e1f658102"
   sglang_kernel_version: "0.4.5"
   flashinfer_version: "0.6.15.post1"
   deepep_commit: "9af0e0d0e74f3577af1979c9b9e1ac2cad0104ee"
--- a/.github/config/image/sglang/ec2-amzn2023.yml
+++ b/.github/config/image/sglang/ec2-amzn2023.yml
@@ -30,7 +30,7 @@ build:
   gdrcopy_version: "2.6"
   torch_cuda_arch_list: "9.0;10.0;10.3"
   dlc_major_version: "1"
-  dlc_minor_version: "3"
+  dlc_minor_version: "4"
 
 release:
   release: true
--- a/.github/config/image/sglang/ec2-amzn2023.yml
+++ b/.github/config/image/sglang/ec2-amzn2023.yml
@@ -4,7 +4,7 @@ image:
 
 metadata:
   framework: "sglang_server"
-  framework_version: "0.5.14+dlc1"
+  framework_version: "0.5.17+dlc1"
   os_version: "amzn2023"
   customer_type: "ec2"
   arch_type: "x86"
@@ -17,11 +17,11 @@ build:
   target: "sglang-ec2"
   python_version: "3.12"
   cuda_version: "13.0.3"
-  sglang_ref: "bc8b3ab1f55c60951b586d84d4aec773a6f654df"
-  sglang_kernel_version: "0.4.4"
-  flashinfer_version: "0.6.12"
+  sglang_ref: "b6a09f38fcc5e96574324b4acc19d421c539cfc6"
+  sglang_kernel_version: "0.4.5"
+  flashinfer_version: "0.6.15.post1"
   deepep_commit: "9af0e0d0e74f3577af1979c9b9e1ac2cad0104ee"
-  mooncake_version: "0.3.11.post1"
+  mooncake_version: "0.3.12.post1"
   torch_version: "2.11.0"
   nccl_version: "2.30.4"
   cudnn_version: "9.16.0.29"
@@ -30,7 +30,7 @@ build:
   gdrcopy_version: "2.6"
   torch_cuda_arch_list: "9.0;10.0;10.3"
   dlc_major_version: "1"
-  dlc_minor_version: "2"
+  dlc_minor_version: "3"
 
 release:
   release: true
--- a/.github/config/image/sglang/sagemaker-amzn2023.yml
+++ b/.github/config/image/sglang/sagemaker-amzn2023.yml
@@ -31,7 +31,7 @@ build:
   gdrcopy_version: "2.6"
   torch_cuda_arch_list: "9.0;10.0;10.3"
   dlc_major_version: "1"
-  dlc_minor_version: "3"
+  dlc_minor_version: "4"
 
 release:
   release: true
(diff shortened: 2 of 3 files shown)


.github/config/image/sglang/ec2-amzn2023.yml
```diff
@@ -30,7 +30,7 @@ build:
   gdrcopy_version: "2.6"
   torch_cuda_arch_list: "9.0;10.0;10.3"
   dlc_major_version: "1"
-  dlc_minor_version: "2"
+  dlc_minor_version: "4"
```
Yadan-Wei: May I know why here we bump two minor version?

.github/config/image/sglang/sagemaker-amzn2023.yml
```diff
@@ -31,7 +31,7 @@ build:
   gdrcopy_version: "2.6"
   torch_cuda_arch_list: "9.0;10.0;10.3"
   dlc_major_version: "1"
-  dlc_minor_version: "2"
+  dlc_minor_version: "4"
```
Yadan-Wei: Same here.

docker/sglang/Dockerfile.amzn2023
```diff
@@ -342,21 +351,92 @@ RUN PYVER=$(python${PYTHON_VERSION} -c "import sys; print(f'{sys.version_info.ma
...
+RUN FLASHINFER_DCP_PATCH=/sgl-workspace/sglang/docker/kimi_k3/flashinfer-perkz-dcp-0.6.15.txt \
+  && FLASHINFER_SITE_PACKAGES="$(python${PYTHON_VERSION} -c 'from pathlib import Path; import flashinfer; print(Path(flashinfer.__file__).resolve().parent.parent)')" \
+  && sed '/^diff --git a\/tests\//,$d' "${FLASHINFER_DCP_PATCH}" \
+     | patch --dry-run --batch --forward --strip=1 --directory="${FLASHINFER_SITE_PACKAGES}" \
+  && sed '/^diff --git a\/tests\//,$d' "${FLASHINFER_DCP_PATCH}" \
+     | patch --batch --forward --strip=1 --directory="${FLASHINFER_SITE_PACKAGES}" \
+  && python${PYTHON_VERSION} -c "import inspect, flashinfer.decode as d; \
+     assert 'enable_dcp' in inspect.signature(d.trtllm_batch_decode_with_kv_cache_mla).parameters, \
+     'flashinfer DCP patch did not apply'; print('flashinfer DCP ok')" \
+  && rm -rf /root/.cache/flashinfer /root/.cache/pip
```
Yadan-Wei, line 439: May I know these patches are temp and will be included in future version release or they need to stay all times.

.github/config/image/sglang/ec2-amzn2023.yml
ezhong0211: Fixed

.github/config/image/sglang/sagemaker-amzn2023.yml
ezhong0211: Fixed

docker/sglang/Dockerfile.amzn2023
ezhong0211: Only one of these is temporary. The FlashInfer DCP patch should go away. It ships inside the sglang tree at docker/kimi_k3/flashinfer-perkz-dcp-0.6.15.txt and upstream applies it only in their own K3-specific image rather than mainline, so once enable_dcp lands in a released flashinfer past 0.6.15.post1 this block should be deleted outright. The --dry-run guard ahead of the real patch is there so a stale patch fails the build loudly instead of half-applying, which is the intended removal signal, and I'm happy to add an explicit TODO marker so it doesn't depend on someone reading the comment. Everything else in that section is meant to stay. The trtllm-gen cubin pool is not a patch but a versioned, sha256-pinned data artifact that K3 selects its MoE kernels from, so it just gets bumped when upstream publishes a new pool tag. The cuda-nvrtc-devel package was simply missing from the runtime dnf list alongside cuda-nvcc and cuda-cudart-devel, and the cutlass-dsl symlink merge fixes a purelib/platlib split unique to AL2023 that breaks "import cutlass.cute" for any model, not just K3, since Debian and Ubuntu collapse both trees into one site-packages dir and upstream never sees it. The three dependency pins correct specs that were floating past their own declared ceilings, setuptools past torch's <82 and numpy past numba's <2.5, so those improve the baseline for every model in the image.

Yadan-Wei (approved):

PR #6536, opened 2026-08-11
Metadata store/image test skip logic test branch

samsullivan1:
## Purpose

  ⚠️  **Validation-only PR — do not merge.** This exists solely to exercise the DLC CI
  test-skip logic end to end against real infrastructure (ECR image reads + the
  `dlc-ci-images` DynamoDB table). Fork/local tests can't cover the cross-account
  registry read and DynamoDB read/write, so this runs it for real in CI.

  The logic under test:
  - `check-test-pass` action → computes `image_content_hash` + `suite_code_hash`,
    queries the table, outputs `skip=true|false`.
  - `record-test-pass` action → writes a PASS row after the suite passes.
  - Wired into `pytorch.tests-unit.yml` for the `pytorch/unit` suite.

  The `test/pytorch/unit/**` edit in this branch is a throwaway trigger commit; it will
  be dropped before the logic is contributed to `main`.

  ## Test Plan

  Confirm the skip cache works against real infra via two CI runs on this PR:

  1. **Run 1 (cache MISS):** table has no matching row → `skip=false` → unit suite runs →
     on pass, `record-test-pass` writes the PASS row. Validates the registry read, both
     hashes, and the DynamoDB **write** path.
  2. **Run 2 (cache HIT):** re-run the same job with no change under `test/pytorch/unit/**`
     (identical hashes) → expect `skip=true` and the docker/pytest steps skipped.
     Validates the DynamoDB **read**/skip path.

  Trigger fans out to the pytorch PR variants, unit suite only (CPU default-runners, no
  build, no GPU/sanity/security/telemetry) — watch **PR - PyTorch 2.11 CPU**.

  ## Test Result

  - [ ] Run 1 (MISS): `skip=false`, suite runs, PASS row recorded — <run link>
  - [ ] Run 2 (HIT): `skip=true`, suite skipped, "cached PASS" summary shown — <run link>


samsullivan1:
Throwaway test-skip benchmark PR — data collected (unit cold+warm, sanity cold+record). Closing.

PR #6568, opened 2026-08-17
feat(metadata store): image test skip logic

samsullivan1:
## Purpose

DLC CI rebuilds and retests every image variant on every push — even when the resulting image is byte-identical to one already tested — re-running suites that cannot have changed and burning significant compute (including GPU runners).

This PR lands the groundwork for a **content-addressed test-skip cache** backed by the `dlc-ci-images` DynamoDB table: once a suite has passed for a given *image content* + *test code*, a later run on the same content skips it instead of provisioning a runner.

  How it works:
  - **`image_content_hash`** — sha256 over the image's ordered rootfs layer
    `diff_ids` and JSON config fields, excluding any non-deterministic fields (e.g. created time stamp)(`scripts/ci/test_skip/hash_image_content.py`).
  - **`suite_code_hash`** — sha256 over the file set a suite is defined by in
    `.github/config/test-suites.yml` (`scripts/ci/test_skip/hash_suite_code.py`).
  - A suite is skippable **iff** a prior PASS row exists for **both** hashes —
    `PK = image_content_hash`, `SK = TEST#<suite>#<suite_code_hash>`
    (`scripts/ci/test_skip/ci_images_db.py`).
  - The **`check-test-pass`** composite action does one batched read up front and
    reports which suites are cache hits, so skippable suites never provision a test
    runner; **`record-test-pass`** writes a PASS row after a suite fully passes.
  - **Fail-open by design:** any error (hash failure, unknown suite, DynamoDB error,
    batch/throttling) falls back to running the suite — a false cache hit (skipping a
    test that should run) is never possible.

**Scope:** this PR adds the cache mechanism and its building blocks — the two hashers, the DynamoDB client, the two composite actions, the reusable check workflow, the suite config, and unit tests. Wiring the actions into the framework pipelines is done incrementally in follow-up changes.

## Test Plan

- Unit tests (no AWS creds required; DynamoDB mocked with `moto`): `python -m pytest test/db --confcutdir test/db -v` covering the hashers, the store read/record paths, the check orchestrator's skip decision + fail-open behavior, and a consistency check that every suite a workflow references exists in `test-suites.yml`.
- End-to-end against a real `dlc-ci-images` table in CI: record a PASS for a suite, then confirm a subsequent run on unchanged image content + test code is reported as a cache hit and the suite is skipped (no runner provisioned); changing the image content produces a fresh hash so the suite runs again.

## Test Result 

- `pytest test/db/` — all pass (36 tests).
- End-to-end in CI: cache hit confirmed against the live table; the skippable suite provisioned zero test runners; the PASS row was recorded and reused on re-run; an image-content change correctly forced a re-run. Tested on PyTorch Unit tests, and Sanity and Telemetry tests on SGLang AMZN 2023 and Ubuntu images.

<details>

## PR Checklist

</details>


.github/actions/check-test-pass/action.yml
```diff
@@ -0,0 +1,103 @@
...
+        RESULT=$(python3 "${{ github.action_path }}/check_test_pass.py" \
+          --image-uri "$RESOLVED_IMAGE_URI" \
+          --suites "$SUITES" \
+          --repo-root "$(git rev-parse --show-toplevel)")
+        [ -n "$RESULT" ] || RESULT='{"skips":{},"image_content_hash":"","suite_code_hashes":{}}'
+        {
+          echo "skips=$(printf '%s' "$RESULT" | jq -c '.skips')"
+          echo "image-content-hash=$(printf '%s' "$RESULT" | jq -r '.image_content_hash')"
+          echo "suite-code-hashes=$(printf '%s' "$RESULT" | jq -c '.suite_code_hashes')"
+        } >> "$GITHUB_OUTPUT"
```
sirutBuasai: Writing github output should all be put into the script function, that way we're not handling output via bash again. Python and bash are both scripting languages so generally we can always do everything in one language. For writing output, we can do

```
github_output_path = os.getenv('GITHUB_OUTPUT')
if github_output_path:
    with open(github_output_path, "a") as f:
        f.write(f"{output_key}={output_value}\n")

File: test/db/test_config_matches_workflows.py
Comment (sirutBuasai) [inline test/db/test_config_matches_workflows.py:1]: These unit tests should be inside `deep-learning-containers/test/db/` instead. Then we can create a `_prcheck.database.yml` workflow that runs those unit tests.

File: .github/actions/check-test-pass/check_test_pass.py [hunk 2]
```diff
@@ -0,0 +1,98 @@
...
+
+def _load_helpers(repo_root):
+    """Import the shared test-skip modules from the repo checkout."""
+    scripts_dir = Path(repo_root) / "scripts" / "ci" / "test_skip"
+    sys.path.insert(0, str(scripts_dir))
+    import hash_image_content
+    import hash_suite_code
+    import test_skip_db
+
+    return hash_image_content, hash_suite_code, test_skip_db
```
sirutBuasai: Why do we need this instead of just importing it in file?

.github/actions/check-test-pass/check_test_pass.py
```diff
@@ -0,0 +1,98 @@
...
+    print(
+        json.dumps(
+            {
+                "skips": skips,
+                "image_content_hash": image_content_hash,
+                "suite_code_hashes": suite_code_hashes,
+            },
+            separators=(",", ":"),
+        )
+    )
```
sirutBuasai: FYI, `pprint` can do the same

.github/actions/record-test-pass/record_pass.sh
```diff
@@ -0,0 +1,43 @@
...
+  {
+    echo "### ✅ Recorded test PASS"
+    echo "Suite \`$SUITE\` passed and was cached."
+    echo ""
+    echo "- image_content_hash: \`$IMAGE_CONTENT_HASH\`"
+    echo "- suite_code_hash: \`$SUITE_CODE_HASH\`"
+  } >>"$GITHUB_STEP_SUMMARY"
+else
+  echo "::warning::Failed to record PASS for '$SUITE' (cache write) — suite still passed."
+fi
```
sirutBuasai: I'm more in favor of writing this script as a logic of its own rather than calling other scripts sequentially.

Instead of calling scripts with some flag that each of the script to process and operate based on those script output, you should think of `hash_image_content` `hash_suite_code` as simply a library of functions.

Then your action script can be a python script that takes in those files as modules, then run the actual logic handling itself using those module library functions.

.github/workflows/_reusable.check-test-pass.yml
sirutBuasai, line 1: Why do we need a reusable workflow instead of just calling the actions directly?

scripts/ci/test_skip/hash_suite_code.py
```diff
@@ -0,0 +1,110 @@
...
+    )
+    parser.add_argument(
+        "--eligible-only",
+        action="store_true",
+        help="Print skip_eligible (true|false) for the suite instead of its hash.",
+    )
+    args = parser.parse_args(argv)
+    if args.eligible_only:
+        print("true" if is_skip_eligible(args.repo_root, args.suite) else "false")
+        return 0
```
sirutBuasai: This shouldn't be part of this script. This function should just be the part of the record_pass test script.

scripts/ci/test_skip/hash_image_content.py
```diff
@@ -0,0 +1,121 @@
...
+        default=DEFAULT_PLATFORM,
+        help=f"Platform to hash (default {DEFAULT_PLATFORM}).",
+    )
+    args = parser.parse_args(argv)
+    print(compute_image_content_hash(args.image_uri, platform=args.platform))
+    return 0
+
+
+if __name__ == "__main__":
+    sys.exit(main())
```
sirutBuasai: Do we even this this file to be an executable and run file directly? I think this file is mainly just a library of functions for actions to be used

scripts/ci/test_skip/ci_images_db.py
sirutBuasai, line 1: might want to rename this file. `test_*` is conventionally reserved for test files.

scripts/ci/test_skip/test_skip_db.py
```diff
@@ -0,0 +1,125 @@
...
+    if args.command == "record":
+        record_test_pass(
+            args.image_content_hash,
+            args.suite,
+            args.suite_code_hash,
+            ci_image_tag=args.ci_image_tag,
+        )
+        return 0
+
+    return 1
```
sirutBuasai: Instead of running this script as a record test sciprt. Make this script simply a db library script that is able to do various db API calls. Then the action scripts can import this library and use the module functions.

scripts/ci/test_skip/hash_image_content.py
```diff
@@ -0,0 +1,121 @@
...
+            "{{json .Image}}",
+        ],
+        capture_output=True,
+        text=True,
+        check=True,
+    )
+    return result.stdout
+
+
+def compute_image_content_hash(image_uri, platform=DEFAULT_PLATFORM):
```
sirutBuasai, line 102: Currently there are two bugs that I can think of.

1) I believe that the root.fs diff layer is only computed on `RUN` `COPY` `ADD` docker commands while `ENV` `LABEL` `CMD` don't change the root.fs layer but only change the image config json when you do `docker inspect`. This is what I vaguely remember though so you should try to test it locally and verify if this is true. If it is, then we might have to include the image config json into part of the hash as well. Just be sure that if we need to include the image config json, if there are any non-deterministic fields in that json, we need to parse it out. (eg: if the image build time is in the json, then we need to remove it bcuz that would make our hash non-deterministic)

2) Any scripts we fetch at runtime won't be detected. Unfortunately this might be somethign we need to accept. For example, if the docker file runs `s3 copy /path/to/s3_object.` If the content of s3_object has changed, it wont detect the new content. If you can think of a way to work around that without redesigning, great. If not, this is something we should document as a limitation in the codebase so that the next developers know when working on it.

.github/actions/check-test-pass/action.yml
samsullivan1: Moved this out of the action and into `check_test_pass.py`. I also moved the report cache hits step into the script.

.github/actions/check-test-pass/check_test_pass.py
samsullivan1: We need to add the `sys.path` insert since `scripts/ci/test_skip/` itself is not a package, I opted to put this logic in a private helper but I agree it would be clearer as a top level import. Fixed.

.github/actions/check-test-pass/check_test_pass.py
samsullivan1: Noted for the future. This got replaced with `_emit_outputs` after moving the github output into check_test_pass.py from the action file.

test/db/test_config_matches_workflows.py
samsullivan1: Created a new test dir `deep-learning-containers/test/db/` and moved the script tests there. I also created the `_prcheck.database.yml` workflow that runs the suite on path-scoped pull requests to main.

.github/actions/record-test-pass/record_pass.sh
samsullivan1: Re wrote this script as record_pass.py with the described import behavior.

scripts/ci/test_skip/hash_image_content.py
samsullivan1: After changing `record_pass.py` and `check_test_pass.py` to instead import these as modules, they do not need to be executable. Changed so they are no longer shebang scripts.

scripts/ci/test_skip/test_skip_db.py
samsullivan1: Done (see other comment responses/revisions).

scripts/ci/test_skip/ci_images_db.py
samsullivan1: Good call, changed to ci_images_db.py.

.github/workflows/_reusable.check-test-pass.yml
samsullivan1: As we wire this into more test pipelines, calling the action in the job would be a fair amount of duplicated code, since we would be copy pasting the ~20-line check job below into every test pipeline file (note either way it should be its own job so we only provision a single runner to do the DB check). With a reusable action, this becomes
```
check:
    needs: [build]
    uses: ./.github/workflows/_reusable.check-test-pass.yml
    with:                                                                                                                                                                               
        config-file: ${{ inputs.config-file }}                                                                                                                    
        image-uri: ${{ needs.build.outputs.image-uri || '' }}                                                                                                     
        suites: [...]
```
in each pipeline.

So this is essentially trading an extra layer of abstraction to keep from duplicating defining the runner, outputs, and account data. I think it could go either way; it is quite thin as an abstraction layer.

scripts/ci/test_skip/hash_image_content.py
samsullivan1: 1. Yes, that is correct. The DiffIDs only capture file system changes, so `ENV`, `LABEL`, and `CMD` will not get picked up by looking at the layers even though they can change run time behavior. I think I overlooked this in the initial design. I added the image config JSON into the hash and excluded the created time stamp and build history fields which are the only non deterministic fields. This all lives in the image config still so the overall hashing behavior is basically unchanged (we can still get the hash without pulling down the image). Also tested locally to ensure hashes are actually reproducible.

2. Hmm, I don't see a good way around this right now. I think the build-time version of this might even be a small unavoidable bug with the current implementation. Commands like `RUN aws s3 cp s3://bucket/obj /destination/path` that pull in external state and bake them into the layers can't be reliably detected since docker hashes the actual command string itself before building to decide whether to re use a layer in the cache (although this will re set daily as the layer cache gets re set, so its more of a bounded edge case). So right now even without test-skip I think its possible to rebuild an image with cache and technically run tests against a stale image:
For example:
```
-> Open PR which builds CI image and runs tests
-> external state changes
-> follow up commit, image rebuilds with full cache hit so we retest the outdated image.
```

 I added a doc string documenting both run-time and build-time limitations in `ci_images_db.py`.

scripts/ci/test_skip/hash_suite_code.py
samsullivan1: Resolved, moved the call to `record_pass.py:37`. I kept the `is_skip_eligible` function itself here since its called by both `record_pass.py` action and the `check_test_pass.py` action scripts.

.github/workflows/_reusable.check-test-pass.yml
sirutBuasai: I'm not leaning either ways to be honest. This works fine as is, if we need to cull, its pretty easy too. This can stay, thanks for the explanation!

test/db/test_hash_image_content.py
```diff
@@ -0,0 +1,118 @@
...
+        "rootfs": {"type": "layers", "diff_ids": DIFF_IDS},
+    }
+    monkeypatch.setattr(image_hasher, "_inspect_image_config", lambda uri: json.dumps(payload))
+    got = image_hasher.compute_image_content_hash(
+        "123.dkr.ecr.us-west-2.amazonaws.com/ci:tag", platform="linux/amd64"
+    )
+    assert got == image_hasher.content_hash(DIFF_IDS, payload["config"])
+
+
+def test_hash_ignores_build_timestamps(monkeypatch):
```
sirutBuasai, line 108: Theres assertion test case on a config change (e.g. Env/Labels) does change the hash

scripts/ci/test_skip/hash_suite_code.py
```diff
@@ -0,0 +1,72 @@
...
+        digest.update(path.read_bytes())
+        digest.update(b"\0")
+    return f"sha256:{digest.hexdigest()}"
+
+
+def is_skip_eligible(repo_root, suite):
+    """Return True iff the suite exists and is marked skip_eligible in the config.
+    An unknown suite is treated as not eligible.
+    """
+    suite_cfg = load_config(repo_root).get(suite)
```
sirutBuasai: instead of loading the config file each time we want to determine if a suite is available. Why don't we just load config once at top level module when it imports and use that config in other functions.

scripts/ci/test_skip/hash_image_content.py
samsullivan1: Follow up on 1.

I further verified the new image_content_hashes are deterministic and did some extra digging into what the config is made of. I'll summarize this and add it to the doc string. TLDR it works as is.

Every docker build is required to hold the following metadata in its top level json:
- `architecture`
- `os`
- `rootfs` (`type` and `diff_ids`)

Then there are optional entries that are only included if they have a non-null value. The `config` json object is included in this list, which appears only if any command that changes the config ran at any point during the build (e.g. there was an CMD line in the Dockerfile). The entries of `config` map 1:1 to a Dockerfile instruction that sets some execution parameter (see https://github.com/opencontainers/image-spec/blob/main/config.md#properties for more details and an example).

`ARG` is not included in `config`, so run time args cannot change the config hash (in particular `CACHE_REFRESH` is not included in the config hash). The only way it gets into the config hash is if it is promoted to ENV or LABEL after being declared. They can however affect the layer hash, but this is OK since we have layer caching.

I checked all the docker files and there are no runtime ARGs promoted to config values. They are all of the form, for example: `docker/sglang/Dockerfile.amzn2023`:
```
line 10: ARG SGLANG_REF=bc8b3ab1f55c60951b586d84d4aec773a6f654df
⋮
line 90: ENV SETUPTOOLS_SCM_PRETEND_VERSION=${FRAMEWORK_VERSION}.amzn2023.${SGLANG_REF}
```
which is a pinned value, so not at risk of changing.

Right now the `image_content_hash` only folds in:
- `diff_ids from rootfs`. These are the hashes of the actual image layer bytes + metadata.
- The config file. These hold only execution parameters declared in the Dockerfile and have (verified) no run time dependency.

So overall the current mechanism is safe and deterministic.

Commands that change the file system:
`RUN`, `COPY`, `ADD`, `WORKDIR` (only if the target path doesn't yet exist)

Commands that change the config object (and their key in the config json):
```
Config Key: Dockerfile Instruction
User: USER
ExposedPorts: EXPOSE
Env: ENV
Entrypoint: ENTRYPOINT
Cmd: CMD
Volumes: VOLUME
WorkingDir: WORKDIR
Labels: LABEL
StopSignal: STOPSIGNAL
```

test/db/test_hash_image_content.py
samsullivan1: Added a unit test asserting this.

scripts/ci/test_skip/hash_suite_code.py
samsullivan1: Good catch. I changed this to parse once and load the config into a global variable at the module level. I made it load lazily on first call rather than on import so we don't do file I/O if we import the module but don't call the function.

sirutBuasai (approved):

PR #6572, opened 2026-08-18
add(docs): enable autocurrency for huggingface-vllm

dwarez:
## Purpose

huggingface-vllm 0.26.0 and 0.27.1 are published on ECR but missing from the Available Images page. Root cause: `docs-pr.sh` skips frameworks not defined in the autocurrency tracker, and huggingface-vllm was never added (past entries were manual: #6393, #6429).

This PR:

1. **Adds the two missing available-images entries** (0.26.0, 0.27.1 — both already live on ECR).
2. **Adds huggingface-vllm to the autocurrency tracker**, so future releases get the docs entry auto-generated at release time.
3. **Gates the version auto-bump on base-image availability** (`wait_for_base_image: true`). huggingface-vllm layers on the AWS vLLM DLC, so an upstream vllm release does not imply a usable base image. With the flag set, the checker verifies the rendered base image exists on ECR before opening the auto-update PR — the PR arrives green instead of failing CI until the base is published. The flag is opt-in; `vllm`/`sglang` behavior is unchanged. The existence check is an anonymous HTTPS call to public.ecr.aws (no credentials) and fails open on lookup errors, so a transient error never blocks an update.
4. **Fixes the tag scheme in `docs-pr.sh`** for HF-contributed images: they publish transformers-scoped tags (`0.27.1-transformers5.10.2-gpu-py312-cu130-ubuntu22.04`), which the AWS-scheme `generate_tags` would get wrong. Detected via `transformers_version` in the release spec; AWS path unchanged.

## Test Plan

- `base_image_exists` run against live public ECR: existing tag → exists; nonexistent tag → absent; unknown registry → warn + proceed.
- Gate block executed against the real tracker config: base published → PR proceeds; base missing → clean skip; frameworks without the flag → gate off.
- `docs-pr.sh` generation simulated for both paths: HF output matches the published ECR tag exactly; AWS output identical to before.
- `mkdocs build --strict` + `pytest test/docs` (74/74) pass; both new tags render in the generated available_images page.

## Test Result

All of the above pass locally. The full checker run happens in CI (the script requires bash 4+).

<details>

## PR Checklist

</details>

### DIFF
--- a/.github/config/autocurrency-tracker.yml
+++ b/.github/config/autocurrency-tracker.yml
@@ -39,3 +39,18 @@ frameworks:
       - path: "docker/sglang/Dockerfile"
         base_image_template: "lmsysorg/sglang:v{version}"
     version_detection: true
+
+  huggingface-vllm:
+    github_repo: "vllm-project/vllm"
+    tag_prefix: "v"
+    # Layered on the AWS vLLM DLC: only open the auto-update PR once the
+    # vllm:{version}-*-sagemaker base image exists on ECR.
+    wait_for_base_image: true
+    config_files:
+      - path: ".github/config/image/huggingface-vllm/sagemaker.yml"
+        prod_image_template: "huggingface-vllm:{major}.{minor}-gpu-py312-cu130-ubuntu22.04"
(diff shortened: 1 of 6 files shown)


dwarez:
closing this for the moment as there are concerns about rate-hit limit failures (already happening with 2 frameworks)

PR #6588, opened 2026-08-19
fix(whisperx): prevent CPU-host startup crash in start_cuda_compat.sh

Yadan-Wei:
## Summary

On a CPU-only host (or any host where the NVIDIA driver version can't be
detected), `scripts/docker/whisperx/start_cuda_compat.sh` aborts the container
entrypoint, so the WhisperX image fails to start even though `server.py` already
falls back to CPU (`DEVICE = "cuda" if torch.cuda.is_available() else "cpu"`).

## Root cause

On a driverless host `NVIDIA_DRIVER_VERSION` resolves to an empty string and is
passed **unquoted** to `verlte`:

```sh
if verlte $NVIDIA_DRIVER_VERSION $CUDA_COMPAT_MAX_DRIVER_VERSION; then
```

so `verlte` receives a single argument and its `[ "$1" = "$2" ]` dereferences an
unbound `$2`. The entrypoints (`dockerd_entrypoint.sh`, `sagemaker_entrypoint.sh`)
`source` this script under `set -euo pipefail`, so `set -u` aborts the entrypoint
before `exec uvicorn`:

```
start_cuda_compat.sh: line 17: $2: unbound variable
```

## Fix

- Add an explicit `[ -z "$NVIDIA_DRIVER_VERSION" ]` guard that skips compat when
  no driver is detected (matching the script's documented "CPU hosts ... left
  untouched" intent).
- Quote the `verlte` arguments.

No behavior change on GPU: a non-empty driver version takes the same `verlte`
comparison as before.

## Testing

**Unit** — `test/whisperx/test_start_cuda_compat.py` (CPU-only unit suite, no
container/GPU) covers the no-driver regression plus the older-driver (compat
added), newer-driver (skipped), and no-compat-package branches. Verified it
**fails on the pre-fix script** (`$2: unbound variable`) and passes after the fix.

**End-to-end** — `test/whisperx/ec2/test_ec2_cpu.py` runs the image with no
`--gpus` on a CPU host and asserts the container starts and serves (`/ping`,
`/v1/models`) plus real transcription (basic json + word timestamps, `tiny`
model for speed). A new `ec2-cpu` job in `whisperx.tests-ec2.yml` runs it on the
CPU `default-runner`; the existing `ec2-gpu` job now ignores the file so it does
not double-run.

Validated on a CPU-only host against the image built from this branch: the
patched entrypoint logs `Skipping CUDA compat setup as no NVIDIA driver was
detected`, reaches `Application startup complete`, and `GET /ping` returns `200`
with `/v1/models` advertising the served model. The e2e test reproduces the
pre-fix abort (`$2: unbound variable`) exactly against the current prod image.

## Note

The same unquoted `verlte` line exists in
`scripts/docker/pytorch/start_cuda_compat.sh` and
`scripts/docker/common/start_cuda_compat.sh`, but those do not currently crash
because their entrypoints invoke the script via `bash` (a child shell that does
not inherit `set -u`) rather than `source`. These can be hardened in a follow-up.

### DIFF
--- a/scripts/docker/whisperx/start_cuda_compat.sh
+++ b/scripts/docker/whisperx/start_cuda_compat.sh
@@ -25,7 +25,15 @@ if [ -f /usr/local/cuda/compat/libcuda.so.1 ]; then
     NVIDIA_DRIVER_VERSION=$(nvidia-smi --query-gpu=driver_version --format=csv,noheader --id=0 2>/dev/null || true)
   fi
   echo "Current installed NVIDIA driver version is ${NVIDIA_DRIVER_VERSION}"
-  if verlte $NVIDIA_DRIVER_VERSION $CUDA_COMPAT_MAX_DRIVER_VERSION; then
+  # No driver detected (CPU host / undetectable): leave LD_LIBRARY_PATH untouched —
+  # compat only bridges an OLD driver, and there is nothing to bridge. This guard is
+  # load-bearing: the entrypoints `source` this under `set -u`, and an empty version
+  # passed unquoted to verlte collapses to a single arg, tripping an unbound-$2 abort
+  # in verlte that kills the entrypoint before uvicorn. Quote the args for the same
+  # reason (defense in depth if a driver version ever contains whitespace).
+  if [ -z "$NVIDIA_DRIVER_VERSION" ]; then
+    echo "Skipping CUDA compat setup as no NVIDIA driver was detected"
+  elif verlte "$NVIDIA_DRIVER_VERSION" "$CUDA_COMPAT_MAX_DRIVER_VERSION"; then
     echo "Adding CUDA compat to LD_LIBRARY_PATH"
     export LD_LIBRARY_PATH=/usr/local/cuda/compat:$LD_LIBRARY_PATH
     echo $LD_LIBRARY_PATH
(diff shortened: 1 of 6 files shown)


scripts/docker/whisperx/start_cuda_compat.sh
```diff
@@ -25,7 +25,15 @@ if [ -f /usr/local/cuda/compat/libcuda.so.1 ]; then
     NVIDIA_DRIVER_VERSION=$(nvidia-smi --query-gpu=driver_version --format=csv,noheader --id=0 2>/dev/null || true)
   fi
   echo "Current installed NVIDIA driver version is ${NVIDIA_DRIVER_VERSION}"
-  if verlte $NVIDIA_DRIVER_VERSION $CUDA_COMPAT_MAX_DRIVER_VERSION; then
+  # No driver detected (CPU host / undetectable): leave LD_LIBRARY_PATH untouched —
+  # compat only bridges an OLD driver, and there is nothing to bridge. This guard is
+  # load-bearing: the entrypoints `source` this under `set -u`, and an empty version
```
Jyothirmaikottu: Can we trim the comment?

Yadan-Wei: I would prefer to keep it because it is different from other frameworks who has start_cuda_compat.sh. The way we use this script is different from other frameworks that's why it will crash on non-GPU device.

Jyothirmaikottu (approved):

PR #6600, opened 2026-08-21
chore(tensorflow): nest training under docker/tensorflow/training/ to match inference

bhanutejagk:
## What

TensorFlow is the only framework family in the repo with both a training and an inference image, and the two used different layouts:

| | before | after |
|---|---|---|
| inference | `docker/tensorflow/inference/2.20/` | unchanged |
| training | `docker/tensorflow/2.21/` | `docker/tensorflow/training/2.21/` |
| inference scripts | `scripts/docker/tensorflow/inference/` | unchanged |
| training scripts | `scripts/docker/tensorflow/entrypoint.sh` | `scripts/docker/tensorflow/training/entrypoint.sh` |

## Wiring updated

Every reference to the moved paths, exhaustively:

- `.github/config/image/tensorflow/2.21-sagemaker-{cpu,cuda}.yml` — `build.dockerfile`
- `docker/tensorflow/training/2.21/Dockerfile.{cpu,cuda}` — `COPY` sources for `pyproject.toml` / `uv.lock` (both the builder and runtime stages) and `entrypoint.sh`
- `.github/workflows/tensorflow-training.pr-2.21-{cpu,cuda}.yml` — `on.paths` and the `build-change` paths-filter

## Bonus: path filters are no longer over-broad

All four TF PR workflows watched `scripts/docker/tensorflow/**`. Because inference was namespaced and training was not, that glob matched both families: a training-only script edit triggered a full **inference** PR run (up to 8 SageMaker endpoints), and vice versa. Namespacing the training scripts makes it possible to scope each workflow to its own subtree, which this PR does — `training/**` for the two training workflows, `inference/**` for the two inference ones.

## Scope

Pure relocation. No Dockerfile stage, build arg, image tag, prod tag, config key, or test changed. The build context is the repo root and the Dockerfile is passed via `-f` (`.github/actions/build-image/action.yml`), so no context path depends on the directory layout.

Deliberately **not** changed, to keep this reviewable as a rename:

- Config filenames stay `2.21-sagemaker-*.yml` (no `-training-`). PyTorch training is likewise `2.13-sagemaker-cuda.yml`, with the job type carried in `metadata.job_type` — the `-inference-` infix on the 2.20 configs exists only to disambiguate against the training configs in the same directory.
- `test/tensorflow/integration/sagemaker/` (training) vs `.../inference/` stays as is. `integration/sagemaker` matches `test/pytorch/integration/sagemaker` and the `pytorch/integration/sagemaker` key in `.github/config/test-suites.yml`.

## Verification

- `test/db` and `test/docs` pass. `test/docs/test_links.py` has 3 failures that reproduce on unmodified `main` and are unrelated to this change.
- `grep` confirms no reference to `docker/tensorflow/2.21`, `scripts/docker/tensorflow/entrypoint.sh`, or the bare `scripts/docker/tensorflow/**` glob remains anywhere in the repo.
- Every path referenced from a TF image config, workflow path filter, or Dockerfile `COPY` was checked to resolve to an existing file.


bhanutejagk:
Superseded by #6603 — same commit, but the branch now lives on aws/deep-learning-containers instead of a fork.

PR #6619, opened 2026-08-24
ray-train: decouple CI from Ray Serve via framework: ray_train

Eren-Jeager123:
## Summary
Decouples ray-train CI from Ray Serve. Both shared `metadata.framework: "ray"`, so the **security allowlist dir** and **build-hook dir** were shared — any edit to either triggered *both* PR workflows, and the build hook carried a `no-op for Ray Serve` guard.

Switches ray-train to `framework: ray_train` (same pattern already used by `ray_llm`, `sglang_server`, `sklearn_1_4_2_py312`) and gives it its own dirs:

| Concern | Before (shared) | After |
|---|---|---|
| Security allowlist | `ecr_scan_allowlist/ray/` | `ecr_scan_allowlist/ray_train/` |
| Build hooks | `ci/build/ray/` | `ci/build/ray_train/` |
| Sanity gate | `framework == ray && job_type == training` | `framework == ray_train` |

Also allowlists the three quinn-proto 0.11.14 aliases from the latest scan (`RUSTSEC-2026-0185`, `CVE-2026-25800`, `GHSA-4w2j-m93h-cj5j`) — same CVE in `/usr/local/bin/uv` from the base image, unfixable until upstream `uv` bundles quinn-proto >= 0.11.15. These now land in the ray-train allowlist **without** triggering the Ray Serve workflows.

## Release impact: none
`metadata.framework` feeds the release spec, and ray-train has `release: true`, so a matching `ray_train` entry was added to the release-logic `frameworks.yml` (separate package, CR to follow). It maps to the **same** `ray` ECR repo and renders the **same** `train-ml-cuda` tag. Verified: 150 tag tests pass.

## Note on telemetry
`test/telemetry/**` is intentionally left shared — it is listed in **all 28** framework PR workflows as cross-framework test infra, not a ray-train<->ray-serve coupling. Telemetry identity stays self-consistent (`x-framework` is asserted against the config value); it now reports `ray_train` instead of `ray`, which also makes the training image distinguishable from Ray Serve in telemetry.

## Follow-up found (not fixed here)
`ray-llm.pr-gpu.yml` watches `ecr_scan_allowlist/ray-llm/**` (hyphen) but the real dir is `ray_llm` (underscore) — so ray-llm allowlist edits do not trigger its own workflow. Out of scope; flagging for the owner.

## Test plan
- [ ] Editing `ecr_scan_allowlist/ray_train/` triggers only the Ray Train PR workflow
- [ ] Editing `ecr_scan_allowlist/ray/` triggers only the Ray Serve PR workflows
- [ ] Security test resolves the new allowlist (quinn-proto CVEs pass)
- [ ] Build hooks still fetch/upload the flash-attn/TE wheel cache
- [ ] Training sanity checks still run on the ray-train image


jinyan-li1 (dismissed):
Thanks!

jinyan-li1 (approved):

PR #6710, opened 2026-09-03
Remove unused dlc_minor_version from HF PyTorch training configs

qinzhounick:
### What
Removes the unused `dlc_minor_version: "0"` key from both HF PyTorch training 2.11 configs (`2.11-ec2.yml`, `2.11-sagemaker.yml`). It is dead config.

### Why it is dead
Keys under `build:` are forwarded as `--build-arg` (uppercased) by `.github/actions/build-image/resolve_build_args.py`. For this family that arg goes nowhere:

- `docker/huggingface/pytorch/Dockerfile` declares no `ARG DLC_MINOR_VERSION` and sets no `LABEL dlc_minor_version`, so Docker discards the unconsumed build-arg.
- The release logic never reads `dlc_minor_version` from this config either — it reads the **image LABEL**.

Contrast with keys that are live: `dlc_major_version` is declared as an `ARG` and applied as a `LABEL` in the Dockerfile, and `python_version` / `cuda_version` are consumed when rendering release tags.

### No behavior change
The released tag stays 3-part (`-v1.0.0`). The version scheme is chosen from the image's `dlc_minor_version` LABEL, and this image **inherits** `dlc_minor_version=0` from its base:

```
$ # labels on public.ecr.aws/deep-learning-containers/pytorch:2.11.0-cu130-amzn2023
dlc_major_version=1
dlc_minor_version=0
framework=pytorch
framework_version=2.11.0
```

Docker labels are inherited by child images, and this Dockerfile does not override that label, so removing the config key cannot affect the resolved version. This PR is cleanup only.

### Testing
- Both configs parse; `resolve_build_args.py` resolves 16 build-args for each, with `DLC_MINOR_VERSION` gone and `DLC_MAJOR_VERSION=1` retained.
- Cross-checked every remaining resolved build-arg against the Dockerfile's `ARG` declarations.
- No `dlc_minor_version` references remain under the family's config directory.


qinzhounick:
test failed due to CVE, unrelated to this change

sallyseok (dismissed):

PR #6732, opened 2026-09-10
feat(vllm): arm64/Graviton CPU vLLM DLC (EC2 + SageMaker, AL2023)

Jyothirmaikottu:
## Description

Adds an arm64/Graviton CPU vLLM DLC (EC2 + SageMaker, AL2023) on **vLLM 0.29.0**, the aarch64 companion to the x86 CPU vLLM DLC in #6721. Carries the shared AL2023 CPU foundation (Dockerfile, entrypoints, build hooks), the arm64 image configs, and a dedicated CPU CI path fully isolated from the GPU path.

## Changes

**CPU image foundation (shared with #6721)**
- `docker/vllm/Dockerfile.cpu.amzn2023`: multi-arch AL2023 CPU build (from vLLM source at the 0.29.0 ref). gcc14 (AL2023 stock gcc 11.5 cannot compile torch-inductor's Armv9 codegen), a `VLLM_CPU_ARM_BF16` build arg for the aarch64 bf16 kernels, `DLC_MAJOR_VERSION` defaulting to 1.
- CPU entrypoints/env (`cpu_dockerd_entrypoint.sh`, `cpu_sagemaker_entrypoint.sh`, `vllm_cpu_env.sh`) and build hooks (`scripts/ci/build/vllm_server/{pre,post}_build.sh`, which no-op for GPU/omni).

**arm64 image configs**
- `ec2-arm64-amzn2023.yml`, `sagemaker-arm64-amzn2023.yml`: `framework_version: 0.29.0`, `vllm_ref: ec4a3a5`, `arch_type: arm64`, `vllm_cpu_arm_bf16: "1"`, `dlc_major_version: 1`, `environment: gamma`.

**Dedicated CPU CI path (isolated from GPU)**
- `vllm.pipeline-cpu.yml`: lean CPU pipeline (build on the Graviton fleet + sanity/unit/security/telemetry + CPU model + CPU SageMaker tests; no upstream/EFA).
- `vllm.pr-cpu-amzn2023.yml`: CPU PR workflow triggering only on CPU paths, discovering `*-arm64-amzn2023.yml`.
- `vllm-cpu-model-tests.yml`, `vllm-cpu-sagemaker-endpoint-tests.yml`: CPU-only test configs (arm64 smoke on the Graviton fleet, SageMaker endpoint on `ml.c7g`/`ml.c8g`).
- `vllm.tests-sagemaker-cpu.yml`: CPU-only fork of the SageMaker test workflow (single-model matrix, no GPU standalone test); the shared GPU `vllm.tests-sagemaker.yml` is left untouched.
- `vllm.tests-model.yml`: adds an optional `test-config` + `device-type` input (defaults preserve GPU behavior byte-for-byte) so the CPU pipeline runs the smoke test without `--gpus` against the CPU config.
- `vllm.pr-amzn2023.yml`: excludes arm64 configs and CPU entrypoint scripts from its triggers so a CPU-only change never builds GPU images.

**Security**
- `vllm_server`/`vllm` ECR scan allowlist: refreshed an expired `review_by` and kept the shared entries so the amzn2023/ubuntu vLLM scans stay green (this PR rebuilds those images via shared build paths).

## Verification

- vLLM `+cpu` serving proven on a real Graviton4 (`c8g.4xlarge`, AL2023): `/health` 200 + valid chat completion (validated on 0.28.0). 0.29.0 is a straight version match to the GPU AL2023 bump (#6761, ref ec4a3a5): `requirements/cpu.txt` keeps the same aarch64 `torch==2.13.0+cpu` and `VLLM_CPU_ARM_BF16` is still CMake-consumed, so the CPU build path is unchanged — 0.29.0 is validated by the CPU PR pipeline (build + smoke + SageMaker endpoint on the Graviton fleet).
- Local checks: build args resolve `VLLM_CPU_ARM_BF16=1`; each CPU test config parses to only its arm64 entry.

## Follow-ups (not in this PR)

`release: false` / `environment: gamma`. Remaining release wiring: arm64 CPU autorelease callers, and flipping to `release: true` + production after gamma validation.


.github/workflows/vllm.pipeline.yml
```diff
@@ -176,14 +188,15 @@ jobs:
     secrets: inherit
 
   model-tests:
-    if: ${{ always() && !cancelled() && inputs.run-model-test && needs.build.result != 'failure' && fromJSON(needs.check.outputs.skips || '{}')['vllm/model'] != true }}
-    needs: [build, check]
+    if: ${{ always() && !cancelled() && inputs.run-model-test && (needs.ci-config.outputs.device-type == 'gpu' || needs.ci-config.outputs.device-type == 'cpu') && needs.build.result != 'failure' && fromJSON(needs.check.outputs.skips || '{}')['vllm/model'] != true }}
```
sallyseok, line 191: Should delete `|| needs.ci-config.outputs.device-type == 'cpu')`

.github/workflows/vllm.pr-amzn2023.yml
```diff
@@ -12,6 +12,7 @@ on:
       - ".github/workflows/vllm.pr-amzn2023.yml"
       - ".github/workflows/vllm.tests-unit.yml"
       - "docker/vllm/Dockerfile.amzn2023"
+      - "docker/vllm/Dockerfile.cpu.amzn2023"
```
sallyseok, line 15: Let's create a separate pr testing workflow file for the cpu Dockerfile. Again, to avoid the situation where a change to one Dockerfile triggers build and tests for the other, unrelated Dockerfile. We can call it `vllm.pr-cpu-amzn2023.yml` and include both the x86 and arm64 in there, or we can separate further into arm64 cpu and x86 cpu if they differ enough in their tests. 

In that file, the discover-configs job would need a different pattern like
```
      - id: discover
        uses: ./.github/actions/discover-configs
        with:
          pattern: ".github/config/image/vllm/*-arm64-amzn2023.yml"
```

.github/config/image/vllm/ec2-arm64-amzn2023.yml
```diff
@@ -0,0 +1,32 @@
...
+  dlc_minor_version: "0"
+
+release:
+  # Autorelease wiring + validation pending; flip to true once CI is green.
+  release: false
+  force_release: false
+  public_registry: true
+  private_registry: true
+  enable_soci: true
+  environment: "production"
```
sallyseok, line 32: This is a new image type, so we should gamma test the release first. Let's set this to `environment: "gamma"`

.github/config/model-tests/vllm-model-tests.yml
```diff
@@ -3,6 +3,7 @@
 # Smoke tests: model serving + inference validation.
 # smoke-test has codebuild-fleet and runner-scale-sets sub-keys.
 # Workflow parsers construct s3_path from s3_prefix + s3_model.
+# required_image_pattern scopes each entry to gpu vs cpu images.
```
sallyseok, line 6: For this file and `.github/config/model-tests/vllm-sagemaker-endpoint-tests.yml`, we should just create new config files for cpu. I want to err towards separating the test triggering file paths so that small changes don't trigger tests for unrelated images

docker/vllm/Dockerfile.cpu.amzn2023
```diff
@@ -0,0 +1,252 @@
...
+ENV VLLM_CPU_OMP_THREADS_BIND=nobind
+
+# =============================================================================
+# STAGE 4: DLC overlay — Amazon DLC customizations on top of the vLLM runtime
+# =============================================================================
+FROM runtime AS base
+
+ARG PYTHON="python3"
+ARG PYTHON_VERSION=3.12
+ARG DLC_MAJOR_VERSION=2
```
sallyseok, line 156: why is this 2

sallyseok (changes requested):

sallyseok (approved):

PR #6735, opened 2026-09-10
feat(pytorch): add PyTorch 2.14 currency scaffold

Eren-Jeager123:
## Summary

Adds PyTorch 2.14.0 (cu133, Amazon Linux 2023): 4 image configs, 4 workflow callers, release-schedule entries, and per-version `pyproject.toml` + `uv.lock`.

Ships with `release.environment: "gamma"` on all 4 configs. The production flip is a separate follow-up once gamma autoreleases have validated the image.

## Pins

| Pin | 2.13 | 2.14 | Note |
|---|---|---|---|
| torch | 2.13.0 | **2.14.0** | cu130 wheel index (no cu133 index exists upstream) |
| torchvision | 0.28.0 | **0.29.0** | |
| torchaudio | 2.11.0 | 2.11.0 | held — upstream `pytorch/audio` has no 2.12+ tag |
| cuda | 13.3.0 | **13.3.1** | newest `nvidia/cuda:*-devel-amzn2023`; `cu133` tag unchanged |
| python | 3.12 | 3.12 | held — see below |
| transformer-engine | 2.17.0 | **2.18.0** | needs a new dep, see below |
| flash-attn | 2.8.3 | **2.8.3.post1** | needs a build fix, see below |
| deepspeed | 0.19.2 | **0.19.6** | |
| efa | 1.49.0 | **1.50.0** | `install_efa_amzn2023.sh` already passes `--disable-ngc` for >=1.48.0 |
| nccl 2.30.7-1, nccl-tests v2.20.0, gdrcopy 2.6 | | held | still newest |

## flash-attn cannot build against torch 2.14 unpatched

torch 2.14's `c10/util/intrusive_ptr.h` uses `std::strong_ordering` and `operator<=>` behind an unconditional `#include <compare>`, so torch C++ extensions must now compile as C++20. flash-attn 2.8.3.post1 hardcodes `-std=c++17`, and nvcc dies on the first object file:

```
torch/include/c10/util/intrusive_ptr.h(775): error: namespace "std" has no member "strong_ordering"
```

There is no flash-attn release to bump to — upstream `main` still pins C++17 for the FA2 extension, and [Dao-AILab/flash-attention#2863](https://github.com/Dao-AILab/flash-attention/issues/2863) is open with the same finding. So the source-build branch in `Dockerfile.cuda` now fetches the sdist, rewrites the pinned `-std=c++17` to `c++20`, and builds from the patched tree.

Verified by compiling `flash_bwd_hdim128_bf16_sm80.cu` against torch 2.14 headers both ways: C++17 reproduces the errors above, C++20 exits 0. Applied unconditionally rather than gated, since every torch from 2.14 on needs it and 2.11–2.13 install the cached wheel instead of reaching this branch.

## Transformer Engine 2.18 needs `nvidia-cudnn-frontend`

TE 2.18 replaced its bundled cuDNN Frontend submodule with the `nvidia-cudnn-frontend` pip package, and `transformer_engine/common/CMakeLists.txt` now hard-fails without `cudnn_frontend.h`. The Dockerfile installs TE with `--no-build-isolation`, so `[build-system] requires` is never installed for us.

Fix: `nvidia-cudnn-frontend>=1.26.0` added to `docker/pytorch/2.14/cuda/pyproject.toml` — the per-version pin dir, not the shared Dockerfile, so 2.11–2.13 are untouched. Resolves to 1.28.0. `builder-base` builds green locally with this in place.

## Licence

`/license.txt` is fetched by the shared Dockerfile, parameterised on `FRAMEWORK_SHORT_VERSION`, which CI derives from `metadata.framework_version` — so 2.14 picks up `pytorch-2.14/license.txt` with no config change.

**The object is uploaded**: `pytorch-2.14/license.txt`, 4030 bytes, verified readable anonymously (the same path the Docker build takes) and carrying `PyTorch 2.14 is licensed from:`. Same composition as the existing `pytorch-2.13` object — upstream `LICENSE` is byte-identical across tags, so the only difference is the version line.

Verified the sanity check end to end against a container holding the real object: correct licence passes, a 2.13 licence in a 2.14 image fails.

## Python held at 3.12

torch 2.14 ships cp312–cp315 wheels, but the shared Dockerfiles hardcode `python3.12` in `LD_LIBRARY_PATH`, so a Python bump would change every PyTorch minor version at once. Out of scope for a currency PR.

## CVE floors aligned

The `sagemaker` extra now matches 2.11/2.12/2.13 pin for pin — `mlflow>=3.16.0`, `cryptography>=50.0.0`, `tornado>=6.5.8`, `sqlparse>=0.6.0`. Floors only, no resolution change: 2.14's locks already resolved those versions.

No `pytorch_runtime-2.14.0.json` ECR scan allowlist is needed — the CVE that required one on older versions is closed at source by the cryptography floor, and the 3-level loader skips missing version files.

## Verification

- `uv lock` resolves cleanly for `2.14/cuda` and `2.14/cpu`; all four versions' pins agree
- `check_release_schedule.py` — OK
- `resolve_build_args.py` — all 2.14 configs resolve; `compute_ci_tag.sh` → `...-py312-cu133-<suffix>`, matching `prod_image`
- Licence object readable anonymously; sanity check passes for 2.14 and fails on a version mismatch
- pre-commit — all hooks pass

Remaining risk is the first CUDA build itself: the wheel cache is keyed by `torch_version`, so it is cold and flash-attn plus TE compile from source.

### DIFF
--- a/.github/config/image/pytorch/2.14-ec2-cpu.yml
+++ b/.github/config/image/pytorch/2.14-ec2-cpu.yml
@@ -1,0 +1,32 @@
+image:
+  name: "pytorch-2.14-ec2-cpu"
+  description: "PyTorch 2.14 CPU training for EC2 instances"
+
+metadata:
+  framework: "pytorch_runtime"
+  framework_version: "2.14.0"
+  os_version: "amzn2023"
+  customer_type: "ec2"
+  arch_type: "x86"
+  device_type: "cpu"
+  job_type: "training"
+  prod_image: "pytorch:2.14-cpu-amzn2023"
+
+build:
+  dockerfile: "docker/pytorch/Dockerfile.cpu"
+  target: "runtime"
+  python_version: "3.12"
+  torch_version: "2.14.0"
+  torchvision_version: "0.29.0"
+  torchaudio_version: "2.11.0"
+  max_jobs: "8"
+  dlc_major_version: "1"
+  dlc_minor_version: "0"
+
+release:
+  release: true
+  force_release: false
+  public_registry: true
+  private_registry: true
+  enable_soci: true
+  environment: "gamma"
--- a/.github/workflows/pytorch.pr-2.14-cuda.yml
+++ b/.github/workflows/pytorch.pr-2.14-cuda.yml
@@ -1,0 +1,127 @@
+name: "PR - PyTorch 2.14 CUDA"
+
+on:
+  pull_request:
+    branches: [main]
+    types: [opened, reopened, synchronize]
+    paths:
+      - ".github/config/image/pytorch/2.14-*-cuda.yml"
+      - ".github/workflows/pytorch.pipeline.yml"
+      - ".github/workflows/pytorch.pr-2.14-cuda.yml"
+      - "docker/pytorch/Dockerfile.cuda"
+      - "docker/pytorch/2.14/cuda/**"
+      - "scripts/docker/common/**"
+      - "scripts/docker/pytorch/**"
+      - "scripts/docker/telemetry/**"
+      - "test/efa/**"
+      - "test/pytorch/multi_gpu/**"
+      - "test/pytorch/multi_node/**"
+      - "test/pytorch/single_gpu/**"
+      - "test/pytorch/unit/**"
+      - "test/sanity/**"
+      - "test/security/data/ecr_scan_allowlist/pytorch_runtime/**"
+      - "test/telemetry/**"
+      - "!docs/**"
+
+permissions:
+  contents: read
+  pull-requests: read
+
+jobs:
+  gatekeeper:
+    runs-on:
+      - codebuild-runner-${{ github.run_id }}-${{ github.run_attempt }}
+        fleet:default-runner
+        buildspec-override:true
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          ref: ${{ github.event.pull_request.base.sha }}
+          fetch-depth: 1
+      - uses: ./.github/actions/pr-permission-gate
+
+  check-changes:
+    needs: [gatekeeper]
+    runs-on:
+      - codebuild-runner-${{ github.run_id }}-${{ github.run_attempt }}
+        fleet:default-runner
+        buildspec-override:true
+    outputs:
+      build-change: ${{ steps.changes.outputs.build-change }}
+      efa-test-change: ${{ steps.changes.outputs.efa-test-change }}
+      multi-gpu-test-change: ${{ steps.changes.outputs.multi-gpu-test-change }}
+      multi-node-test-change: ${{ steps.changes.outputs.multi-node-test-change }}
+      sagemaker-test-change: ${{ steps.changes.outputs.sagemaker-test-change }}
+      sanity-test-change: ${{ steps.changes.outputs.sanity-test-change }}
+      single-gpu-test-change: ${{ steps.changes.outputs.single-gpu-test-change }}
+      telemetry-test-change: ${{ steps.changes.outputs.telemetry-test-change }}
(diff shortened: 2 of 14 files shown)


.github/config/image/pytorch/2.14-ec2-cpu.yml
```diff
@@ -0,0 +1,32 @@
...
+  dlc_major_version: "1"
+  dlc_minor_version: "0"
+
+release:
+  release: true
+  force_release: false
+  public_registry: true
+  private_registry: true
+  enable_soci: true
+  environment: "gamma"
```
sallyseok, line 32: We can keep the gamma for this PR but we probably don't need to gamma test, since this release path and tag pattern are already well-established

.github/workflows/pytorch.pr-2.14-cuda.yml
```diff
@@ -0,0 +1,127 @@
...
+      matrix:
+        include: ${{ fromJson(needs.discover.outputs.configs) }}
+    uses: ./.github/workflows/pytorch.pipeline.yml
+    with:
+      config-file: ${{ matrix.config_file }}
+      tag-suffix: pr-${{ github.event.pull_request.number }}
+      build: ${{ needs.check-changes.outputs.build-change == 'true' }}
+      run-efa-test: ${{ needs.check-changes.outputs.build-change == 'true' || needs.check-changes.outputs.efa-test-change == 'true' }}
+      run-multi-gpu-test: false   # NOTE: Temporarily disable test due to CI resource contention
+      run-multi-node-test: false  # NOTE: Temporarily disable test due to CI resource contention
```
sallyseok, line 119: For these two flags `run-multi-gpu-test` and `run-multi-node-test`, it looks like the existing PR workflow files also have these set to false, but I do think we should enable these and make sure it passes at least once for this new version

docker/pytorch/2.14/cuda/pyproject.toml
```diff
@@ -0,0 +1,62 @@
...
+    "cryptography>=50.0.0",  # CVE-2026-69247, CVE-2026-69249 (transitive)
+    "pyasn1>=0.6.4",  # CVE-2026-59884 / CVE-2026-59885 / CVE-2026-59886 (transitive)
+    "sqlparse>=0.6.0",  # CVE-2026-54284, CVE-2026-59893, CVE-2026-71491 (transitive via mlflow-skinny)
+]
+
+[tool.uv]
+
+[[tool.uv.index]]
+name = "pytorch-cu130"
+url = "https://download.pytorch.org/whl/cu130"
```
sallyseok, line 56: Do you happen to know why we bumped to cu133 for pytorch 2.13? Looks like upstream pytorch only supports up to cu132 (https://pytorch.org/get-started/locally/). Would it be better to set this to cu132 instead of cu130 so there's less of a gap? If the tests are passing I guess it doesn't matter

.github/workflows/pytorch.pr-2.14-cuda.yml
sallyseok: I'm ok with merging this as-is since all checks are passing, and then enable these in a separate PR before setting release environment to production

docker/pytorch/2.14/cuda/pyproject.toml
sallyseok: Found my answer here lol https://github.com/aws/deep-learning-containers/pull/6365#discussion_r3591875219

sallyseok (approved):

.github/workflows/pytorch.pr-2.14-cuda.yml
Eren-Jeager123: I'll keep gamma this time.

PR #6746, opened 2026-09-14
temp fix: sglang upstream-tests CI - seed hellaswag + floor datasets for pyarrow

jinyan-li1:
Two fixes for the sglangse-a-test-1-gpu-small):
1. HellaSwag — the upstream data URL (rowanz/hellaswag) was removed, so the download 404s and the suite fails. Seed the validation data from the HF mirror (Rowan/hellaswag) and point the test's cache check at the seeded file so it's used instead of re-downloading. Remove once sglang_ref includes sgl-project/sglang#39892 (dropped HellaSwag for MMLU).
2. datasets/pyarrow — olnsionType (removed inpyarrow ≥18), crashing on import and killing the whole suite. Floor datasets>=2.19 (uses pa.ExtensionType).

**Test:** fixed failing test: https://github.com/aws/deep-learning-containers/actions/runs/35381464712/job/105732614141?pr=6746


jinyan-li1:
Do not merge yet, root cause is upstream

sallyseok (approved):

